PR #8, opened 2021-05-06
Adds initial PEG, basic recognizer API, and Error/Result APIs.

almann:
Adds `pest`/`pest_derive` dependencies and a very basic PEG definition
that is just a stream of PartiQL keywords.  This implementation is just
to flesh out some of the usage of `pest`, namely the derivation macro
and its error APIs.

This commit exposes a function `recognize_partiql` that runs the parser
over an input slice and returns `()` on success or fails on invalid
syntax.  Given that the syntax is just a stream of keywords, this is not
particularly interesting right now.

Adds the following modules:
* `result` contains the main `ParserError` enum, `Position`, and
  `ParserResult` alias.
* `prelude` exports most of `result` and will be used in the future for
  things like traits as well.
* `peg` is the module that hosts the low-level Pest PEG parser and the
  grammar definition file.

Other notes:
* Adds `From` implementation for `pest::error::Error` for `ParserError`
  to do basic translation of the relatively obtuse Pest error to
  something a bit more friendly.
* Updates README around versioning.
* Changes stub tests to just be no-op for the non-parser packages.
* Adds dependency for `thiserror` for less boilerplate error
  definitions.
* Adds dev dependency for `rstest` for table driven testing.


partiql-parser/src/peg.rs
```diff
@@ -0,0 +1,48 @@
...
+//! Contains the [Pest](https://pest.rs) defined parser for PartiQL and a wrapper APIs that
+//! can be exported for users to consume.
+
+use crate::prelude::*;
+use pest::Parser;
+use pest_derive::Parser;
+
+#[derive(Parser)]
+#[grammar = "partiql.pest"]
+struct PartiQLParser;
```
almann, line 12: This is the macro that generates the Rust PEG parser from our declarative definition.  Note that the parser is fairly low-level (returning `Pair` type structures), so we'll never export this directly but wrap it into something that will produce an AST.

partiql-parser/src/peg.rs
```diff
@@ -0,0 +1,48 @@
...
+/// Recognizer for PartiQL queries.
+///
+/// Returns `Ok(())` in the case that the input is valid PartiQL.  Returns `Err([ParserError])`
+/// in the case that the input is not valid PartiQL.
+///
+/// This API will be replaced with one that produces an AST in the future.
+pub fn recognize_partiql(input: &str) -> ParserResult<()> {
+    PartiQLParser::parse(Rule::Keywords, input)?;
+    Ok(())
+}
```
almann, line 23: Just something to pass/fail for now...

partiql-parser/src/prelude.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+//! Convenience export of common traits and basic types that are almost always
+//! needed when using the parser APIs.
+
+pub use crate::result::ParserError;
+pub use crate::result::ParserResult;
+pub use crate::result::Position;
```
almann, line 8: The idea here is we'll put our most basic common types/traits that anyone using the parser will invariably need to import so users can `use partiql_parser::prelude::*`.  I've seen this pattern in various libraries, but it isn't totally ubiquitous so perhaps we'll back out of this later...

partiql-parser/src/result.rs
```diff
@@ -0,0 +1,82 @@
...
+use pest::error::{ErrorVariant, LineColLocation};
+use std::fmt;
+use thiserror::Error;
+
+/// Position in the source for an error.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Position {
+    Unknown,
+    At { line: usize, column: usize },
+}
```
almann, line 14: This variant makes it easier to deal with the case that we might not have a position.  Could've used `Option<Position>` and make this a `struct` but that makes implementing `Display` and embedding in our `ParserError` more clumsy, so I just encoded the variants here.

partiql-parser/src/result.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl ParserError {
+    /// Convenience function to create a [SyntaxError].
+    fn syntax_error<S: Into<String>>(message: S, position: Position) -> Self {
+        Self::SyntaxError {
+            message: message.into(),
+            position,
+        }
+    }
```
almann, line 39: This makes it easier to construct `ParserError::SyntaxError` from anything that could be a `String` (in Rust literals like `"hello"` are not `String` but `&str` [which is read as string slice])

partiql-parser/src/result.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl<R> From<pest::error::Error<R>> for ParserError
+where
+    R: fmt::Debug,
+{
+    fn from(error: pest::error::Error<R>) -> Self {
```
almann, line 46: This makes it so we can convert a Pest error to a `ParserError` automatically via the `?` error handling operator in Rust.

almann (commented):
Some comments walking through the rationale of this change below.

partiql-parser/src/partiql.pest
```diff
@@ -0,0 +1,263 @@
+WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
+
+// really basic rules to just detect a sequence of keywords
+// two dip our toes in Pest
+
+Keywords = _{ SOI ~ Keyword+ ~ EOI}
```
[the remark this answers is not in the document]
marcbowes: Even this is useful (e.g. for syntax highlighting).

marcbowes (approved):

zslayton (approved):
A few minor notes to consider below.

partiql-parser/src/result.rs
```diff
@@ -0,0 +1,82 @@
...
+        // obtain the line/column information from the Pest error
+        let (line, column) = match error.line_col {
+            LineColLocation::Pos((line, column)) => (line, column),
+            LineColLocation::Span((line, column), _) => (line, column),
```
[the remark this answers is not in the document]
almann: It may or may not be relevant--I figure we can add more information if we need to and add a variant to the `Position` enum below.

partiql-parser/src/result.rs
almann: This has changed a little in #12 to make the structure more explicit (but not to the degree you are suggesting).  Basically it still looks a lot like `Option`, but the main reason for the breakout was actually a friction between the `Display` implementation in the derived `Error` (via `thiserror`) and `Option<...>` which was my original variant.

The other reason why it is not `Option` is that I think we might have this be potentially a span if applicable, but that is not a given...

partiql-parser/src/result.rs
```diff
@@ -0,0 +1,82 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Position {
+    Unknown,
+    At { line: usize, column: usize },
+}
+
+impl fmt::Display for Position {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Position::Unknown => write!(f, "UNKNOWN POSITION"),
```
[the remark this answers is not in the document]
almann: That seems reasonable.

PR #12, opened 2021-05-10
Initial API for `Scanner`.

almann:
Provides an interface for _lexing_ PartiQL.  Traditionally we would
factor this as the low-level interface for the parser, but with the PEG
based implementation, we just surface the relevant parts of the parser
as a rule to parse with.

This commit helps explore how to effectively integrate with the somewhat
low-level APIs for parsing that Pest provides.

Notes:
* Adds `scanner` module.
* Adds `Scanner` trait to the prelude.
* Makes `PartiQLParser` public to crate.
* Refactors `LineAndColumn` as a tuple for `Position::At`.
  - Adds this to the `prelude`.
* Changes entry point for PEG to `Query` and added `Scanner` as the
  entry point rules for implementing the `Scanner` API.
* Adds `PairsExt`/`PairExt` trait/impl to add utility methods for working
  with Pest `Pairs`/`Pair`.
* Adds `LineAndColumn::move_relative` and cleans up some doc/doc tests.

Pest links for reference:
* https://pest.rs/book/parser_api.html - API description
* https://docs.rs/pest/2.1.3/pest/index.html - API Reference

This PR is dependent on #8 and as such is targeting a staging branch but I did not want to prevent review.

### DIFF
--- a/partiql-parser/src/lib.rs
+++ b/partiql-parser/src/lib.rs
@@ -32,8 +32,8 @@
 //!         Keyword(kw) => assert_eq!("FROM", kw),
 //!     }
 //!     // the other thing we can do is get line/column information from a token
-//!     assert_eq!(LineAndColumn::at(1, 8), second.start_loc());
-//!     assert_eq!(LineAndColumn::at(1, 12), second.end_loc());
+//!     assert_eq!(LineAndColumn::at(1, 8), second.start());
+//!     assert_eq!(LineAndColumn::at(1, 12), second.end());
 //!
 //!     // this API is built on immutable slices, so we can restart scanning from any token
 //!     scanner = first.into();
--- a/partiql-parser/src/lib.rs
+++ b/partiql-parser/src/lib.rs
@@ -32,8 +32,8 @@
 //!         Keyword(kw) => assert_eq!("FROM", kw),
 //!     }
 //!     // the other thing we can do is get line/column information from a token
-//!     assert_eq!(LineAndColumn::at(1, 8), second.start());
-//!     assert_eq!(LineAndColumn::at(1, 12), second.end());
+//!     assert_eq!(LineAndColumn::try_at(1, 8)?, second.start());
+//!     assert_eq!(LineAndColumn::try_at(1, 12)?, second.end());
 //!
 //!     // this API is built on immutable slices, so we can restart scanning from any token
 //!     scanner = first.into();
--- a/partiql-parser/src/lib.rs
+++ b/partiql-parser/src/lib.rs
@@ -2,10 +2,54 @@
 
 //! Provides a parser for the [PartiQL][partiql] query language.
 //!
+//! # Usage
+//!
+//! An API to interact with PartiQL tokens is the [`mod@scanner`] module.
+//! The [`scanner()`] function creates a [`Scanner`](scanner::Scanner) instance
+//! that one can use to parse tokens incrementally from some input slice.
+//!
+//! ```
+//! use partiql_parser::prelude::*;
+//! use partiql_parser::scanner;
+//!
+//! fn main() -> ParserResult<()> {
+//!     use partiql_parser::scanner::Content::*;
+//!
+//!     let mut scanner = scanner("SELECT FROM");
+//!     let first = scanner.next_token()?;
+//!
+//!     // get the parsed variant of the token
+//!     match first.content() {
+//!         Keyword(kw) => assert_eq!("SELECT", kw),
+//!     }
+//!     // the entire text of a token can be fetched--which looks the roughly the
+//!     // same for a keyword.
+//!     assert_eq!("SELECT", first.text());
+//!     
+//!     let second = scanner.next_token()?;
+//!     // get the parsed variant of the token
+//!     match second.content() {
+//!         Keyword(kw) => assert_eq!("FROM", kw),
+//!     }
+//!     // the other thing we can do is get line/column information from a token
+//!     assert_eq!(LineAndColumn::at(1, 8), second.start_loc());
+//!     assert_eq!(LineAndColumn::at(1, 12), second.end_loc());
+//!
+//!     // this API is built on immutable slices, so we can restart scanning from any token
+//!     scanner = first.into();
+//!     let second_again = scanner.next_token()?;
+//!     assert_eq!(second, second_again);
+//!     
+//!     Ok(())
+//! }
+//! ```
+//!
 //! [partiql]: https://partiql.org
 
 mod peg;
 pub mod prelude;
 pub mod result;
+pub mod scanner;
 
 pub use peg::recognize_partiql;
+pub use scanner::scanner;
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -1,9 +1,15 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
 
-// really basic rules to just detect a sequence of keywords
-// two dip our toes in Pest
+// TODO implement a full grammar, this is a very primitive version to start
+//      working with Pest and its APIs.
 
-Keywords = _{ SOI ~ Keyword+ ~ EOI}
+// Entry point for full query parsing
+Query = _{ SOI ~ Keyword+ ~ EOI}
+
+// Entry point for query "scanning"
+// Note that this is factored this way to support an iteration style API
+// where we can call back into this rule on subsequent input.
+Scanner = _{ SOI ~ Keyword }
 
 Keyword = { AllKeywords }
 
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -28,7 +28,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 if let Some(other_pair) = self.next() {
                     syntax_error(
                         format!("Expected one token pair, got: {:?}, {:?}", pair, other_pair),
-                        pair.start_loc().into(),
+                        pair.start().into(),
                     )?;
                 }
                 Ok(pair)
@@ -44,20 +44,20 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
 /// Extension methods for working with [`Pair`].
 pub(crate) trait PairExt<'val, R: RuleType> {
     /// Translates the start position of the [`Pair`] into a [`LineAndColumn`].
-    fn start_loc(&self) -> LineAndColumn;
+    fn start(&self) -> LineAndColumn;
 
     /// Translates the end position of the [`Pair`] into a [`LineAndColumn`].
-    fn end_loc(&self) -> LineAndColumn;
+    fn end(&self) -> LineAndColumn;
 }
 
 impl<'val, R: RuleType> PairExt<'val, R> for Pair<'val, R> {
     #[inline]
-    fn start_loc(&self) -> LineAndColumn {
+    fn start(&self) -> LineAndColumn {
         self.as_span().start_pos().line_col().into()
     }
 
     #[inline]
-    fn end_loc(&self) -> LineAndColumn {
+    fn end(&self) -> LineAndColumn {
         self.as_span().end_pos().line_col().into()
     }
 }
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -28,7 +28,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 if let Some(other_pair) = self.next() {
                     syntax_error(
                         format!("Expected one token pair, got: {:?}, {:?}", pair, other_pair),
-                        pair.start().into(),
+                        pair.start()?.into(),
                     )?;
                 }
                 Ok(pair)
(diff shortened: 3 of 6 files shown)


partiql-parser/src/partiql.pest
```diff
@@ -1,9 +1,13 @@
...
-// two dip our toes in Pest
+// TODO implement a full grammar, this is a very primitive version to start
+//      working with Pest and its APIs.
 
-Keywords = _{ SOI ~ Keyword+ ~ EOI}
+// Entry point for full query parsing
+Query = _{ SOI ~ Keyword+ ~ EOI}
+
+// Entry point for query "scanning"
+Scanner = _{ SOI ~ Keyword }
```
almann, line 10: Two different "starting" rules, we use `SOI` here to indicate start of input (to also trigger white space parsing).  Notice that the `Scanner` rule does not have repetitions or `EOI`, the idea is to make it more of an "iterator" style API.

partiql-parser/src/peg.rs
```diff
@@ -4,12 +4,63 @@
...
+                }
+                Ok(pair)
+            }
+            None => syntax_error(
+                "Expected at one token pair, got nothing!",
+                Position::Unknown,
+            ),
+        }
+    }
+}
```
almann, line 42: A successful parse from `Pest` returns `Pairs` which is an iterator of `Pair` which itself is a terrible name for a rule that matched some span of text (I believe the term "pair" alluding to the start and end of a match).

A rule match returns zero or more `Pair`, this API says, return an error unless it is exactly one `Pair`.

partiql-parser/src/peg.rs
```diff
@@ -4,12 +4,63 @@
...
+    #[inline]
+    fn start_loc(&self) -> LineAndColumn {
+        self.as_span().start_pos().line_col().into()
+    }
+
+    #[inline]
+    fn end_loc(&self) -> LineAndColumn {
+        self.as_span().end_pos().line_col().into()
+    }
+}
```
almann, line 63: Translates the baroque `Pair`/`Span` to start/end line positioning.

partiql-parser/src/peg.rs
```diff
@@ -18,7 +69,7 @@ struct PartiQLParser;
 ///
 /// This API will be replaced with one that produces an AST in the future.
 pub fn recognize_partiql(input: &str) -> ParserResult<()> {
-    PartiQLParser::parse(Rule::Keywords, input)?;
+    PartiQLParser::parse(Rule::Query, input)?;
```
almann, line 72: Better entry point name.

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+    /// ```
+    ///
+    /// ```
+    /// # use partiql_parser::prelude::*;
+    /// assert_eq!(
+    ///     LineAndColumn::at(21, 2),
+    ///     LineAndColumn::at(2, 15).move_relative(LineAndColumn::at(20, 2))
+    /// );
+    /// ```
+    pub fn move_relative(self, location: LineAndColumn) -> Self {
```
almann, line 66: This API is needed so we can use the PEG "continuation" style.  Call in, remember where we left off, and re-translate the parsers line/column positioning because every time we call into the PEG, it thinks we're starting at `(1, 1)`.

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+        }
+    }
+}
+
+impl From<(usize, usize)> for LineAndColumn {
+    fn from(line_and_column: (usize, usize)) -> Self {
+        let (line, column) = line_and_column;
+        Self::at(line, column)
+    }
+}
```
almann, line 88: This makes it so we can convert a raw tuple into our typed one by just saying something like `(x, y).into()`.

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+
+/// A possible position in the source.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum Position {
+    /// Variant indicating that there *is no* known location in source for some context.
     Unknown,
-    At { line: usize, column: usize },
+    /// Variant indicating that there *is* a known location in source for some context.
+    At(LineAndColumn),
```
almann, line 102: This is split out of the `enum`, because we have contexts where there is always a `LineAndColumn` and other contexts where there _may_ not be.

partiql-parser/src/scanner.rs
```diff
@@ -0,0 +1,270 @@
...
+    remainder: Remainder<'val>,
+}
+
+impl<'val> PartiQLScanner<'val> {
+    fn do_next_token(&mut self) -> ParserResult<Token<'val>> {
+        // the scanner rule is expected to return a single node
+        let pair: Pair<'val, Rule> =
+            PartiQLParser::parse(Rule::Scanner, self.remainder.input)?.exactly_one()?;
+        let start_location = self.remainder.offset.move_relative(pair.start_loc());
+        let end_location = self.remainder.offset.move_relative(pair.end_loc());
```
almann, line 114: Notice here we translate the parser which has a logical line/column from its start, to where we left off.

partiql-parser/src/scanner.rs
```diff
@@ -0,0 +1,270 @@
...
+
+impl<'val> Scanner<'val> for PartiQLScanner<'val> {
+    fn next_token(&mut self) -> ParserResult<Token<'val>> {
+        let start_loc = self.remainder.offset;
+        self.do_next_token().map_err(|e| match e {
+            ParserError::SyntaxError { message, position } => {
+                let position = match position {
+                    Position::Unknown => Position::Unknown,
+                    // make sure to translate line/column position from where we started
+                    Position::At(location) => Position::At(start_loc.move_relative(location)),
```
almann, line 149: Here, we translate the line information from our continuation.

almann:
I added a doc test to illustrate usage of the `Scanner` API, here it is rendered as an image

image

partiql-parser/src/scanner.rs
```diff
@@ -0,0 +1,270 @@
...
+/// A token can be thought of as a sort of continuation, it knows where it is in the input list
+/// and allows scanning to resume from it.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Token<'val> {
+    /// The underlying value of the token.
+    content: Content<'val>,
+    /// Start location for this token.
+    start_location: LineAndColumn,
+    /// End location for this token.
+    end_location: LineAndColumn,
```
dlurton, line 62: Should these be in a struct by themselves?

```
struct Span {
    /// Start location for this token.
    start_location: LineAndColumn,
    /// End location for this token.
    end_location: LineAndColumn,
}
```
??

almann: They could be--though the question is to what end in the API? Do you want to see `Token::span() -> Span`

almann: Added #14  to track this.

dlurton: `Token::span() -> Span` is the idea.

partiql-parser/src/lib.rs
```diff
@@ -2,10 +2,54 @@
...
+//!
+//! fn main() -> ParserResult<()> {
+//!     use partiql_parser::scanner::Content::*;
+//!
+//!     let mut scanner = scanner("SELECT FROM");
+//!     let first = scanner.next_token()?;
+//!
+//!     // get the parsed variant of the token
+//!     match first.content() {
+//!         Keyword(kw) => assert_eq!("SELECT", kw),
```
marcbowes, line 23: suggest you write this as `if let` as this code will fail doctest once we add more types.

partiql-parser/src/lib.rs
```diff
@@ -2,10 +2,54 @@
...
+//!     }
+//!     // the entire text of a token can be fetched--which looks the roughly the
+//!     // same for a keyword.
+//!     assert_eq!("SELECT", first.text());
```
marcbowes, line 27: was expecting `as_str()` here

partiql-parser/src/partiql.pest
marcbowes: maybe added the iterator bit to the comment

partiql-parser/src/peg.rs
```diff
@@ -4,12 +4,63 @@
...
+                    syntax_error(
+                        format!("Expected one token pair, got: {:?}, {:?}", pair, other_pair),
+                        pair.start_loc().into(),
+                    )?;
+                }
+                Ok(pair)
+            }
+            None => syntax_error(
+                "Expected at one token pair, got nothing!",
+                Position::Unknown,
```
marcbowes, line 38: FIXME?

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+    /// Constructs a [`LineAndColumn`].
+    #[inline]
+    pub fn at(line: usize, column: usize) -> Self {
+        Self(line, column)
+    }
+
+    /// Returns a [`LineAndColumn`] that moves to the given position relative
+    /// to this one as a sort of "origin."
+    ///
+    /// Note that this positioning is 1-based, so moving `(1, 1)` to `(1, 1)` is a no-op.
```
marcbowes, line 32: Why is the move 1-based? "Move 1 line" not moving 1 line seems weird

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+
+impl fmt::Display for LineAndColumn {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "line {}, column {}", self.0, self.1)
+    }
+}
+
+/// A possible position in the source.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub enum Position {
```
marcbowes, line 98: Do you want EOF?

marcbowes (approved):

partiql-parser/src/result.rs
almann: Maybe the name is not right--this isn't taking a delta, change `location` to use `self` as the origin.  I am not sure what I should call that, but `move_relative` is not right--let me think about that.

partiql-parser/src/result.rs
almann: Do you have a use case where distinguishing `EndOfFile` as a position is needed rather than the line/column that happens to be at the `EndOfFile`?  I do think we might need a `UnexpectedEndOfFile` type error case, but not sure about this one, an example might help.

partiql-parser/src/lib.rs
marcbowes: SymbolToken can *not* have a str, right (`$0`)? so maybe that's an important difference - that `text()` might not be meaningful. Whereas here it seems obvious that the keyword can be "viewed as a str".

I see your pov. though, and I agree it makes sense.

partiql-parser/src/peg.rs
marcbowes: sorry that was a lame comment!

is it always the case there is no position? I guess it could be none because you're at EOF or something..?

partiql-parser/src/result.rs
marcbowes: ah, ok

partiql-parser/src/result.rs
marcbowes: doing EOF in the error + the L+C makes sense!

partiql-parser/src/lib.rs
almann: BTW, not that this is canonical or anything, but I've been more or less inspired by:

https://rust-lang.github.io/api-guidelines/naming.html

I think what I have here is more like a *getter* versus a *conversion*:

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

>The difference between getters and conversions (C-CONV) can be subtle and is not always clear-cut. For example `TempDir::path` can be understood as a getter for the filesystem path of the temporary directory, while `TempDir::into_path` is a conversion that transfers responsibility for deleting the temporary directory to the caller. Since `path` is a getter, it would not be correct to call it `get_path` or `as_path`.

partiql-parser/src/peg.rs
almann: The only way this could happen is if I wrote a Pest grammar with a rule that matched something but it and all sub-rules were silent and did not produce a `Pair` at all.  This would mean that I as the rule writer wrote something pretty insane and I think it is almost certainly a bug.

Add this to the reason why I think Pest is a little too low-level for my tastes as far as its APIs are concerned (though I think logically consistent).

partiql-parser/src/result.rs
```diff
@@ -4,33 +4,140 @@
...
+        match (location, self) {
+            (LineAndColumn(base_line, base_column), LineAndColumn(dest_line, dest_column)) => {
+                let diff_line = dest_line - 1;
+                if diff_line > 0 {
+                    // we're moving lines, adjust the line and take the target column as-is
+                    LineAndColumn::at(base_line + diff_line, dest_column)
+                } else {
+                    // same line from base, adjust only the column
+                    let diff_column = dest_column - 1;
+                    LineAndColumn::at(base_line, base_column + diff_column)
```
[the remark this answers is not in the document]
almann: Promoted these comments into the doc tests.

partiql-parser/src/result.rs
almann: It is a good question, I am really gun shy around panics, but using an error hear also feels like overkill, I think the answer is probably error, but this is pretty close to where I'd use an assert in C.

partiql-parser/src/scanner.rs
```diff
@@ -0,0 +1,270 @@
...
+        self.end_location
+    }
+
+    /// Returns the slice of the input encompassing the entire token.
+    pub fn text(&self) -> &str {
+        self.text
+    }
+
+    /// Returns the remainder of the input after this token.
+    pub fn rem(&self) -> &str {
```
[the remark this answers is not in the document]
almann: It is an interesting question--I wasn't really planning on exposing `Remainder` as a public API.  How about

```rust
text_after(&self) -> &str
```

?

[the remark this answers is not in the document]
almann: Also, I will rename the various `start_loc` and `end_loc` to just `start`/`end`.

zslayton (approved):

marcbowes (approved):

partiql-parser/src/peg.rs
```diff
@@ -4,12 +4,63 @@
...
-use pest::Parser;
+use crate::result::syntax_error;
+use pest::iterators::{Pair, Pairs};
+use pest::{Parser, RuleType};
 use pest_derive::Parser;
 
 #[derive(Parser)]
 #[grammar = "partiql.pest"]
-struct PartiQLParser;
+pub(crate) struct PartiQLParser;
```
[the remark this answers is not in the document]
almann: Already done as #22

PR #15, opened 2021-05-12
Adds support for string literals and identifiers in `Scanner`.

almann:
Refactors the PEG to make each keyword and explicit rule that has the
form `KeywordName_`.  This allows us to easily refer to them in rules
without having to have magic strings.  The rationale for the `_` suffix
is to make sure it doesn't collide with the other grammar rules and not
have a wordy suffix like `Keyword` on every rule.

Also:
* Adds case folding to keywords.
* Makes error cases in the scanner/recognizer tests less brittle.
* Adds `Token` to the `Query` rule and updates tests for
  `recognize_partiql`.

Depends on #8 and #12--as such is targeting a staging branch but I did not want to prevent review.

Rendered update of the doc test that exercises some of this:

image

### DIFF
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -4,266 +4,551 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
 //      working with Pest and its APIs.
 
 // Entry point for full query parsing
-Query = _{ SOI ~ Keyword+ ~ EOI}
+Query = _{ SOI ~ Token+ ~ EOI}
 
 // Entry point for query "scanning"
 // Note that this is factored this way to support an iteration style API
 // where we can call back into this rule on subsequent input.
-Scanner = _{ SOI ~ Keyword }
+Scanner = _{ SOI ~ Token }
 
-Keyword = { AllKeywords }
-
-AllKeywords = _{
-      SqlKeywords
-    | PartiqlKeywords
+Token = _{
+      Keyword
+    | String
+    | Identifier
 }
 
-SqlKeywords = _{
-      ^"absolute"
-    | ^"action"
-    | ^"add"
-    | ^"all"
-    | ^"allocate"
-    | ^"alter"
-    | ^"and"
-    | ^"any"
-    | ^"are"
-    | ^"as"
-    | ^"asc"
-    | ^"assertion"
-    | ^"at"
-    | ^"authorization"
-    | ^"avg"
-    | ^"begin"
-    | ^"between"
-    | ^"bit"
-    | ^"bit_length"
-    | ^"by"
-    | ^"cascade"
-    | ^"cascaded"
-    | ^"case"
-    | ^"cast"
-    | ^"catalog"
-    | ^"char"
-    | ^"character"
-    | ^"character_length"
-    | ^"char_length"
-    | ^"check"
-    | ^"close"
-    | ^"coalesce"
-    | ^"collate"
-    | ^"collation"
-    | ^"column"
-    | ^"commit"
-    | ^"connect"
-    | ^"connection"
-    | ^"constraint"
-    | ^"constraints"
-    | ^"continue"
-    | ^"convert"
-    | ^"corresponding"
-    | ^"count"
-    | ^"create"
-    | ^"cross"
-    | ^"current"
-    | ^"current_date"
-    | ^"current_time"
-    | ^"current_timestamp"
-    | ^"current_user"
-    | ^"cursor"
-    | ^"date"
-    | ^"deallocate"
-    | ^"dec"
-    | ^"decimal"
-    | ^"declare"
-    | ^"default"
-    | ^"deferrable"
-    | ^"deferred"
-    | ^"delete"
-    | ^"desc"
-    | ^"describe"
-    | ^"descriptor"
-    | ^"diagnostics"
-    | ^"disconnect"
-    | ^"distinct"
-    | ^"domain"
-    | ^"double"
-    | ^"drop"
-    | ^"else"
-    | ^"end"
-    | ^"end-exec"
-    | ^"escape"
-    | ^"except"
-    | ^"exception"
-    | ^"exec"
-    | ^"execute"
-    | ^"exists"
-    | ^"external"
-    | ^"extract"
-    | ^"date_add"
-    | ^"date_diff"
-    | ^"false"
-    | ^"fetch"
-    | ^"first"
-    | ^"float"
-    | ^"for"
-    | ^"foreign"
-    | ^"found"
-    | ^"from"
-    | ^"full"
-    | ^"get"
-    | ^"global"
-    | ^"go"
-    | ^"goto"
-    | ^"grant"
-    | ^"group"
-    | ^"having"
-    | ^"identity"
-    | ^"immediate"
-    | ^"in"
-    | ^"indicator"
-    | ^"initially"
-    | ^"inner"
-    | ^"input"
-    | ^"insensitive"
-    | ^"insert"
-    | ^"int"
-    | ^"integer"
-    | ^"intersect"
-    | ^"interval"
-    | ^"into"
-    | ^"is"
-    | ^"isolation"
-    | ^"join"
-    | ^"key"
-    | ^"language"
-    | ^"last"
-    | ^"left"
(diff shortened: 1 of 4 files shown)


partiql-parser/src/partiql.pest
```diff
@@ -4,264 +4,549 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
 // Entry point for query "scanning"
-Scanner = _{ SOI ~ Keyword }
+Scanner = _{ SOI ~ Token }
 
-Keyword = { AllKeywords }
+Token = _{
+      Keyword
+    | String
+    | Identifier
+}
```
almann, line 16: Entry point updates.

partiql-parser/src/partiql.pest
```diff
@@ -4,264 +4,549 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+
+//
+// Individual Keyword Definitions
+//
+
+Absolute_ = { ^"absolute" }
+Action_ = { ^"action" }
+Add_ = { ^"add" }
+All_ = { ^"all" }
```
almann, line 313: This just breaks up every keyword into a rule with an easy to refer to name that should not collide with any real rules.

partiql-parser/src/peg.rs
```diff
@@ -76,20 +86,35 @@ pub fn recognize_partiql(input: &str) -> ParserResult<()> {
...
+                    position: expected_position,
+                    ..
+                }),
+                Err(ParserError::SyntaxError {
+                    position: actual_position,
+                    ..
+                }),
+            ) => {
+                // just compare the positions for syntax errors...
+                assert_eq!(expected_position, actual_position)
```
almann, line 112: This is so we don't have to match on the message.

partiql-parser/src/scanner.rs
```diff
@@ -9,15 +9,19 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum Content<'val> {
-    /// A PartiQL keyword.  Contains the slice for
-    Keyword(&'val str),
+    /// A PartiQL keyword.  Contains the slice for the keyword case folded to upper case.
+    Keyword(Cow<'val, str>),
+    /// An identifier.  Contains the slice for the text of the identifier.
+    Identifier(Cow<'val, str>),
+    /// A string literal.  Contains the slice for the content of the literal.
+    String(Cow<'val, str>),
```
almann, line 24: A `Cow` can be either an *owned* value (`String`) or a borrowed reference.  This gives us a bit more flexibility to use the raw input when we can, but construct a heap allocated object if we need to manipulate it.

https://doc.rust-lang.org/std/borrow/enum.Cow.html

partiql-parser/src/scanner.rs
```diff
@@ -68,8 +72,8 @@ pub struct Token<'val> {
 
 impl<'val> Token<'val> {
     /// Returns the parsed content of this token.
-    pub fn content(&self) -> Content<'val> {
-        self.content
+    pub fn content(&self) -> &Content<'val> {
+        &self.content
     }
```
almann, line 77: Can't return by value anymore because the `Cow` makes `Content` potentially expensive to do so.

partiql-parser/src/scanner.rs
```diff
@@ -119,13 +137,19 @@ impl<'val> PartiQLScanner<'val> {
...
+                let ident_pair = pair.into_inner().exactly_one()?;
+                match ident_pair.as_rule() {
+                    Rule::NonQuotedIdentifier => Content::Identifier(ident_pair.as_str().into()),
+                    Rule::QuotedIdentifier => {
+                        Content::Identifier(normalize_quoted_ident(ident_pair.as_str()))
+                    }
+                    _ => return ident_pair.syntax_error(),
+                }
             }
+            _ => return pair.syntax_error(),
```
almann, line 152: Here is where I am cleaning up the token for the `Content` value.  This will likely need to be exposed as an internal API for the full parser to leverage.

partiql-parser/src/scanner.rs
```diff
@@ -174,14 +198,15 @@ pub fn scanner(input: &str) -> PartiQLScanner {
...
+    use crate::result::syntax_error;
     use rstest::*;
 
     #[rstest]
     #[case::single_keyword(
         "  SELECT  ",
         vec![
             Ok(Token {
-                content: Content::Keyword("SELECT"),
+                content: Content::Keyword("SELECT".into()),
```
almann, line 209: Convert the string slice ref into a `Cow`.

partiql-parser/src/scanner.rs
```diff
@@ -258,9 +423,18 @@ mod test {
...
+                        position: expected_position,
+                        ..
+                    }),
+                    Err(ParserError::SyntaxError {
+                        position: actual_position,
+                        ..
+                    }),
+                ) => {
+                    // just compare the positions for syntax errors...
+                    assert_eq!(expected_position, actual_position);
```
almann, line 437: Similar to the other code--this probably suggests we should extract this out now that I think about it.

almann: Added #16 to track this.

partiql-parser/src/scanner.rs
marcbowes, line 24: What about calling this a `StringLiteral` instead of `String`? I read the type first (and was confused) and then the comment (and was un-confused).

marcbowes (approved):

partiql-parser/src/scanner.rs
almann: That seems reasonable to me--I will do that.

partiql-parser/src/scanner.rs
```diff
@@ -105,6 +109,20 @@ pub struct PartiQLScanner<'val> {
...
+#[inline]
+fn normalize_string_lit(raw_text: &str) -> Cow<str> {
+    raw_text[1..(raw_text.len() - 1)].replace("''", "'").into()
+}
+
+/// Removes PartiQL escapes from a quoted identifier and dequotes it.
+#[inline]
+fn normalize_quoted_ident(raw_text: &str) -> Cow<str> {
+    raw_text[1..(raw_text.len() - 1)]
+        .replace(r#""""#, r#"""#)
```
zslayton, line 122: Same thought here RE: avoiding a `String` copy when possible.

partiql-parser/src/scanner.rs
```diff
@@ -117,8 +135,19 @@ impl<'val> PartiQLScanner<'val> {
         self.remainder = self.remainder.consume(start_off + text.len(), pair.end()?);
 
         let content = match pair.as_rule() {
-            Rule::Keyword => Content::Keyword(text),
-            _ => return syntax_error(format!("Unexpected rule: {:?}", pair), pair.start()?.into()),
+            Rule::Keyword => Content::Keyword(text.to_uppercase().into()),
```
zslayton, line 138: Another `Cow<str>` opportunity: [`to_uppercase`](https://doc.rust-lang.org/std/primitive.str.html#method.to_uppercase) will also allocate even if the input is already all caps.

zslayton (commented):
There are a few places where we're not leveraging `Cow<str>`'s ability to avoid allocations/copies.

Not sure it's worth the dependency, but there's a small crate called [`cow-utils`](https://crates.io/crates/cow-utils) that provides [`cow_replace`](https://docs.rs/cow-utils/0.1.2/cow_utils/trait.CowUtils.html#tymethod.cow_replace), [`cow_to_uppercase`](https://docs.rs/cow-utils/0.1.2/cow_utils/trait.CowUtils.html#tymethod.cow_to_uppercase) etc.

almann:
> There are a few places where we're not leveraging `Cow<str>`'s ability to avoid allocations/copies.
> 
> Not sure it's worth the dependency, but there's a small crate called [`cow-utils`](https://crates.io/crates/cow-utils) that provides [`cow_replace`](https://docs.rs/cow-utils/0.1.2/cow_utils/trait.CowUtils.html#tymethod.cow_replace), [`cow_to_uppercase`](https://docs.rs/cow-utils/0.1.2/cow_utils/trait.CowUtils.html#tymethod.cow_to_uppercase) etc.

Thanks for the heads up on these I've added #28 to track this.

PR #19, opened 2021-05-14
Initial API for `Scanner`.

### LINKED ISSUE #13: Implement a `Scanner` Trait
I've recently opened #12 which adds this, but I wanted to have an issue to capture a bit more context about why this API is useful.

First, the goal of this API is *not* to proving lexing to the parser.  We are currently using [Pest][pest] for this and Parser Expression Grammars (PEGs) are scannerless in the sense that tokenization is part of the grammar definition.  Basically, the PEG is doing this, so a reasonably question might be "why bother?"

Here are the primary reasons I believe we should add this API:

* First, it will give us some experience abstracting over [Pest][pest].  This is a practical consideration and not an API one.  That said, the underlying `Pairs`/`Pair`/`RuleType`/`Token` API of [Pest][pest] is pretty low-level, and while logical, is most definitely something we don't want to burden our users with.
* Second, when talking to some potential customers of this API (e.g. @marcbowes), this could be useful for powering tools like the AWS QLDB shell where some amount of syntactic support besides a full parser is useful for the tool's UX.
* Thirdly, but related to the second point, this could be very useful in supporting a full blown PartiQL language server.  A WASM build of the parser via a `Scanner`/`Parser` that can be "continued" could be conceivably used in a editor like VS Code.

[pest]: https://pest.rs/

almann:
Provides an interface for _lexing_ PartiQL.  Traditionally we would
factor this as the low-level interface for the parser, but with the PEG
based implementation, we just surface the relevant parts of the parser
as a rule to parse with.

This commit helps explore how to effectively integrate with the somewhat
low-level APIs for parsing that Pest provides.

* Adds `scanner` module.
* Adds `Scanner` trait to the prelude.
* Makes `PartiQLParser` public to crate.
* Refactors `LineAndColumn` as a tuple for `Position::At`.
  - Adds this to the `prelude`.
* Changes entry point for PEG to `Query` and added `Scanner` as the
  entry point rules for implementing the `Scanner` API.
* Adds `PairsExt`/`PairExt` trait/impl to add utility methods for working
  with Pest `Pairs`/`Pair`.
* Adds `LineAndColumn::position_from` and cleans up some doc/doc tests.

Resolves #13.

### DIFF
--- a/partiql-parser/src/lib.rs
+++ b/partiql-parser/src/lib.rs
@@ -2,10 +2,54 @@
 
 //! Provides a parser for the [PartiQL][partiql] query language.
 //!
+//! # Usage
+//!
+//! An API to interact with PartiQL tokens is the [`mod@scanner`] module.
+//! The [`scanner()`] function creates a [`Scanner`](scanner::Scanner) instance
+//! that one can use to parse tokens incrementally from some input slice.
+//!
+//! ```
+//! use partiql_parser::prelude::*;
+//! use partiql_parser::scanner;
+//!
+//! fn main() -> ParserResult<()> {
+//!     use partiql_parser::scanner::Content::*;
+//!
+//!     let mut scanner = scanner("SELECT FROM");
+//!     let first = scanner.next_token()?;
+//!
+//!     // get the parsed variant of the token
+//!     match first.content() {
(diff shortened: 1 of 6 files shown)


almann:
Closing as I can change the base of the previous PR.

PR #20, opened 2021-05-18
Adds numeric literals to grammar.

almann:
Specifically adds the `Number` rule for parsing decimals and integers.
Also adds the `IntegerLiteral` and `DecimalLiteral` to capture the
respective literal tokens.

Notes:
* Adds `num-traits`, `num-bigint`, and `bigdecimal` to provide support
  for integer/decimal parsing.
* Adds `PairExt::unexpected()` for dealing with unexpected rules.
* Adds `PairExt::syntax_error()` for easily creating syntax errors from a pair.
* Updates scanner doc test to `panic!` for unexpected tokens.
* Fixes the bad recognizer test because it only has scanner support and we
  cannot detect that `99_RANCH` is syntactically incorrect but lexes to
  `99` and `_RANCH`.  Real PartiQL parsing will fix this in the future.

### DIFF
--- a/partiql-parser/Cargo.toml
+++ b/partiql-parser/Cargo.toml
@@ -23,6 +23,9 @@ version = "0.0.0"
 pest = "~2.1.3"
 pest_derive = "~2.1.0"
 thiserror = "~1.0.24"
+num-traits = "~0.2.14"
+num-bigint = "~0.4.0"
+bigdecimal = "~0.2.0"
 
 [dev-dependencies]
 rstest = "~0.9.0"
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -14,8 +14,38 @@ Scanner = _{ SOI ~ Token }
 Token = _{
       Keyword
     | String
+    | Number
     | Identifier
 }
+
+//
+// Numeric Literals
+//
+
+Number = ${
+      DecimalExp
+    | Decimal
+    | Integer
+}
+
+DecimalExp = {
+    Decimal ~ ("e" | "E") ~ Integer
+}
+
+Decimal = {
+    Integer? ~ "." ~ Fraction
+}
+
+Fraction = {
+    Digit*
+}
+
+Integer = {
+   Sign? ~ Digit+
+}
+
+Sign = _{ "+" | "-" }
+Digit = _{ '0'..'9' }
 
 //
 // String Literals
@@ -40,7 +70,7 @@ NonQuotedIdentifier = @{
 
 NonQuotedIdentifierStart = _{ "_" | "$" | 'a'..'z' | 'A'..'Z' }
 
-NonQuotedIdentifierCont = _{ NonQuotedIdentifierStart | '0'..'9' }
+NonQuotedIdentifierCont = _{ NonQuotedIdentifierStart | Digit }
 
 QuotedIdentifier = @{ "\"" ~ QuotedIdentifierContent* ~ "\"" }
 
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -40,6 +40,9 @@ Fraction = {
     Digit*
 }
 
+// XXX this explicitly supports arbitrary zero prefixing in various places
+//     which is part of the SQL grammar and also supported in implementations
+//     like Postgres/SQLite/MySQL/etc.
 Integer = {
    Sign? ~ Digit+
 }
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -46,8 +46,11 @@ pub(crate) trait PairExt<'val, R: RuleType> {
     /// Translates the end position of the [`Pair`] into a [`LineAndColumn`].
     fn end(&self) -> ParserResult<LineAndColumn>;
 
-    /// Returns an `Err` with a syntax error from the pair
-    fn syntax_error<T>(&self) -> ParserResult<T>;
+    /// Returns an `Err` with a syntax error from the unexpected pair.
+    fn unexpected<T>(&self) -> ParserResult<T>;
+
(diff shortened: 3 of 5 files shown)


partiql-parser/Cargo.toml
```diff
@@ -23,6 +23,9 @@ version = "0.0.0"
 pest = "~2.1.3"
 pest_derive = "~2.1.0"
 thiserror = "~1.0.24"
+num-traits = "~0.2.14"
+num-bigint = "~0.4.0"
+bigdecimal = "~0.2.0"
```
almann, line 28: `num-traits` is explicitly pulled in to get `from_str_radix` which is in the `Num` trait.

partiql-parser/src/partiql.pest
```diff
@@ -14,9 +14,39 @@ Scanner = _{ SOI ~ Token }
...
 
+//
+// Numeric Literals
+//
+
+Number = ${
+      DecimalExp
+    | Decimal
+    | Integer
+}
```
almann, line 29: The `$` means *compound atomic*.  Which means that sub-rules can produce pairs, but whitespace does not apply, which we definitely don't want in number parsing.

See: https://docs.rs/pest_derive/2.1.0/pest_derive/#modifiers

partiql-parser/src/scanner.rs
```diff
@@ -123,6 +137,18 @@ fn normalize_quoted_ident(raw_text: &str) -> Cow<str> {
...
+where
+    T: Num<FromStrRadixErr = E>,
+    R: RuleType,
+    E: std::fmt::Display,
+{
+    match T::from_str_radix(pair.as_str(), 10) {
+        Ok(value) => Ok(value),
+        Err(e) => pair.syntax_error(format!("Could not parse number {}: {}", pair.as_str(), e)),
+    }
+}
```
almann, line 150: This just makes it easy to parse and translate errors to syntax error from the `Pair` that represents the token for the number, whether it is a decimal or integer.

partiql-parser/src/scanner.rs
```diff
@@ -144,10 +170,20 @@ impl<'val> PartiQLScanner<'val> {
...
+                    _ => return ident_pair.unexpected(),
+                }
+            }
+            Rule::Number => {
+                let number_pair = pair.into_inner().exactly_one()?;
+                match number_pair.as_rule() {
+                    Rule::Integer => Content::IntegerLiteral(parse_num(number_pair)?),
+                    Rule::Decimal | Rule::DecimalExp => {
+                        Content::DecimalLiteral(parse_num(number_pair)?)
+                    }
```
almann, line 182: This is where we use that generic `parse_num` to fill in the literal, we leverage type inference to select `BigInt` and `BigDecimal` as appropriate.

partiql-parser/src/scanner.rs
```diff
@@ -349,6 +385,174 @@ mod test {
...
+            Ok(Token {
+                content: Content::IntegerLiteral(1.into()),
+                start: LineAndColumn::at(1, 1),
+                end: LineAndColumn::at(1, 2),
+                text: "1",
+                remainder: Remainder {
+                    input: " -0099 1.1 +00055.023100 99.1234e0010",
+                    offset: LineAndColumn::at(1, 2)
+                }
+            }),
```
almann, line 400: I think it is getting time to add a macro to make these cases easier to write...

almann: Tracking as #21.

partiql-parser/src/partiql.pest
```diff
@@ -14,9 +14,39 @@ Scanner = _{ SOI ~ Token }
...
+
+DecimalExp = {
+    Decimal ~ ("e" | "E") ~ Integer
+}
+
+Decimal = {
+    Integer? ~ "." ~ Fraction
```
zslayton, line 36: This will accept some decimals that aren't legal Ion (like `00000.0`, multiple leading zeros) and it will reject others that are legal Ion (like `42.`, no trailing digits). I don't know if PartiQL has different syntax for decimal literals than Ion does or if this is just a first pass that we'll iron out in a future commit, but I wanted to mention it just in case.

partiql-parser/src/peg.rs
```diff
@@ -64,8 +67,12 @@ impl<'val, R: RuleType> PairExt<'val, R> for Pair<'val, R> {
...
     }
 
-    fn syntax_error<T>(&self) -> ParserResult<T> {
-        syntax_error(format!("Unexpected rule: {:?}", self), self.start()?.into())
+    fn unexpected<T>(&self) -> ParserResult<T> {
+        self.syntax_error(format!("Unexpected rule: {:?}", self))
+    }
+
+    fn syntax_error<T, S: Into<String>>(&self, message: S) -> ParserResult<T> {
+        syntax_error(message, self.start()?.into())
```
zslayton, line 75: If `self.start()` returns `Err`, that's what will bubble up instead of the actual syntax error. Is that possible/likely? At a glance I'd think we would want something like:
```rust
// Handle the bad location case separately somehow
let position = self.start().ok_or_else(|| Position::Unknown).into();
// Always return the result from `syntax_error`
syntax_error(message, position)
```

partiql-parser/src/partiql.pest
almann: Yes, according to the SQL spec things like `0000012`, `.001`, `000.001e000010` are all allowed and I have verified that I didn't just port the bad behavior from the original PartiQL parser...

I will add a comment that is explicit about this, also note that I have explicit positive tests for these cases as well.

partiql-parser/src/peg.rs
almann: This is definitely a better mapping of a bad position--will change.

zslayton (approved):

Issue #23, opened 2021-05-26
Add `IllegalState` Error


almann:
We have internal errors that are not syntax or `InvalidArg`, we should add some sort of `IllegalState` error enum.

---

This is probably better modeled as illegal state.

_Originally posted by @almann in https://github.com/partiql/partiql-lang-rust/pull/12#discussion_r635494493_

Issue #18, opened 2021-05-12
Refactor `ParserError` to be a `struct`.


almann:
Currently our `ParserError` is an `enum` with a single `SyntaxError` variant:

https://github.com/partiql/partiql-lang-rust/blob/4aa104c65ee7b69eca967c10e823016e00c91493/partiql-parser/src/result.rs#L130-L136

More likely than not, we all `ParserErrors` will have a message with optional position information, we should refactor the variant into some kind of `ParserErrorType` or something that could be as simple as just the tag (e.g. `SyntaxError`), but could also have error specific data.

See also #16, no matter what we do here, we'll like a utility method to do non-message equality.

PR #24, opened 2021-05-28
First issue

tbmreza:
*Issue #, if available:* #23

*Description of changes:* Adds a unit variant `IllegalState` to ParserError enum. [Users](https://github.com/partiql/partiql-lang-rust/pull/12#discussion_r635494493) of the enum would call `illegal_state()` convenience function.

### DIFF
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -4,7 +4,8 @@
 //! can be exported for users to consume.
 
 use crate::prelude::*;
-use crate::result::syntax_error;
+// use crate::result::syntax_error;
+use crate::result::{illegal_state, syntax_error};
 use pest::iterators::{Pair, Pairs};
 use pest::{Parser, RuleType};
 use pest_derive::Parser;
@@ -33,10 +34,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 }
                 Ok(pair)
             }
-            None => syntax_error(
-                "Expected at one token pair, got nothing!",
-                Position::Unknown,
-            ),
+            None => illegal_state(),
         }
     }
 }
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -4,7 +4,6 @@
 //! can be exported for users to consume.
 
 use crate::prelude::*;
-// use crate::result::syntax_error;
 use crate::result::{illegal_state, syntax_error};
 use pest::iterators::{Pair, Pairs};
 use pest::{Parser, RuleType};
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -33,7 +33,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 }
                 Ok(pair)
             }
-            None => illegal_state("Expected at one token pair, got nothing!"),
+            None => illegal_state("Expected at least one token pair, got nothing!"),
         }
     }
 }
--- a/partiql-parser/src/peg.rs
+++ b/partiql-parser/src/peg.rs
@@ -33,7 +33,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 }
                 Ok(pair)
             }
-            None => illegal_state(),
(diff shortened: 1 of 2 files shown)


partiql-parser/src/peg.rs
```diff
@@ -33,10 +33,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 }
                 Ok(pair)
             }
-            None => syntax_error(
-                "Expected at one token pair, got nothing!",
-                Position::Unknown,
-            ),
+            None => illegal_state(),
```
marcbowes, line 36: Did you consider keeping the message so that the error comes with some context?

marcbowes (commented):
Thanks for the contribution! I have a comment that I think should be addressed.

partiql-parser/src/result.rs
```diff
@@ -190,6 +190,10 @@ pub enum ParserError {
     /// Indicates that there is a problem with run-time violation of some API.
     #[error("Invalid Argument: {message}")]
     InvalidArgument { message: String },
+
+    /// Indicates that there is an internal error that isn't SyntaxError nor InvalidArgument.
+    #[error("Illegal State")]
+    IllegalState,
```
almann, line 196: I think following the pattern above to have a `message` field would be a good idea.

I do think in the future we'll make this a bit more factored out as part of #18, but for now the slightly duplicative `{ message: String }` pattern is reasonable.

partiql-parser/src/result.rs
```diff
@@ -190,6 +190,10 @@ pub enum ParserError {
     /// Indicates that there is a problem with run-time violation of some API.
     #[error("Invalid Argument: {message}")]
     InvalidArgument { message: String },
+
+    /// Indicates that there is an internal error that isn't SyntaxError nor InvalidArgument.
```
almann, line 194: Minor suggested edit for the doc comment:

```suggestion
    /// Indicates that there is an internal error that was not due to user input or API violation.
```

almann (commented):
Overall, this is pretty close, I do think `IllegalState` should have a message field (similar to `InvalidArgument`, and that would give you the capability to address the feedback from @marcbowes.

partiql-parser/src/peg.rs
tbmreza: Yeah, the problem statement was a bit unclear to me. My understanding is this:

The None arm is practically impossible to reach, it's an _illegal_ state. Therefore, we could omit the (presumably unhelpful) message and the (unknown) position information and dedicate a unit variant for this state.

Should I keep the message ("Expected at one token pair, got nothing!")?

tbmreza:
Thank you for the review. I have committed the suggested changes.

partiql-parser/src/peg.rs
```diff
@@ -33,10 +33,7 @@ impl<'val, R: RuleType> PairsExt<'val, R> for Pairs<'val, R> {
                 }
                 Ok(pair)
             }
-            None => syntax_error(
-                "Expected at one token pair, got nothing!",
-                Position::Unknown,
-            ),
+            None => illegal_state("Expected at one token pair, got nothing!"),
```
marcbowes, line 36: Not your fault, but the message doesn't seem right. Should it be "Expected at _least_ one" or "Expected ~~at~~ one"?

partiql-parser/src/peg.rs
almann: This one is something I probably should've made clearer.  It is something that almost rises to the assertion level, but from a utility perspective could happen if you create a Pest rule that is *silent* and matches will return such an *empty* `Pairs`.

```rust
use pest::iterators::Pairs;
use pest::Parser;
use pest_derive::Parser;

#[derive(Parser)]
#[grammar_inline = "ab = _{ \"ab\" }"]
struct FooParser;

fn main() {
    let pairs: Pairs<Rule> = FooParser::parse(Rule::ab, "ab").unwrap();
    assert_eq!(0, pairs.count());
}
```

partiql-parser/src/peg.rs
almann: This one is my fault, I can clean it up.

```suggestion
            None => illegal_state("Expected at least one token pair, got nothing!"),
```

almann (approved):
:rocket:

marcbowes (approved):

PR #25, opened 2021-05-28
Streamlines the `Scanner` Tests.

### LINKED ISSUE #21: Add Macro for Scanner Test Cases
The test cases for the `Scanner` are getting verbose to write, we should consider adding a macro to make this more streamlined.

---

I think it is getting time to add a macro to make these cases easier to write...

_Originally posted by @almann in https://github.com/partiql/partiql-lang-rust/pull/20#discussion_r634055915_

almann:
Previously, the `Scanner` tests required a lot of manual position
expectations which was tedious to write.  This replaces that with a test
harness that can do the math from chunks of input that the test writer
defines as whitespace or token boundaries.  This also adds the macro
`scanner_test_case!` which makes it so that one can use the harness with
a friendly syntax. The old tests are ported to the new macro.

Also:
* Adds assertions to the internal `LineAndColumn::at`.
* Adds convenience constructor around `Content`.

Resolves #21.

### DIFF
--- a/partiql-parser/src/scanner.rs
+++ b/partiql-parser/src/scanner.rs
@@ -379,20 +379,20 @@ mod test {
         }};
         // termination case -- no more chunks to process
         (@inner $test_case:ident) => {};
-        // final whitespace without a terminating ',' -- just delegate to the general form
+        // final whitespace chunk without a terminating ',' -- just delegate to the general form
         (@inner $test_case:ident $lit:literal) => {
             scanner_test_case!(@inner $test_case $lit ,)
         };
-        // final whitespace without a terminating ',' -- just delegate to the general form
+        // final token chunk without a terminating ',' -- just delegate to the general form
         (@inner $test_case:ident $lit:literal => $expr:expr) => {
             scanner_test_case!(@inner $test_case $lit => $expr ,)
         };
-        // add whitespace for a chunk and continue processing
+        // add whitespace for a chunk of string and continue...
         (@inner $test_case:ident $lit:literal , $($tail:tt)*) => {
             $test_case.add_text($lit);
             scanner_test_case!(@inner $test_case $($tail)*)
         };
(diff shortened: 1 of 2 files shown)


partiql-parser/src/scanner.rs
```diff
@@ -238,371 +263,281 @@ mod test {
...
-            syntax_error("IGNORED MESSAGE", Position::at(1, 25)),
+        scanner_test_case![
+            "SelEct" => keyword("SELECT"),
+            " ",
+            "'✨✨✨'" => string_literal("✨✨✨"),
+            " ",
+            "fROM" => keyword("FROM"),
+            " ",
+            r#""┬─┬""# => identifier("┬─┬"),
+            " "
```
marcbowes, line 493: (Picking one case at random.)

What is the value in specifying the input *and* output (stuff on the left?) e.g.

```
"'✨✨✨'" => string_literal("✨✨✨"),
```

the plain literals make sense to me, but the pairs don't.

Like, for example, why not just have:

```
string_literal("✨✨✨"),
```

and, in the macro, you can transform that to the pairs. Nice thing about that is you could generate multiple test cases, e.g. adding whitepace or changing casing or something..

marcbowes: btw. not saying you need to make your fancy macro even more fancy. just asking if I'm missing some specific point to the apparent duplication.

marcbowes (approved):

partiql-parser/src/scanner.rs
almann: The main reason why I didn't add that sophistication is I didn't want to start auto-generating lexical syntax--I still wanted that to be driven by the test writer.  Specifically, there are different ways to write things like identifiers so that could get complicated fast.  For other literals like integers, the lexical content could be very different than the logical form, and I wanted that to be easy to test (all the zero pad stuff for example).

Issue #17, opened 2021-05-12
Add `COMMENT` Rule in Pest Grammar


almann:
Support PartiQL's comments.

PR #26, opened 2021-05-29
Add COMMENT to Scanner grammar rule

tbmreza:
*Issue:* 
#17 

*Description of changes:* 
I added COMMENT rule to the pest grammar. I was hesitant to edit the Rule::Query as well. Should I?

### DIFF
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -1,8 +1,9 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
 
+COMMENT = _{ LineComment | BlockComment }
+
 BlockComment = { "/*" ~ ( BlockComment | (!"*/" ~ ANY) )* ~ "*/" }
-UntilEolComment = { "--" ~ ( !"\n" ~ ANY )* ~ "\n" }
-COMMENT = _{ UntilEolComment | BlockComment }
+LineComment = { "--" ~ ( !"\n" ~ ANY )* ~ "\n" }
 
 // TODO implement a full grammar, this is a very primitive version to start
 //      working with Pest and its APIs.
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -1,4 +1,5 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
+COMMENT = _{ "--" ~ ANY* }
 
 // TODO implement a full grammar, this is a very primitive version to start
 //      working with Pest and its APIs.
@@ -9,7 +10,7 @@ Query = _{ SOI ~ Token+ ~ EOI}
 // Entry point for query "scanning"
 // Note that this is factored this way to support an iteration style API
 // where we can call back into this rule on subsequent input.
-Scanner = _{ SOI ~ Token }
+Scanner = _{ SOI ~ !COMMENT ~ Token }
 
 Token = _{
       Keyword
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -1,5 +1,8 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
-COMMENT = _{ "--" ~ ANY* }
+
+BlockComment = { "/*" ~ ( BlockComment | (!"*/" ~ ANY) )* ~ "*/" }
+UntilEolComment = { "--" ~ ( !"\n" ~ ANY )* ~ "\n" }
+COMMENT = _{ UntilEolComment | BlockComment }
 
 // TODO implement a full grammar, this is a very primitive version to start
 //      working with Pest and its APIs.
@@ -10,7 +13,7 @@ Query = _{ SOI ~ Token+ ~ EOI}
 // Entry point for query "scanning"
 // Note that this is factored this way to support an iteration style API
 // where we can call back into this rule on subsequent input.
-Scanner = _{ SOI ~ !COMMENT ~ Token }
+Scanner = _{ SOI ~ Token }
 
 Token = _{
       Keyword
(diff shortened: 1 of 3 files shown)


tbmreza:
I'll need to update this after #25 is merged.

partiql-parser/src/partiql.pest
```diff
@@ -9,7 +10,7 @@ Query = _{ SOI ~ Token+ ~ EOI}
 // Entry point for query "scanning"
 // Note that this is factored this way to support an iteration style API
 // where we can call back into this rule on subsequent input.
-Scanner = _{ SOI ~ Token }
+Scanner = _{ SOI ~ !COMMENT ~ Token }
```
almann, line 13: You should not need to have a negative assertion for this, I am guessing this is here because of the `ANY*` rule above.

There is an interesting API question as to whether the scanner should return comments which would help for use cases like syntax highlighting.  For now I think they should be supported as white space (i.e. ignored by the scanner). we can revisit this later.

almann (changes requested):
Thanks for taking on this issue.  Specifically I think we need a fix to the single line comment definition and add nested block comments (as per SQL-99).

Note that for `COMMENT` and `WHITESPACE` we don't need special rules for dealing with them in [non-atomic contexts][1].

[1]: https://docs.rs/pest_derive/2.1.0/pest_derive/#whitespace-and-comment

almann:
> I'll need to update this after #25 is merged.

This has been merged.  A bit of rework will be required as you pointed out, but hopefully the tests are easier to write now.

partiql-parser/src/partiql.pest
```diff
@@ -1,4 +1,5 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
+COMMENT = _{ "--" ~ ANY* }
```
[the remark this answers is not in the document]
tbmreza: Whoops, forgive my lack of research. I only looked for "comment" in the getting started page and proceeded with it. 

Got it, I'll consult the kotlin impl.

tbmreza:
I believe it's all good now.

partiql-parser/src/partiql.pest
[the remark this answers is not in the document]
almann: No worries, I should've put these references in the issue.

I will say, we should be a bit careful about just using the Kotlin implementation for the block comment:

https://github.com/partiql/partiql-lang-kotlin/blob/ce06bcf8203ea963596fc723e0bb7c5d1b422dfd/lang/src/org/partiql/lang/syntax/SqlLexer.kt#L378-L386

This code appears to *not* support nested block comments which I believe is part of the SQL specification.  It isn't the end of the world if we make this a TODO in the Rust grammar, but something to consider when using that implementation as an example.

partiql-parser/src/scanner.rs
```diff
@@ -399,6 +399,63 @@ mod test {
...
+    #[case::comment_block_nested(
+        scanner_test_case![
+            "employee" => identifier("employee"),
+            " /* ",
+            "CASE",
+            " /* ",
+            "WHERE",
+            " */ ",
+            "employee",
+            " */ ",
```
almann, line 455: Would be good to test a couple interesting cases here, some extra newline characters in the mix is a good idea, bit also unbalanced comments as an error case (see `bad_tokens` for a place where the unbalanced comments could go to be tested).

partiql-parser/src/partiql.pest
```diff
@@ -1,5 +1,9 @@
 WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
 
+BlockComment = { "/*" ~ ( BlockComment | (!"*/" ~ ANY) )* ~ "*/" }
+UntilEolComment = { "--" ~ ( !"\n" ~ ANY )* ~ "\n" }
+COMMENT = _{ UntilEolComment | BlockComment }
```
almann, line 5: Minor style suggestion edit, the rules in this file typically go from top to bottom, so this just re-ordered them to be consistent, and renames `UntilEolComment` to `LineComment`.

```suggestion
COMMENT = _{ LineComment | BlockComment }

BlockComment = { "/*" ~ ( BlockComment | (!"*/" ~ ANY) )* ~ "*/" }
LineComment = { "--" ~ ( !"\n" ~ ANY )* ~ "\n" }
```

almann (commented):
This looks pretty good!  I have some minor suggestions and one in particular around testing that I think is worth doing and I think we can :ship: this.

tbmreza:
I have added a few unbalanced comment cases that I can think of.

partiql-parser/src/scanner.rs
```diff
@@ -589,6 +589,11 @@ mod test {
 
     #[rstest]
     #[case::bad_identifier("💩")]
+    #[case::unterminated_line_comment("-- DROP")]
+    #[case::unbalanced_block_comment("/*\n\n SELECT /* WHERE */")]
+    #[case::unbalanced_block_comment("/* CASE do WHEN re THEN mi ELSE fa END /*")]
+    #[case::unbalanced_block_comment("/*SELECT /* FROM /* FULL OUTER JOIN */ */ ")]
+    #[case::unbalanced_block_comment("/*/*/*/*/*/*/*/*[ascii art here]*/*/*/*/*/*/*/ ")]
```
almann, line 596: Minor edits to the test case names:

```suggestion
    #[case::unbalanced_block_comment_nested("/*\n\n SELECT /* WHERE */")]
    #[case::unbalanced_block_comment_end_dangling("/* CASE do WHEN re THEN mi ELSE fa END /*")]
    #[case::unbalanced_block_comment_nested_open2("/*SELECT /* FROM /* FULL OUTER JOIN */ */ ")]
    #[case::unbalanced_block_comment_deeply_nested("/*/*/*/*/*/*/*/*[ascii art here]*/*/*/*/*/*/*/ ")]
```

almann (approved):
This looks great!

partiql-parser/src/scanner.rs
tbmreza: Reasonable naming to me, thanks.

almann: Thanks--I neglected to apply this suggestion, so I opened this change in #27.

PR #30, opened 2021-06-02
Adds the scanner rules for punctuation

almann:
Adds the Pest grammar rules for these tokens and particularly
adds the lookahead assertions for the various edge cases around the
punctuation tokens that interact with each other and comments.

Adds special `Content` enum variants for `.`/`*`/`?` and a basic variant
for `Operator` and `Delimiter`.

This should the final pre-requisite for all of the terminal parse rules in
the PEG and allow us to start adding the parser rules for expressions.

An explicit TODO is around modeling the various operators as their own
enum or enum variants, right now they are returned as normalized string
content.

### DIFF
--- a/partiql-parser/src/scanner.rs
+++ b/partiql-parser/src/scanner.rs
@@ -573,6 +573,7 @@ mod test {
             "0.0e000" => decimal_literal_from_str("0.0")
         ]
     )]
+    #[case::no_trailing_zeros(scanner_test_case!["1231231." => decimal_literal_from_str("1231231")])]
     #[case::delimiters(
         scanner_test_case![
             "[" => delimiter("["),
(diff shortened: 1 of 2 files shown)


almann:
> An explicit TODO is around modeling the various operators as their own
> enum or enum variants, right now they are returned as normalized string
> content.

See #31 for details on this TODO.

zslayton (approved):

partiql-parser/src/scanner.rs
```diff
@@ -208,6 +243,11 @@ impl<'val> PartiQLScanner<'val> {
                     _ => return number_pair.unexpected(),
                 }
             }
+            Rule::Dot_ => Content::Dot,
+            Rule::Star_ => Content::Star,
+            Rule::Parameter => Content::Parameter,
```
alancai98, line 248: Is there a reason tests haven't been added for Dot, Star, and Parameter?

almann: Good catch, got so caught up in my other punctuation tests especially around potentially ambiguous cases that I forgot. will add.

alancai98 (approved):

PR #33, opened 2021-06-04
Fills in the start of a real PartiQL grammar.

almann:
The grammar breaks out `SelectExpression` from `ValueExpression` but
unifies them with grouping.  This is actually not supported in the
Kotlin parser, but is actually the correct form in my opinion.

Also cleans up atomicity/silencing in the dependent terminal rules and
while doing this, the spacing was adjusted to make this easier to read.
More specifically, all punctuation is basically silent because grammar rules
don't want them to be in the parse, and keywords are also silent because
the grammar enforces the structure.  Operators on the other hand are
atomic and visible because they are used in the parse tree to drive
precedence climbing/Pratt processing and are thus required to be
visible.

This is clearly not a complete grammar, nor are the tests sufficient,
but this is a good starting place to build in that structure.

Some explicit TODOs here:

* The special operators/binary/unary operators.
* The various `SELECT` projection forms:
  - `SELECT *`/`SELECT` list
* Joins in the `FROM` clause
* Various other clauses `GROUP BY`/`HAVING`/`ORDER BY` (set
  expressions).

### DIFF
--- a/partiql-parser/src/partiql.pest
+++ b/partiql-parser/src/partiql.pest
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
 
 COMMENT = _{ LineComment | BlockComment }
 
-BlockComment = { "/*" ~ (BlockComment | (!"*/" ~ ANY))* ~ "*/" }
-LineComment = { "--" ~ (!"\n" ~ ANY )* ~ "\n" }
+BlockComment    = _{ "/*" ~ (BlockComment | (!"*/" ~ ANY))* ~ "*/" }
+LineComment     = _{ "--" ~ (!"\n" ~ ANY )* ~ "\n" }
 
 // TODO implement a full grammar, this is a very primitive version to start
 //      working with Pest and its APIs.
 
-// Entry point for full query parsing
-Query = _{ SOI ~ Token+ ~ EOI}
+//
+// Entry point for full query parsing.
+//
+Query = _{ SOI ~ Expression ~ EOI}
 
-// Entry point for query "scanning"
+//
+// Expressions
+//
+
+Expression = _{
+      SelectExpression
+    | ValueExpression
+}
+
+// TODO finish the SELECT statement
+
+SelectExpression = {
+    Select ~ (From ~ (Where)?)?
+}
+
+Select = _{
+    SelectValue
+}
+
+SelectValue = {
+    Select_ ~ Value_ ~ ValueExpression
+}
+
+// TODO joins
+
+From = {
+    From_ ~ ValueExpression ~ (As_)? ~ (Identifier)?
+}
+
+Where = {
+    Where_ ~ ValueExpression
+}
+
+// TODO all the operators / paths
+
+ValueExpression = {
+      Term
+    | LeftParen_ ~ Expression ~ RightParen_
+}
+
+// TODO fill in all the atoms
+
+Term = _{
+      NullLiteral
+    | BooleanLiteral
+    | String
+    | Number
+    | Identifier
+    | ArrayConstructor
+    | BagConstructor
+    | TupleConstructor
+}
+
+ArrayConstructor = {
+    LeftBracket_ ~ Values? ~ RightBracket_
+}
+
+BagConstructor = {
+    LeftDoubleAngle_ ~ Values? ~ RightDoubleAngle_
+}
+
+Values = _{
+    (ValueExpression ~ (Comma_ ~ ValueExpression)* ~ Comma_?)
+}
+
+TupleConstructor = {
+    LeftCurly_ ~ TupleField ~ (Comma_ ~ TupleField)* ~ Comma_? ~ RightCurly_
+}
+
+TupleField = {
+    ValueExpression ~ Colon_ ~ ValueExpression
+}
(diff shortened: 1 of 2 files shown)


partiql-parser/src/partiql.pest
```diff
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+ArrayConstructor = {
+    LeftBracket_ ~ Values? ~ RightBracket_
+}
+
+BagConstructor = {
+    LeftDoubleAngle_ ~ Values? ~ RightDoubleAngle_
+}
+
+Values = _{
+    (ValueExpression ~ (Comma_ ~ ValueExpression)* ~ Comma_?)
```
dlurton, line 78: This would seem to support trailing commas?  Not sure that's intentional.

partiql-parser/src/partiql.pest
```diff
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+}
+
+TupleConstructor = {
+    LeftCurly_ ~ TupleField ~ (Comma_ ~ TupleField)* ~ Comma_? ~ RightCurly_
```
dlurton, line 82: Trailing commas here too... It probably was intentional, but still worth checking.

dlurton (approved):
Nothing too terribly controversial here I suppose

One question, however: did you intend to support trailing commas in the grammar or am I misreading it?

partiql-parser/src/partiql.pest
almann: It is--does this contradict what we did in the Kotlin grammar?

I do like this and I think we support it, but to be honest, I wasn't sure if we actually implemented it.

almann: Went and looked it up, we don't support it, I can remove this for now.

https://github.com/partiql/partiql-lang-kotlin/blob/7ac588979c68f33155b3df0fd37f01babc65f84f/lang/src/org/partiql/lang/syntax/SqlParser.kt#L2802-L2832

partiql-parser/src/partiql.pest
```diff
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+
+Expression = _{
+      SelectExpression
+    | ValueExpression
+}
+
+// TODO finish the SELECT statement
+
+SelectExpression = {
+    Select ~ (From ~ (Where)?)?
```
alancai98, line 28: So `From` is optional here? In partiql-lang-kotlin, a `FROM` clause is currently required, but there was an enhancement [issue](https://github.com/partiql/partiql-lang-kotlin/issues/210) to make it optional. From the [spec](https://partiql.org/assets/PartiQL-Specification.pdf#page=9), `FROM` seems required in the BNF grammar.

partiql-parser/src/partiql.pest
```diff
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+BagConstructor = {
+    LeftDoubleAngle_ ~ Values? ~ RightDoubleAngle_
+}
+
+Values = _{
+    (ValueExpression ~ (Comma_ ~ ValueExpression)*)
+}
+
+TupleConstructor = {
+    LeftCurly_ ~ TupleField ~ (Comma_ ~ TupleField)* ~ RightCurly_
```
alancai98, line 82: Are empty tuples allowed? Could also add an empty tuple test.

partiql-parser/src/partiql.pest
```diff
@@ -2,18 +2,102 @@ WHITESPACE = _{ " " | "\t" | "\x0B" | "\x0C" | "\r" | "\n" }
...
+//
+// Entry point for full query parsing.
+//
+Query = _{ SOI ~ Expression ~ EOI}
+
+//
+// Expressions
+//
+
+Expression = _{
```
alancai98, line 20: Does `ValueExpression` correspond to `expr_query` from the [spec](https://partiql.org/assets/PartiQL-Specification.pdf#page=9)? And similarly for `SelectExpression` to `sfw_query`?

alancai98 (commented):
Had a few questions for the pest grammar. Otherwise LGTM

partiql-parser/src/partiql.pest
dlurton: technically, `from ` should be optional in the Kotlin implementation IMHO this is optional in other SQLs

partiql-parser/src/partiql.pest
almann: Yes, will add.

partiql-parser/src/partiql.pest
almann: Basically yes, I avoided the names because I thought them to be confusing and this is going to be the basis for the grammar in the spec eventually.

partiql-parser/src/partiql.pest
almann: Yes, although non-standard, we should make it optional.

alancai98 (approved):

Issue #43, opened 2021-06-10
`pest2ion` Tool Cannot Generate Pretty/Compact Text for PartiQL Grammar


almann:
This is probably an upstream bug with `ion-rs`/`ion-c-sys`, but when serializing the PartiQL grammar, we get an unexpected EOF:

Binary:
```
❯ cargo run --package pest-ion --bin pest2ion -- partiql-parser/src/peg/partiql.pest -b | wc -c
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/pest2ion partiql-parser/src/peg/partiql.pest -b`
14897
```

Text:
```
❯ RUST_BACKTRACE=1 cargo run --package pest-ion --bin pest2ion -- partiql-parser/src/peg/partiql.pest -t | wc -c
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target/debug/pest2ion partiql-parser/src/peg/partiql.pest -t`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: IonCError { code: 20, message: "IERR_UNEXPECTED_EOF", additional: "iERR Result" }', /home/ANT.AMAZON.COM/almann/.cargo/registry/src/github.com-1ecc6299db9ec823/ion-c-sys-0.4.10/src/writer.rs:612:50
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs:92:14
   2: core::option::expect_none_failed
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:1329:5
   3: core::result::Result<T,E>::unwrap
             at /home/ANT.AMAZON.COM/almann/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1037:23
   4: <ion_c_sys::writer::IonCWriterHandle as core::ops::drop::Drop>::drop
             at /home/ANT.AMAZON.COM/almann/.cargo/registry/src/github.com-1ecc6299db9ec823/ion-c-sys-0.4.10/src/writer.rs:612:13
[...]

Issue #37, opened 2021-06-07
Flatten Pest to Ion `sequence`/`choice` Nodes


almann:
The `sequence`/`choice` nodes in the Pest grammar AST are binary.  We should allow them to flatten such that:

```
(choice
  (choice
    (identifier "a")
    (identifier "b")
  )
  (identifier "c")
)
```

Turn into:

```
(choice
  (identifier "a")
  (identifier "b")
  (identifier "c")
)
```

This makes downstream consumption easier.  Note that `sequence` has higher precedence than `choice`.

PR #44, opened 2021-06-10
Adds `pest2ion` command

### LINKED ISSUE #36: Add CLI Interface to Pest to Ion Converter
As #35 implies, we will have a library to convert Pest grammars to Ion.  We should provide a CLI interface for that.

almann:
Adds a basic CLI driver over the `pest-ion` crate.  This driver supports
setting the input file (defaulting to STDIN), an output file (defaulting
to STDOUT), and flags around the Ion data format.

An explicit TODO is to figure out a good way to test this probably using
integration tests and spawning the command. (See #42)

Resolves #36.

### DIFF
--- a/pest-ion/src/bin/pest2ion/main.rs
+++ b/pest-ion/src/bin/pest2ion/main.rs
@@ -14,7 +14,7 @@ fn main() -> Result<()> {
         .author(crate_authors!())
         .about(crate_description!())
         .arg(
-            Arg::with_name("FILE")
+            Arg::with_name("INPUT_FILE")
                 .help("The input file to parse (defaults to STDIN)")
                 .index(1),
         )
@@ -46,7 +46,7 @@ fn main() -> Result<()> {
         .group(ArgGroup::with_name("format").args(&["text", "binary", "pretty"]))
         .get_matches();
 
-    let elem = if let Some(file_name) = matches.value_of("FILE") {
+    let elem = if let Some(file_name) = matches.value_of("INPUT_FILE") {
         Path::new(file_name).try_pest_to_element()?
     } else {
         // no file argument means read from stdin
--- a/pest-ion/src/bin/pest2ion/main.rs
+++ b/pest-ion/src/bin/pest2ion/main.rs
@@ -1,0 +1,82 @@
+// Copyright Amazon.com, Inc. or its affiliates.
+
+use anyhow::Result;
+use clap::{crate_authors, crate_description, crate_version, App, Arg, ArgGroup};
+use ion_rs::value::writer::{ElementWriter, Format, TextKind};
+use pest_ion::{from_read, TryPestToElement};
+use std::fs::File;
+use std::io::{stdin, stdout, Write};
+use std::path::Path;
+
+fn main() -> Result<()> {
+    let matches = App::new("Pest to Ion Converter")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .about(crate_description!())
+        .arg(
+            Arg::with_name("FILE")
+                .help("The input file to parse (defaults to STDIN)")
+                .index(1),
+        )
+        .arg(
+            Arg::with_name("OUTPUT_FILE")
+                .long("output")
+                .short("o")
+                .takes_value(true)
+                .help("Writes output to the given file (defaults to STDOUT)"),
+        )
+        .arg(
+            Arg::with_name("text")
+                .long("text")
+                .short("t")
+                .help("Generate Ion text (default)"),
+        )
+        .arg(
+            Arg::with_name("binary")
+                .long("binary")
+                .short("b")
+                .help("Generate Ion binary"),
+        )
+        .arg(
+            Arg::with_name("pretty")
+                .long("pretty")
+                .short("p")
+                .help("Generate Ion pretty printed text"),
+        )
+        .group(ArgGroup::with_name("format").args(&["text", "binary", "pretty"]))
+        .get_matches();
+
+    let elem = if let Some(file_name) = matches.value_of("FILE") {
+        Path::new(file_name).try_pest_to_element()?
+    } else {
+        // no file argument means read from stdin
+        from_read(stdin()).try_pest_to_element()?
+    };
+
+    // currently Ion element requires a fixed buffer to serialize to, let's choose something
+    // relatively big until this limitation is lifted
+    const BUFFER_SIZE: usize = 32 * 1024 * 1024;
+    let mut out_buf = vec![0u8; BUFFER_SIZE];
+
+    let format = if matches.is_present("binary") {
+        Format::Binary
+    } else if matches.is_present("pretty") {
+        Format::Text(TextKind::Pretty)
+    } else {
+        Format::Text(TextKind::Compact)
+    };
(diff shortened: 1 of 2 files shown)


pest-ion/src/bin/pest2ion/main.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::io::{stdin, stdout, Write};
+use std::path::Path;
+
+fn main() -> Result<()> {
+    let matches = App::new("Pest to Ion Converter")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .about(crate_description!())
+        .arg(
+            Arg::with_name("FILE")
```
zslayton, line 17: `INPUT_FILE`/`OUTPUT_FILE` would have a nicer symmetry IMO.

pest-ion/src/bin/pest2ion/main.rs
```diff
@@ -0,0 +1,82 @@
...
+                .help("The input file to parse (defaults to STDIN)")
+                .index(1),
+        )
+        .arg(
+            Arg::with_name("OUTPUT_FILE")
```
zslayton, line 22: I think `FILE` above is all caps because it's an anonymous argument. If that's true, then this should be `output-file`?

pest-ion/src/bin/pest2ion/main.rs
```diff
@@ -0,0 +1,82 @@
...
+                .short("b")
+                .help("Generate Ion binary"),
+        )
+        .arg(
+            Arg::with_name("pretty")
+                .long("pretty")
+                .short("p")
+                .help("Generate Ion pretty printed text"),
+        )
+        .group(ArgGroup::with_name("format").args(&["text", "binary", "pretty"]))
```
zslayton, line 46: I'm unsure how this works. Do I need to do `--format=p`? For reference, this is [how the `ion-cli` does it](
https://github.com/amzn/ion-cli/blob/4ff67c313b9f2c00c73a4f4cc9a21d0ef009d263/src/bin/ion/commands/dump.rs#L48-L54):

```rust
            Arg::with_name("format")
                .long("format")
                .short("f")
                .takes_value(true)
                .default_value("pretty")
                .possible_values(&["binary", "text", "pretty"])
                .help("Output format"),
```

pest-ion/src/bin/pest2ion/main.rs
```diff
@@ -0,0 +1,82 @@
...
+                .long("text")
+                .short("t")
+                .help("Generate Ion text (default)"),
+        )
+        .arg(
+            Arg::with_name("binary")
+                .long("binary")
+                .short("b")
+                .help("Generate Ion binary"),
+        )
```
dlurton, line 39: Not sure what the use-case for outputting binary ion from this tool is...  Do you expect the size benefits of the binary format be of significant benefit?

pest-ion/src/bin/pest2ion/main.rs
```diff
@@ -0,0 +1,82 @@
...
+        .arg(
+            Arg::with_name("pretty")
+                .long("pretty")
+                .short("p")
+                .help("Generate Ion pretty printed text"),
+        )
```
dlurton, line 45: I find it a stretch to believe that the space savings of non-pretty text or binary will be enough to justify the additional complexity.  Unless there's something you're seeing here that I'm not?

dlurton (approved):
I won't block on this, but... I guess I don't understand the need to output anything other than pretty Ion text format.  Unless you're expecting tens of megabytes of Ion to come out this tool?

pest-ion/src/bin/pest2ion/main.rs
almann: The use case is actually giving me a way to work around #43 (which apparently there is a bug with serializing deeply nested Ion).  #37 probably helps, but I see no reason not to give a little flexibility here.

pest-ion/src/bin/pest2ion/main.rs
almann: The reference is [here](https://docs.rs/clap/2.33.3/clap/struct.ArgGroup.html).  Basically I have a mutual exclusion to `-p`/`-b`/`-t` flags.  If you think we should use an option flag to be more consistent with `ion-cli` (which is marginally verbose), I am not opposed.

pest-ion/src/bin/pest2ion/main.rs
almann: It actually appears in the help, and this is somewhat consistent to help notation I've seen elsewhere:

```
❯ cargo run --package pest-ion --bin pest2ion -- --help
   Compiling pest-ion v0.0.0 (/home/ANT.AMAZON.COM/almann/CLionProjects/partiql-lang-rust/pest-ion)
    Finished dev [unoptimized + debuginfo] target(s) in 0.88s
     Running `target/debug/pest2ion --help`
Pest to Ion Converter 0.0.0
PartiQL Team <partiql-team@amazon.com>
A simple Pest grammar to Ion converter

USAGE:
    pest2ion [FLAGS] [OPTIONS] [FILE]

FLAGS:
    -b, --binary     Generate Ion binary
    -h, --help       Prints help information
    -p, --pretty     Generate Ion pretty printed text
    -t, --text       Generate Ion text (default)
    -V, --version    Prints version information

OPTIONS:
    -o, --output <OUTPUT_FILE>    Writes output to the given file (defaults to STDOUT)

ARGS:
    <FILE>    The input file to parse (defaults to STDIN)
```

pest-ion/src/bin/pest2ion/main.rs
almann: That is reasonable, I will change this.

pest-ion/src/bin/pest2ion/main.rs
zslayton: `ion-cli` takes the slightly more verbose route (`-f binary`) largely to preserve flag name options for other use cases. Virtually every command needs to support different Ion formats. Some may have really good specialized uses for `-b`, `-p`, and `-t`. `pest2ion` is much more limited in scope, so consuming the shortcut namespace isn't as much of an issue.

almann: Yeah, to be honest, I only exposed these flags because of #43, there is an underlying bug in `ion-c-sys`, so instead of hard coding to binary, I decided to add these.

My question to you, should we make this consistent?  Not opposed, though not really planning on adding this as an `ion-cli` sub-command.

pest-ion/src/bin/pest2ion/main.rs
zslayton: You're right, now that I see it in the context of the output it does look right.

pest-ion/src/bin/pest2ion/main.rs
zslayton: > My question to you, should we make this consistent? Not opposed, though not really planning on adding this as an ion-cli sub-command.

I think you can leave it as-is. If this becomes widely used enough to justify adding it to `ion-cli` we can always tweak the interface in that version for consistency.

zslayton (approved):
🚀

PR #47, opened 2021-06-12
Adds `pest2ion` CLI test for PartiQL Grammar

almann:
Adds a `TestCase` struct to encapsulate the test case parameters for
testing the CLI.  Makes a simple conversion from `&'static str` pair
which is for when you have simple in/out type of tests.  Adds function
to generate a test case out of the PartiQL grammar file which doesn't
really validate anything about the grammar serialization other than we
get something the library would produce.

### DIFF
--- a/pest-ion/tests/cli.rs
+++ b/pest-ion/tests/cli.rs
@@ -2,6 +2,7 @@
 
 use anyhow::Result;
 use assert_cmd::Command;
+use ion_rs::value::owned::OwnedElement;
 use ion_rs::value::reader::*;
 use rstest::*;
 use std::fs::File;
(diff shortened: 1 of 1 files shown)


pest-ion/tests/cli.rs
```diff
@@ -16,8 +17,49 @@ enum FileMode {
...
+    /// The text of the Pest grammar to test
+    pest_text: S,
+    /// The expected Ion
+    expected_ion: OwnedElement,
+}
+
+impl<S: AsRef<str>> TestCase<S> {}
+
+impl From<(&'static str, &'static str)> for TestCase<&'static str> {
+    /// Simple conversion for static `str` slices into a test acse
```
zslayton, line 31:
```suggestion
    /// Simple conversion for static `str` slices into a test case
```

pest-ion/tests/cli.rs
```diff
@@ -16,8 +17,49 @@ enum FileMode {
...
 
+/// Simple wrapper for a CLI test case.
+struct TestCase<S: AsRef<str>> {
+    /// The text of the Pest grammar to test
+    pest_text: S,
+    /// The expected Ion
+    expected_ion: OwnedElement,
+}
+
+impl<S: AsRef<str>> TestCase<S> {}
```
zslayton, line 28: What's this empty `impl` block doing?

pest-ion/tests/cli.rs
```diff
@@ -29,14 +71,19 @@ enum FileMode {
             }
         }
     "#
-)]
-fn run_it(
-    #[case] pest_src: &str,
-    #[case] ion_text: &str,
+).into())]
+#[case::partiql(partiql_test_case())]
+fn run_it<S: AsRef<str>>(
```
zslayton, line 76: Is this `S: AsRef<str>` doing anything? I see `as_ref()` is called on `pest_text` below, but that's already a string.

pest-ion/tests/cli.rs
almann: Nothing--this was a remnant from when I was going to make a `new`.

pest-ion/tests/cli.rs
almann: Yes, in some cases we throw `&'static str` in the case where we load PartiQL's grammar, we pass in a `String`, so this just normalizes the test case input.

zslayton (approved):

abhikuhikar (approved):
Ship it.

PR #56, opened 2022-02-19
Add Experimental AST

am357:
*Issue #, if available:* 52

*Description of changes:*
This commit includes an experimental implementation of PartiQL AST
 based on the current kotlin implementation which has been
auto-generated by `partiql-ir-generator`:
- https://github.com/partiql/partiql-ir-generator

We want to create an experimental PartiQL AST so that:

- solidify our expected core model (at least for the first iteration)
  prior to developing AST code generation i.e. using
  partiql-ir-generator spec or a substitute.
- we can get an early review with having integration with parser
  in the picture.

This implementation does _not_ include:
- The addition of MetaContainer and with_meta method.
  For code region data, We try to see if a substitute approach like
  using Span works as a substitute approach.

- The implementation of std::fmt::Display Trait for ast entities.
  Will be tracked in a separate issue.
- The tree traversal and transform utilities.
  Will be tracked in a separate issue.

### DIFF
--- a/partiql-ast/src/experimental/ast.rs
+++ b/partiql-ast/src/experimental/ast.rs
@@ -5,14 +5,14 @@
 //! Two main entities in the module are [`Item`] and [`ItemKind`]. `Item` represents an AST element
 //! and `ItemKind` represents a concrete type with the data specific to the type of the item.
 
-// Types defined in this module are mostly from the current partiql-ir-generator spec. and comments
+// Structures in this module are mostly from the current `partiql-ir-generator` spec. and comments
 // are excerpts from the spec definition.
 // https://github.com/partiql/partiql-lang-kotlin/blob/4bcfc7f73d3e6e54286bcc03a54d5f6425eec4cc/lang/resources/org/partiql/type-domains/partiql.ion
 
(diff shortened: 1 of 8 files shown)


am357:
Looks like the codecov has skipped, will work on it while the review is in progress.

partiql-ast/src/experimental/ast.rs
```diff
@@ -112,25 +212,140 @@ pub enum ExprKind {
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct BitStringLit {
+    pub value: String,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub struct HexStringLit {
+    pub value: String,
+}
```
jpschorr, line 318: Since these `Lit` fields have only a single value, they are effectively 'newtype's. 
Should we consider using [tuple structs](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#using-tuple-structs-without-named-fields-to-create-different-types) for them? Or even inlining them directly in their containing enum?

am357: Was thinking perhaps we need to enrich them later down the track with additional data (code span, etc.). What do you think?

Issue #58, opened 2022-03-14
Add Prototype Parser


jpschorr:
Prototype a parser (or parsers) from text into the experimental AST in https://github.com/partiql/partiql-lang-rust/issues/52 & https://github.com/partiql/partiql-lang-rust/issues/56

This should allow us to:
- evaluate the experimental AST
-  evaluate 1 or more parsing strategies for ergonomics, completeness, performance, etc.

### DoD

- a PR for reviewing the prototype parser(s)
- next steps

PR #62, opened 2022-03-21
Rework embedded ion & comment lexers.

jpschorr:
#58

### Rework embedded ion & comment lexers.

* Add embeddable block comment lexer
* Add embeddable ion lexer, including block comment lexer
* Refactor PartiQL lexer to embed the ion & block comment lexers
* Add ion lexing & bench tests
* Add lexer error tests

### DIFF
--- a/partiql-parser/src/lalr/lexer.rs
+++ b/partiql-parser/src/lalr/lexer.rs
@@ -19,9 +19,192 @@ pub enum LexicalError {
     /// Embedded Ion value is not properly terminated.
     #[error("Lexing error: unterminated ion literal")]
     UnterminatedIonLiteral(Spanned<(), usize>),
+    /// Comment is not properly terminated.
+    #[error("Lexing error: unterminated comment")]
+    UnterminatedComment(Spanned<(), usize>),
     /// Any other lexing error.
     #[error("Lexing error: unknown error")]
     Unknown,
+}
+
+type CommentToken = SpannedResult<String, usize, LexicalError>;
+
+#[derive(Logos, Debug, Clone, PartialEq, Eq)]
+enum Comment {
+    #[error]
+    #[regex(r"[^/*]+", logos::skip)]
+    Any,
+    #[token("*/")]
+    End,
+    #[token("/*")]
+    Start,
+}
+
+/// A lexer for block comments (enclosed between '/*' & '*/')
+pub(crate) struct CommentLexer<'a> {
+    /// Wrap a logos-generated lexer
+    lexer: logos::Lexer<'a, Comment>,
+    nested_block: bool,
+}
+
+impl<'a> CommentLexer<'a> {
+    /// Creates a new lexer over `input` text.
+    pub fn new(input: &'a str) -> Self {
+        CommentLexer {
+            lexer: Comment::lexer(input),
+            nested_block: false,
+        }
+    }
+
+    fn create(lexer: logos::Lexer<'a, Comment>) -> Self {
+        CommentLexer {
+            lexer,
+            nested_block: false,
+        }
+    }
+
+    fn exit(self) -> logos::Lexer<'a, Comment> {
+        self.lexer
+    }
+
+    fn with_nested_block(mut self) -> Self {
+        self.nested_block = true;
+        self
+    }
+
+    /// Parses a single (possibly nested) block comment and returns it
+    fn next(&mut self) -> Option<CommentToken> {
+        let Span { start, .. } = self.lexer.span();
+        let mut nesting = 1;
+        let nesting_inc = if self.nested_block { 1 } else { 0 };
+        'comment: loop {
+            match self.lexer.next() {
+                Some(Comment::Any) => continue,
+                Some(Comment::Start) => nesting += nesting_inc,
+                Some(Comment::End) => {
+                    nesting -= 1;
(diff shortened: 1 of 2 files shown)


partiql-parser/src/lalr/lexer.rs
```diff
@@ -475,32 +589,86 @@ mod tests {
...
+    fn err_unterminated_ion_comment() {
+        let query = r#" `/*12345678`"#;
+        let toks: Result<Vec<_>, LexicalError> = EmbeddedIonLexer::new(query).collect();
+        assert!(toks.is_err());
+        let error = toks.unwrap_err();
+        assert!(matches!(
+            error,
+            LexicalError::UnterminatedComment((2, (), 11))
+        ));
+        assert_eq!(error.to_string(), "Lexing error: unterminated comment");
```
alancai98, line 672: Nice set of tests! I could only think of a couple more tests that could be added:
- query starting with block comment closing, `*/`
  - Something like: `*/ SELECT g`
- within an Ion literal, start with a block comment closing, `*/`
  - Something like: `` `*/` SELECT g``

alancai98 (commented):
New Ion and comment lexers looks good to me. Only had a question about how to more easily verify the span's offsets and a couple more test suggestions.

partiql-parser/src/lalr/lexer.rs
```diff
@@ -447,11 +538,34 @@ mod tests {
...
                       'b':1}` "#;
         let mut lexer = Lexer::new(ion_value);
-        if let (_start, Token::Ion(s), _end) = lexer.next().unwrap().unwrap() {
-            assert_eq!(ion_value.trim().trim_matches('`'), s);
-        } else {
-            panic!("Lex failed")
-        }
+
+        let tok = lexer.next().unwrap().unwrap();
+        assert!(matches!(tok, (1, Token::Ion(ion_str), 154) if ion_str == ion_value.trim()));
```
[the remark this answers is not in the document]
jpschorr: > Is there an easier way to verify the end offset (i.e. `154`) other than manually counting?

Not really.

> Is there a way to convert an offset to a line and column number?

It generally takes a lot of additional work to do that in the lexer and pipe it all the way through everything. It's generally easier to do that retroactively, once you know you need to present such information. (e.g. we could lex with an alternative lexer that only tabulates the offsets of all line beginnings. then offset -> line&col is a rather simple calculation).

partiql-parser/src/lalr/lexer.rs
```diff
@@ -447,11 +538,34 @@ mod tests {
...
+    fn nested_comments() {
+        let comments = r##"/*  
+                                    /*  / * * * /
+                                    /*  ' " ''' ` 
+                                    */  text
+                                    */  1 2 3 4 5 6,7,8,9 10.112^5
+                                    */"##;
+
+        let nested_lex = CommentLexer::new(comments).with_nesting();
+        assert_eq!(nested_lex.into_iter().count(), 1);
```
am357, line 561: qq: just trying to understand how iterator works in this case.

Here we have 3 blocks nested inside each other:
```bash
/*
 /*
   / * * * /
 

   /*
     ' " ''' ` 
   */

   text
  */

  1 2 3 4 5 6,7,8,9 10.112^5
*/
```

Why does the _count_ returns 1?

jpschorr: Lexers really aren't designed to deal with nested or recursive structures; that's the domain of parsers. Here, the idea is to parse this as 1 single comment (the outer comment). It happens to contain other comments, in the same way that a string may contain escaped quotes.

At the end of the day, we really just want to skip past comments, so recognizing the nesting and encompassing it is sufficient.

am357: > Lexers really aren't designed to deal with nested or recursive structures; that's the domain of parsers. Here, the idea is to parse this as 1 single comment (the outer comment). It happens to contain other comments, in the same way that a string may contain escaped quotes.
> 
> At the end of the day, we really just want to skip past comments, so recognizing the nesting and encompassing it is sufficient.

Makes sense, thanks for elaborating. Totally resonate with that (we can't use FAs for nested structures).

partiql-parser/src/lalr/lexer.rs
jpschorr: At least as currently designed, these are both valid lexes:

- `*/ SELECT g`
  - This is a `*` operator, then a `/` operator, then a `SELECT` keyword, then a `g` identifier
  - This might be worth adding a token for, so we can provide better ergonomics of error messaging, though...
- `` `*/` SELECT g ``
  -  This is an ion string, then a `SELECT` keyword, then a `g` identifier
     - (the ion string is technically invalid, but this lexer is just trying to find the text between the backticks that we'll have a proper ion library parse)

am357 (approved):

partiql-parser/src/lalr/lexer.rs
am357: This is interesting, at the end of the day machines just do their job :)

Maybe a TODO for the first item?

alancai98: Ok, got it. 1st case be a parser error then?

alancai98 (approved):

Issue #60, opened 2022-03-17
Add `Item` formatting to experimental AST


am357:
As PartiQL _contributor_ I want to have `Item` formatting for [the experimental AST](https://github.com/partiql/partiql-lang-rust/issues/52) so that I can view the resulting AST from parsing if required.

#### DoD
- We have formatting for `Item` as a merged PR.

PR #68, opened 2022-03-22
Add Display formatting to Item

am357:
*Issue #, if available:* #60

*Description of changes:*

This change adds a Display formatting to the `Item` and re-uses Debug
formatting for now.

Example:
```bash
Query(Query { expr: Expr { kind: Lit(Lit { kind: NumericLit(NumericLit { kind: Int32Lit(Int32Lit { value: 12 }) }) }) } })
```

### DIFF
--- a/partiql-ast/src/experimental/ast.rs
+++ b/partiql-ast/src/experimental/ast.rs
@@ -10,8 +10,8 @@
 // https://github.com/partiql/partiql-lang-kotlin/blob/4bcfc7f73d3e6e54286bcc03a54d5f6425eec4cc/lang/resources/org/partiql/type-domains/partiql.ion
 
 // TODO Add documentation.
-// TODO Add Display formatting.
 
+use std::fmt;
 use rust_decimal::Decimal as RustDecimal;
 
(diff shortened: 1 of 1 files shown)


partiql-ast/src/experimental/ast.rs
```diff
@@ -23,6 +23,13 @@ pub struct Item {
     // and how it contributes to backward compatibility.
 }
 
+impl fmt::Display for Item {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
```
alancai98, line 27: nit: signature from the [docs](https://doc.rust-lang.org/std/fmt/trait.Display.html) is slightly different. I don't think it should matter, though.

```
fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result
```

alancai98 (approved):

jpschorr (approved):

partiql-ast/src/experimental/ast.rs
am357: Yep, in this case we don't use the _lifetime_ but I'll change it in a new commit.

PR #70, opened 2022-03-22
Instrument the lexers to count lines in the lexed source.

jpschorr:
#58 

Instrument the lexers to count lines in the lexed source.

### DIFF
--- a/partiql-parser/src/lalr/lexer.rs
+++ b/partiql-parser/src/lalr/lexer.rs
@@ -1,7 +1,7 @@
 use crate::result::LineAndColumn;
 use itertools::Itertools;
 use logos::{Logos, Span};
-use smallvec::SmallVec;
+use smallvec::{smallvec, SmallVec};
 use std::cmp::max;
 use std::ops::Range;
 use thiserror::Error;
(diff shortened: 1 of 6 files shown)


partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,64 @@
...
+/// assert_eq!(tracker.at(0), LineAndColumn::at(1,1));
+/// assert_eq!(tracker.at(6), LineAndColumn::at(2,1));
+/// assert_eq!(tracker.at(400), LineAndColumn::at(4,375));
+/// ```
+#[derive(Default)]
+pub struct LineOffsetTracker {
+    newlines: SmallVec<[usize; 8]>,
+}
+
+impl LineOffsetTracker {
```
jpschorr, line 35: Revisiting this, I realize this won't handle unicode. We'll need to actually count the chars in a line, not just convert the byte offsets into column offsets.

partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,112 @@
...
+    /// `offset` is the byte offset
+    #[inline(always)]
+    fn line_num_from_byte_offset(&self, offset: usize) -> usize {
+        match self.line_starts.binary_search(&offset) {
+            Err(i) => i,
+            Ok(i) => i + 1,
+        }
+    }
+
+    /// Convert a source offset into a
```
alancai98, line 86: Line can be deleted

partiql-parser/src/lalr/lexer.rs
```diff
@@ -31,9 +136,16 @@ pub enum LexicalError {
 type CommentToken = SpannedResult<String, usize, LexicalError>;
 
 #[derive(Logos, Debug, Clone, PartialEq, Eq)]
+#[logos(extras = &'s mut LineOffsetTracker)]
```
alancai98, line 139: qq: what does this `logos` annotation w/ `extras` do?

partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,112 @@
...
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if `num` is not within the number of lines
+    /// of source seen (i.e. 1 <= `num` <= self.num_lines()`).
+    #[inline(always)]
+    fn line_span_from_line_num(&self, num: usize) -> Range<usize> {
+        assert!(1 <= num);
+        assert!(num <= self.num_lines());
+        if num >= self.num_lines() {
```
am357, line 68: Since we're  asserting `num <= N` above, shouldn't we remove `>` here?

partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,112 @@
...
+        LineOffsetTracker {
+            line_starts: smallvec![0], // line 1 starts at offset `0`
+        }
+    }
+}
+
+impl LineOffsetTracker {
+    /// Record a newline at `span` in the source
+    #[inline(always)]
+    pub fn record(&mut self, span: Range<usize>) {
```
am357, line 46: Does it make sense to only accept `span.end`?

partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,112 @@
...
+    #[inline(always)]
+    fn line_num_from_byte_offset(&self, offset: usize) -> usize {
+        match self.line_starts.binary_search(&offset) {
+            Err(i) => i,
+            Ok(i) => i + 1,
+        }
+    }
+
+    /// Convert a source offset into a
+    /// Calculates a [`LineAndColumn`] for a byte offset into a given `&str`
```
am357, line 87: nit: _from the given `&str`_?

partiql-parser/src/lalr/lexer.rs
```diff
@@ -570,29 +707,73 @@ mod tests {
...
+                Token::Identifier("a".to_owned())
+            ],
+            toks.into_iter().map(|(_s, t, _e)| t).collect::<Vec<_>>()
+        );
+
+        assert_eq!(offset_tracker.num_lines(), 3);
+        assert_eq!(offset_tracker.at(query, 0), LineAndColumn::at(1, 1));
+        assert_eq!(offset_tracker.at(query, 1), LineAndColumn::at(1, 2));
+        assert_eq!(offset_tracker.at(query, 9), LineAndColumn::at(2, 1));
+        assert_eq!(offset_tracker.at(query, 19), LineAndColumn::at(3, 1));
```
alancai98, line 732: nit: could add a line for checking the end of the query. Something like

```
        assert_eq!(offset_tracker.at(query, 28), LineAndColumn::at(3, 10));
```

partiql-parser/src/lalr/lexer.rs
```diff
@@ -570,29 +707,73 @@ mod tests {
...
+
+        Ok(())
+    }
+
+    #[test]
+    fn select_unicode() -> Result<(), LexicalError> {
```
am357, line 738: +1

partiql-parser/src/lalr/lexer.rs
alancai98: Also, is there any way to easily check for panics in assertions (other than creating a test w/ the `should_panic` annotation)? I'm wondering if we should test the panic behavior of `at`:

https://github.com/partiql/partiql-lang-rust/blob/8b9ae1993671b94dcdb2b7a17d747a7c522b3bfc/partiql-parser/src/lalr/lexer.rs#L92-L96

partiql-parser/src/lalr/lexer.rs
jpschorr: > Associated type Extras for the particular lexer. This can be set using #[logos(extras = MyExtras)] and accessed inside callbacks.

From https://docs.rs/logos/0.12.0/logos/trait.Logos.html#associatedtype.Extras

partiql-parser/src/lalr/lexer.rs
jpschorr: From the point of view of `LineOffsetTracker`, yes. 

But, I left it accepting a span to make the Logos lexer definition a little easier to read.
Ideally, `record` will be inlined so it shouldn't matter from a performance perspective.

partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,7 +1,107 @@
...
+
+    /// Calculate the number of lines of source seen so far.
+    #[inline(always)]
+    pub fn num_lines(&self) -> usize {
+        self.line_starts.len()
+    }
+
+    /// Calculates the byte offset span ([`Range`]) of a line.
+    ///
+    /// `num` is the line number (1-indexed) for which to  calculate the span
```
am357, line 58: nit: could also add `max` def.

am357 (approved):
🚢

alancai98 (approved):

PR #75, opened 2022-03-24
Initial conformance test suite runner implementation

alancai98:
Issue: #78 

Initial implementation of conformance test suite runner in Rust using parsing test data from `partiql-tests`.
- Creates a new crate called `partiql-conformance-tests`
- Adds `partiql-tests` as a git submodule to `partiql-conformance-tests` crate
- `build.rs` file loads the parser test data and generates test functions for each file
  - Uses the [codegen](https://docs.rs/codegen/0.1.3/codegen/) crate to generate the tests
  - Tests are integration tests (i.e. located in `tests/` directory). TBD if this is the approach we want to take.

FYI, this is the first Rust code I've written outside some toy examples, so feel free to critique.


am357 (dismissed):
Apologies as I approved by mistake (apparently been on a wrong window and didn't realize).

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+
+fn main() {
+    let pass_parse_test_files: Vec<PathBuf> = all_ion_files_in(PASS_PARSE_TESTS_DIR);
+    let fail_parse_test_files: Vec<PathBuf> = all_ion_files_in(FAIL_PARSE_TESTS_DIR);
+
+    for pass_parse_test_file in &pass_parse_test_files {
+        let mut scope = create_pass_parse_scope(pass_parse_test_file);
+        scope.import(PARSER_IMPORT.0, PARSER_IMPORT.1);
+        let full_file_name = to_full_file_name("pass_parse", pass_parse_test_file);
+        let dest_path: PathBuf = ["tests", full_file_name.as_str()].iter().collect();
```
alancai98, line 24: Destination path is `tests/pass_parse_<file_name>.rs` (and is `tests/fail_parse_<file_name>.rs`. This is unconventional for build.rs scripts. From the [docs](https://doc.rust-lang.org/cargo/reference/build-script-examples.html):

>In general, build scripts should not modify any files outside of OUT_DIR. It may seem fine on the first blush, but it does cause problems when you use such crate as a dependency, because there's an implicit invariant that sources in .cargo/registry should be immutable. cargo won't allow such scripts when packaging.

Previously, I had generated the tests to a file in the `OUT_DIR`, which generates to the `target` folder. But there was limited IDE support to run the tests.

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+            ))
+            .line("assert!(parse_result.is_err());");
+        generated_fun
+    }
+}
+
+/// Parses the given .ion file `path` and creates a new [`codegen::Scope`] with a
+/// [`codegen::Function`] for each test in `path`.
+///
+/// Assumes the given tests in `path` all parse without errors.
```
alancai98, line 143: Currently, the data file parsing assumes the files have the correct schema and are at the correct directory (e.g. successful parse tests are in `pass/parse` directory). Can we make these assumptions, assuming there's data file validation done by `partiql-tests`? Otherwise, I can add more checks in the test runner, verifying the schema.

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+    dest_str
+}
+
+trait TestFunction {
+    fn to_test_function(&self) -> Function;
+}
+
+/// Represents a conformance test with a PartiQL statement that will parse without errors
+/// TODO: add additional parse checks (e.g. ast)
+struct ParsePassTestCase {
```
alancai98, line 86: As another TODO, need to model these with all the fields present in the test data schema (e.g. `assert` is missing).

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+        };
+    }
+    scope
+}
+
+/// Parses the given .ion file `path` and creates a new [`codegen::Scope`] with a
+/// [`codegen::Function`] for each test in `path`.
+///
+/// Assumes the given tests in `path` give errors when parsed.
+fn create_fail_parse_scope(path: &Path) -> Scope {
```
alancai98, line 223: There's a lot of repetition here with `create_pass_parse_scope`. I can reduce a lot by changing the modeling of the `ParsePassTestCase`/`ParseFailTestCase` and modifying the function signatures.

partiql-conformance-tests/build.rs
jpschorr: I think there's likely utility in not making those assumptions blindly. 

For example, it might be nice to prototype a test case just by adding some test cases to an ion file and pointing this generator at it.

partiql-conformance-tests/build.rs
jpschorr: I think this is ok for now. Essentially we're just harnessing the build script to generate the test code. 

The alternatives are either not be able to run the tests easily to debug or to use our own pre-processing step, which doesn't seem worth it.

partiql-conformance-tests/tests/fail_parse_select-from-where.rs
```diff
@@ -0,0 +1,7 @@
+use partiql_parser::lalr_parse;
+
+#[test]
+fn SFW_without_WHERE_expression() {
```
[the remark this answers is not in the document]
jpschorr: Even if the generated tests don't get checked in, they will cause annoying warnings during compile. I recommend adding an `#[allow(non_snake_case)]` annotation.

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+
+impl TestFunction for ParseFailTestCase {
+    /// Creates a [`codegen::Function`] that checks the parsing of `statement` results in a parsing
+    /// error.
+    fn to_test_function(&self) -> Function {
+        let complete_test_name = match &self.namespace {
+            Some(ns) => format!("{}_{}", ns, &self.test_name),
+            None => self.test_name.to_owned(),
+        };
+        let mut generated_fun = Function::new(complete_test_name.replace(' ', "_").as_str());
```
jpschorr, line 128: `replace(' ', "_")` appears a couple times in this file.

You'll probably want to extract out some `escape_*` functions to do this escaping, as it's not just spaces in the test names that you'll have to deal with eventually (ion allows some punctuation in symbol names and also, you may have to mangle names that happen to also be rust reserved words)

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+            .with_extension("rs")
+            .file_name()
+            .unwrap()
+            .to_str()
+            .unwrap()
+    )
+}
+
+/// Converts the scope to a String that is passed through codegen's default formatter
+fn scope_to_formatted_string(scope: Scope) -> String {
```
jpschorr, line 72: I think this is [Scope.to_string](https://docs.rs/codegen/0.1.3/codegen/struct.Scope.html#method.to_string)

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+use std::fs;
+use std::fs::File;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use walkdir::WalkDir;
+
+const PASS_PARSE_TESTS_DIR: &str = "partiql-tests/partiql-test-data/pass/parser";
+const FAIL_PARSE_TESTS_DIR: &str = "partiql-tests/partiql-test-data/fail/parser";
+const PARSER_IMPORT: (&str, &str) = ("partiql_parser", "lalr_parse");
+
```
jpschorr, line 15: There are some things that you can `println` in main here to [direct cargo](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) how and when to rerun your build script.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "partiql-conformance-tests/partiql-tests"]
```
jpschorr, line 1: We'll need to document somewhere whatever steps needed to be taken to pull down the test submodules (`git submodule update`) in additon to just the repo clone.

partiql-conformance-tests/tests/fail_parse_select-from-where.rs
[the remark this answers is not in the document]
jpschorr: I think, perhaps, we shouldn't check them in.
Checking them in is duplicating the test data, potentially into a lot of code, that not all user might ever care to generate and/or run.

For that matter, perhaps this crate's `build.rs` should succeed even if the submodule hasn't been pulled down? (for the case that users just want to use the PartiQL code, not test it's conformance?)

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+                    .to_owned();
+
+                let test_case = ParsePassTestCase {
+                    test_name,
+                    statement,
+                    namespace: None,
+                };
+                scope.push_fn(test_case.to_test_function());
+            }
+            IonType::List => {
```
jpschorr, line 174: This seems to only allow one level of namespacing?

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+                                ion_data
+                                    .annotations()
+                                    .next()
+                                    .unwrap()
+                                    .text()
+                                    .unwrap()
+                                    .to_owned(),
+                            );
+
+                            let test_struct = tc.as_struct().unwrap();
```
jpschorr, line 189: This duplicates the above parsing. I think it would be better to model this a simply recursion of ion lists (representing namespace levels that get turned into scopes) with structs at the bottom (representing the test function that gets added to the closest enclosing scope)

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+        let dest_path: PathBuf = ["tests", full_file_name.as_str()].iter().collect();
+        File::create(dest_path)
+            .expect("File creation failed")
+            .write_all(scope_to_formatted_string(scope).as_bytes())
+            .unwrap_or_else(|error| panic!("Failure when writing to file: {:?}", error));
+    }
+}
+
+/// Returns a vector of all .ion files in the directory `dir`
+fn all_ion_files_in(dir: &str) -> Vec<PathBuf> {
```
jpschorr, line 44: It might be a bit more idiomatic to write:

```rust
fn all_ion_files_in(dir: &str) -> walkdir::Result<Vec<PathBuf>> {
    WalkDir::new(dir)
        .into_iter()
        .filter_map(|result| match result {
            Ok(entry) => {
                if let Some(OsStr::new("ion")) = entry.path().extension() {
                    Some(Ok(entry))
                } else {
                    None
                }
            }
            _ => Some(result),
        })
        .collect()
}
```

And unwrap the error in the calling function. 

I'd rather not panic like this in a utility function.

But idiomatically, you can replace `.unwrap_or_else(|error| panic!("Failure during dir traversal: {:?}", error))` with `.expect("Failure during dir traversal")`

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
...
+            entry.unwrap_or_else(|error| panic!("Failure during dir traversal: {:?}", error))
+        })
+        .filter(|entry| entry.path().is_file())
+        .map(|entry| entry.path().to_owned())
+        .filter(|f| f.extension() == Some(ion_file_extension));
+    ion_file_iterator.collect()
+}
+
+/// Returns a file name with prefix prepended and 'rs' as the extension
+fn to_full_file_name(prefix: &str, ion_file: &Path) -> String {
```
jpschorr, line 58: Readability might be improved with 

```rust
fn to_full_file_name(prefix: &str, ion_file: &Path) -> String {
    let ion_file = ion_file
        .with_extension("rs")
        .file_name()
        .unwrap()
        .to_str()
        .unwrap();
    format!("{}_{}", prefix, ion_file)
}
```

partiql-conformance-tests/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+version = "0.0.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[build-dependencies]
+walkdir = "2.3"
+ion-rs = "0.6.0"
+codegen = "0.1.3"
+partiql-parser = { path = "../partiql-parser" }
```
jpschorr, line 24: I don't think this is needed as a build dep, is it?

partiql-conformance-tests/build.rs
am357: I see these two (validation in partiql-tests and in each language runner) being somewhat independent as we might have different levels of validations based on requirements in target languages. Agree with Josh.

partiql-conformance-tests/tests/fail_parse_select-from-where.rs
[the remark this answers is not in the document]
am357: > I think, perhaps, we shouldn't check them in. Checking them in is duplicating the test data, potentially into a lot of code, that not all user might ever care to generate and/or run.
> 
> For that matter, perhaps this crate's `build.rs` should succeed even if the submodule hasn't been pulled down? (for the case that users just want to use the PartiQL code, not test it's conformance?)

The only issue I see with not having them checked-in is that we always have to deal with this implicit thing; on the flip side, having them checked-in, will give visibility on the effective tests currently on the fly. Agree with suppressing warnings and letting the build to proceed.

partiql-conformance-tests/build.rs
```diff
@@ -0,0 +1,296 @@
+use codegen::{Formatter, Function, Scope};
```
am357, line 1: A documentation block here seems useful.

partiql-conformance-tests/build.rs
am357: Just a food for thought: do we see using this package as a dependency in future?

partiql-conformance-tests/build.rs
am357: Agreeing with Josh; on the other side, if it's too much of an effort, maybe we can just restrict the test names for now.

partiql-conformance-tests/build.rs
jpschorr: I don't think so. Especially if all of the parsing, model, & generation code are pulled out to a different crate.

alancai98 (commented):
Thanks for the review and detailed feedback! Should have addressed most of the comments in this review.

New build should fail since one of the parser tests `panic`s. Still have some TODOs that need addressed. I wanted to get a feel for if this is headed in the right direction before adding more docs+tests.

partiql-conformance-tests/src/main.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
jpschorr, line 1: Is this needed because Cargo expects either a `bin` or `lib` and doesn't really expect an integration-test-only crate?

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,102 @@
+use crate::schema::TestCaseKind::Parse;
+use codegen::{Function, Module, Scope};
+
+/// Conformance test document containing namespaces and/or tests
+pub struct TestDocument {
```
jpschorr, line 5: Does there need to be a difference between a `TestDocument` and a `Namespace`? Perhaps the "root" level for a file could be a `Namespace` with that file's full path?

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,102 @@
...
+                test_fn
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
```
jpschorr, line 96: can remove this

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,102 @@
...
+        let result = 2 + 2;
+        assert_eq!(result, 4);
+    }
+
+    // TODO: add tests checking the conversions between test structs and CodeGen functions
```
jpschorr, line 101: This might be a good use for snapshot testing; perhaps https://insta.rs/

For a codegen use case like this, snapshot testing can serve as rudimentary documentation as well. For codegen it's much easier to compare the input data structures and output code than to try to intuit the code that will be output by inspecting tests that only touch edges and pieces of the functionality.

jpschorr: But I think it's fine to leave as a TODO for this initial PR. Let's add another issue to M1.5

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+pub enum ParseAssertions {
+    ParsePass,
+    ParseFail,
+}
+
+impl TestDocument {
+    /// Converts a `TestDocument` into a `Scope`
+    pub fn generate_scope(&self) -> Scope {
```
jpschorr, line 47: Long term, I think it would be better to have the generation not be an impl on the schema objects themselve, but maybe rather on a `Generator` object. This way if additional state is needed it can be present on the `Generator` and doesn't need to be threaded through each call specifically.

But this is fine for this PR; we can refactor as we learn more.

jpschorr (approved):
Some comments and questions. Nothing blocking.

partiql-conformance-test-generator/src/schema.rs
alancai98: I think having the distinction between a top-level/root-level test document and a namespace is helpful to have. It makes mapping the schema objects to the codegen objects slightly clearer imo. (i.e. mapping `TestDocument` to a codgen's [scope](https://docs.rs/codegen/latest/codegen/struct.Scope.html) and `Namespace` to codgen's [module](https://docs.rs/codegen/latest/codegen/struct.Module.html)).

Also I think modeling root-level as a `Namespace` could be misused (e.g. including a root-level file as some other root-file's `Namespace`).

alancai98:
https://github.com/partiql/partiql-lang-rust/pull/75/commits/2c52b7ba5500e67fe49aa8083db3f6220500cee7 and https://github.com/partiql/partiql-lang-rust/pull/75/commits/856e8508c27fd4bf9c52312f661279a3653738ad were new commits added along with merge from `main`. Changes from these commits:
- Created a `Generator` struct that contains the actual conversion from schema to codegen objects (was previously all in `schema.rs`)
- Changes the generated file structure to replicate the source `partiql-tests`' file structure.
  - Previously, had used the paths as the filenames, which led to long names
  - Now, models the same file structure, escapes the directories/filenames (e.g. '-' is not allowed as a module name), and adds `mod.rs` files for each directory
- Adds a newline to end of generated files to get rid of `cargo fmt` error

partiql-conformance-tests/tests/fail_parse_select-from-where.rs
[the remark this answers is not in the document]
am357: > Regarding checking the tests in, I agree with the comments. Since there may be a lot of tests in the future, I'll not include the tests. If we find use in having the tests checked in and visible (e.g. for checking the GH actions logs), I can always check them in.

The question is: does the fact that we're going to have a lot of tests in the future is sufficient for us to skip test check-in?  Let's discuss about this further (non-blocking for this PR).

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,42 @@
+/// Conformance test document containing namespaces and/or tests
+pub struct TestDocument {
+    pub(crate) namespaces: Namespaces,
+    pub(crate) test_cases: TestCases,
```
am357, line 4: Non-blocking for this PR: If not mistaken, this means we can have test cases in the root of a test document which translates to having tests directly under a `Scope`; does it mean that we can have tests that belong to no namespace? or we are considering an implicit root namespace.

partiql-conformance-test-generator/src/schema.rs
```diff
@@ -0,0 +1,42 @@
+/// Conformance test document containing namespaces and/or tests
```
am357, line 1: Non-blocking for this PR: Does having a TODO that references the test ISL and later on schema code generation based on that makes sense here?

am357 (approved):
LGTM w/ some non-blocking comments.

partiql-conformance-test-generator/src/schema.rs
alancai98: >If not mistaken, this means we can have test cases in the root of a test document which translates to having tests directly under a Scope

Correct.

> does it mean that we can have tests that belong to no namespace? or we are considering an implicit root namespace.

For now, I'm considering `TestDocument` as a "root"/"top"-level namespace.

partiql-conformance-test-generator/src/schema.rs
alancai98: I can add a TODO linking to the test ISL once it's added to `partiql-tests`. Also will add a TODO for schema code generation based on ISL.

partiql-conformance-test-generator/src/generator.rs
```diff
@@ -0,0 +1,63 @@
...
+    module
+}
+
+/// Converts a test case into a testing `Function`
+fn test_case_to_function(test_case: &TestCase) -> Function {
+    match &test_case.test_kind {
+        Parse(ParseTestCase { parse_assertions }) => {
+            let mut test_fn: Function = Function::new(&test_case.test_name);
+            test_fn.attr("test").line(format!(
+                "let parse_result = partiql_parser::parse_partiql(\"{}\");",
```
alancai98, line 47: Discovered in some tests I added to `partiql-tests` that I need to escape the quotes here. Possibly using a [raw string literal](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals). Will fix in an upcoming commit.

alancai98: Fixed in https://github.com/partiql/partiql-lang-rust/pull/75/commits/b5ab2b2ccf960636ee232d0fcebfe0ae52565a59

am357 (approved):

jpschorr (approved):

partiql-conformance-test-generator/src/lib.rs
```diff
@@ -0,0 +1,188 @@
...
+use std::collections::HashSet;
+use std::ops::Add;
+
+/// Converts a vector of Ion data into a `TestDocument`, which can be composed of `Namespace`s
+/// and `TestCase`s. `Namespace`s must be provided as IonLists while `TestCase`s must be provided
+/// as IonStructs. Other Ion types will result in a panic.
+///
+/// When encountering a duplicate namespace/test case, a namespace/test case will be added with
+/// '_0' suffixed to the end of the name.
+pub fn ion_data_to_test_document(all_ion_data: Vec<OwnedElement>) -> TestDocument {
```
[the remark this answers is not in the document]
alancai98: Yeah makes sense. Will move to its own file in a future PR. Added a TODO for now.

alancai98 (commented):
> Nit: should we move them to a backup dir first and delete once the generation is successful?

I can look into moving it into a backup directory (e.g. some subdirectory of the `build/` folder). Though I haven't encountered a situation yet where I've needed to use the old generated tests during a failed build. Added a TODO for now.

am357:
> I can look into moving it into a backup directory (e.g. some subdirectory of the build/ folder). Though I haven't encountered a situation yet where I've needed to use the old generated tests during a failed build. Added a TODO for now.

Makes sense; I was considering a situation in which we run the build, we remove the dir and then we get a build failure. This causes us to loose the tests so we need to revert our changes to re-generate the tests in case we want to look into those.

Issue #77, opened 2022-03-24
Improve the Ergonomics and Documentation of the Parser API


jpschorr:
After #76 (Decide which prototype parser to use), we should be able to clean up the various bits of the external Parser API.

Potential sub-tasks:
 - Wrap all parser-generator errors in our own exported `Error` type
   - See also #18 & #16
 - Improve the exported `Result` type to include the `Error` type
 - Decide whether Lexing is an exported feature of the parser

PR #96, opened 2022-04-04
Refactor ParserError

### LINKED ISSUE #88: Add additional error recovery points to the parser.
PR #86 adds error recovery to `TableReference` (essentially for each thing that can be a "FROM source"). This PR is to demonstrate the ability and put the rest of the necessary stuff in place for multiple errors.

I think we should likely explore other such failure points for each possible clause of SFW, and likely for ExprTerm as well.

It'll be nice to be able to catch multiple parse errors per parse.

_Originally posted by @jpschorr in https://github.com/partiql/partiql-lang-rust/pull/86#discussion_r840975708_

### LINKED ISSUE #89: Handle all of LALRPOP's ParseError types.
> You could add an issue to track adding `expected`. And could reference at this TODO: https://github.com/partiql/partiql-lang-rust/blob/dcef86525302fdc990688b8c2220717d92d22376/partiql-parser/src/lalr/mod.rs#L91

_Originally posted by @alancai98 in https://github.com/partiql/partiql-lang-rust/pull/86#discussion_r840963898_

Really, we should address all of the `todo`s in https://github.com/partiql/partiql-lang-rust/blob/dcef86525302fdc990688b8c2220717d92d22376/partiql-parser/src/lalr/mod.rs#L87-L103

### LINKED ISSUE #90: `Located`'s `Loc` generic should likely be constrained by `Display`, not `Debug`
This will involve implementing `Display` for at least `BytePosition`, `LineAndCharPosition`, and `LineAndColumn`

jpschorr:
#77
FIxes #88
FIxes #89
FIxes #90

- Move LALRPOP's error recovery location to ExprTerm. (Adding it in other clauses and in ExprTerm leads to grammar ambiguities)
- Move `LineOffsetTracker` into `partiql-source-map` crate
- Refactor `Location` to be `Display`able
- Refactor `ParserError` to handle all of LALRPOP's internal error types
- Refactor `ParserError` to use `Cow` instead of private `Token` or bare `String` types
- Refactor `ParserError` to be `Display`able


partiql-parser/src/lalr/lexer.rs
```diff
@@ -1,152 +1,25 @@
-use partiql_source_map::location::{ByteOffset, BytePosition, LineAndCharPosition, LineOffset};
+use partiql_source_map::location::{ByteOffset, BytePosition, ToLocated};
+use std::borrow::Cow;
```
alancai98, line 2: I haven't seen Cow (clone-on-write) pointers before. In what situations are they used?

alancai98 (approved):

partiql-parser/src/lalr/mod.rs
```diff
@@ -53,26 +50,35 @@ pub fn parse_partiql(s: &str) -> ParserResult {
 
     let parsed: LalrpopResult = grammar::QueryParser::new().parse(s, &mut errors, lexer);
 
+    process_errors(s, &offsets, parsed, errors)
+}
+
+pub fn process_errors<'input, T>(
```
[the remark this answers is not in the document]
jpschorr: Nope, not supposed to be `pub`; I'll remove.

partiql-parser/src/lalr/lexer.rs
jpschorr: For the usage here (basically either a `str` or a `String`), see https://jwilm.io/blog/from-str-to-cow/

partiql-parser/src/lalr/mod.rs
```diff
@@ -409,44 +411,63 @@ mod tests {
...
+                        .into(),
                     },
                 })
-            ));
+            );
+        }
+
+        #[test]
+        fn eof() {
+            let res = parse_partiql(r#"SELECT"#);
```
alancai98, line 466: Interestingly, PostgreSQL doesn't give an error for this query:

```
postgres=# SELECT;
--
(1 row)
```
While MySQL and SQLite give an error. I think it makes sense to give an error in this case

partiql-parser/src/result.rs
```diff
@@ -2,33 +2,62 @@
...
+    Loc: Display,
 {
     /// Indicates that there was a problem with syntax.
-    #[error("Syntax Error: {} at [{:?}]", _0.inner, _0.location)]
+    #[error("Syntax Error: {} at `{}`", _0.inner, _0.location)]
     SyntaxError(Located<String, Loc>),
 
+    /// There were not enough tokens to complete a parse
+    #[error("Unexpected end of input")]
+    UnexpectedEndOfInput,
```
alancai98, line 48: I saw the `Loc` argument was removed in the latest commit. Is there any value in having this argument for this error? I guess it will always point beyond the last parsed token.

PostgreSQL, for example, points out a syntax error when it expects another token before the end of input.
```
postgres=# SELECT * FROM foo WHERE;
ERROR:  syntax error at or near ";"
LINE 1: SELECT * FROM foo WHERE;
                               ^
postgres=#
```

alancai98 (approved):
Had a couple non-blocking questions. Changes since first review look good otherwise

partiql-parser/src/lalr/mod.rs
jpschorr: I suppose you could interpret `SELECT` as the zero-column projection from the zero-row collection. Outside of conceptual completeness, I can't think of the utility of this now -- and we could always add it later.

partiql-parser/src/result.rs
jpschorr: Dealing with the indices of the location outside of the source just got too tricky for the library and it's customers; I'd rather be explicit and say the input was exhausted unexpectedly.

Issue #79, opened 2022-03-24
[conformance-testing] Create GitHub Actions workflow to report conformance test suite results


alancai98:
Involves creating a workflow that runs the conformance tests and reports back the failing and passing tests. Should also compare the results with the target branch (similar to codecov comparing coverage).

alancai98:
Would be good for the generated conformance report to also show which tests pass or fail between runs. The `cargo test` outputs the number of passing and failing but looking at the delta for these numbers could be misleading as there could be some tests that previously passed but are now failing.

PR #107, opened 2022-04-26
Update `partiql-tests` submodule; add feature to `partiql-conformance-tests` crate

alancai98:
Changes as part of this PR:
- Updates `partiql-tests` submodule following latest test additions in https://github.com/partiql/partiql-tests/pull/10.
- Disables running the conformance tests by default (i.e. `cargo test`) and adds a feature flag to `partiql-conformance-tests` that can be set to run the conformance tests
  - More details included in the `partiql-conformance-tests` `README.md`
- Adds a GitHub actions workflow step that runs the conformance tests separately using the [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) flag
  - This will allow the GitHub Actions build to succeed even if a conformance test fails
  - Further conformance report generation will be done as part of #79

### DIFF
--- a/.github/workflows/ci_build_test.yml
+++ b/.github/workflows/ci_build_test.yml
@@ -36,11 +36,21 @@ jobs:
         with:
           command: build
           args: --verbose --workspace
-      - name: Cargo Test
+      - name: Cargo Test excluding conformance tests
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --verbose --workspace
+      # Currently just runs the conformance tests and won't fail the build for a conformance test failure (more details
+      # on the 'continue-on-error' workflow syntax can be found here:
+      # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error
+      # Further conformance report generation will be part of https://github.com/partiql/partiql-lang-rust/issues/79.
+      - name: Cargo Test of the conformance tests (can fail)
+        uses: actions-rs/cargo@v1
+        continue-on-error: true
+        with:
+          command: test
+          args: --verbose --package partiql-conformance-tests --features "conformance_test"
       - name: Rustfmt Check
         uses: actions-rs/cargo@v1
         with:
--- a/.github/workflows/ci_build_test.yml
+++ b/.github/workflows/ci_build_test.yml
@@ -41,10 +41,13 @@ jobs:
         with:
           command: test
           args: --verbose --workspace
-      # Currently just runs the conformance tests and won't fail the build for a conformance test failure (more details
-      # on the 'continue-on-error' workflow syntax can be found here:
+      # Currently just runs the conformance tests. Further conformance report generation will be part of
+      # https://github.com/partiql/partiql-lang-rust/issues/79.
+      # Uses the 'continue-on-error' workflow syntax which means the GitHub Actions build will succeed even if the
+      # workflow step failed. Results are still viewable (example run:
+      # https://github.com/partiql/partiql-lang-rust/runs/6186121711?check_suite_focus=true#step:8:686)
+      # More details on the 'continue-on-error' workflow syntax can be found here:
       # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error
-      # Further conformance report generation will be part of https://github.com/partiql/partiql-lang-rust/issues/79.
       - name: Cargo Test of the conformance tests (can fail)
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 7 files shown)


alancai98:
CI fails due to some `rustfmt` issues, which will be fixed in an upcoming commit. `cargo test` works as expected (i.e. doesn't run the conformance tests).

Will add a workflow section for running the conformance tests.

partiql-conformance-tests/build.rs
```diff
@@ -46,6 +46,18 @@ fn main() -> io::Result<()> {
...
 
-    dir_to_mods(tests_path);
+    let sub_tests_dir = "partiql_tests";
+    let sub_tests_path = Path::new(sub_tests_dir);
+    dir_to_mods(&tests_path.join(sub_tests_path));
+    File::create(&tests_path.join("mod.rs"))
+        .expect("mod.rs created in root test folder")
+        .write_all(
+            format!(
+                "#[cfg(feature = \"conformance_test\")]\n#[rustfmt::skip]\nmod {};\n",
```
alancai98, line 56: Adds the "conformance_test" flag to the top-level `mod.rs` file in the `tests/` directory. 

And also skips the rustfmt checks for the generated files. `rustfmt` was complaining in a previous [GH Actions run](https://github.com/partiql/partiql-lang-rust/runs/6183724158?check_suite_focus=true#step:8:47) about the order of modules in `mod.rs` and about reformatting longer lines.

.github/workflows/ci_build_test.yml
```diff
@@ -36,11 +36,21 @@ jobs:
...
         with:
           command: test
           args: --verbose --workspace
+      # Currently just runs the conformance tests and won't fail the build for a conformance test failure (more details
+      # on the 'continue-on-error' workflow syntax can be found here:
+      # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error
+      # Further conformance report generation will be part of https://github.com/partiql/partiql-lang-rust/issues/79.
+      - name: Cargo Test of the conformance tests (can fail)
+        uses: actions-rs/cargo@v1
+        continue-on-error: true
```
alancai98, line 50: Using `continue-on-error` means the GH Actions build will still succeed even if the workflow step failed. Results are still viewable (example [run](https://github.com/partiql/partiql-lang-rust/runs/6186121711?check_suite_focus=true#step:8:686)).

am357: How about adding the same comment to the code.

partiql-conformance-tests/README.md
```diff
@@ -0,0 +1,20 @@
...
+---
+
+Running an individual test (or subset of tests) can vary by the IDE being used. Using CLion, you may need to first edit
+the test run configuration and enable the "Use all features in test" checkbox or explicitly add the 
+`--features "conformance_test"` test option.
+
+Using the command line, you can run an individual test with the following:
+```shell
+cargo test --package partiql-conformance-tests --test <test name or full mod path> --features "conformance_test" -- --exact
+```
```
am357, line 20: Could we also add the same to the repo's README?

partiql-conformance-tests/build.rs
am357: Perhaps this is also comment worthy in the code.

am357 (approved):
Overall LGTM. Left some comments with regards to doc/commenting in the code.

partiql-conformance-tests/build.rs
jpschorr: Or we should run `rustfmt` on the generated code

partiql-conformance-tests/build.rs
```diff
@@ -46,6 +46,18 @@ fn main() -> io::Result<()> {
...
+                sub_tests_dir
+            )
+            .as_bytes(),
+        )
+        .unwrap_or_else(|error| panic!("Failure when writing to file: {:?}", error));
```
jpschorr, line 61: `unwrap_or_else` with a panic inside can be replaced with `expect`: https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

jpschorr (approved):

partiql-conformance-tests/build.rs
alancai98: Changed to run `rustfmt` on the generated code.

partiql-conformance-tests/build.rs
alancai98: Changed to use `expect` at a few places

am357 (approved):

Issue #108, opened 2022-04-29
Resolve Parser conformance test failures


am357:
Before considering this milestone as complete we need to ensure all parser conformance tests defined in this milestone are passing.

### DoD
1. All conformance tests are passing.

### References
- [Test Report](https://paste.debian.net/hidden/7cfc487d/)

PR #109, opened 2022-04-29
Resolve parser conformance tests (initial PR)

am357:
*Issue #, if available:* #108

Resolves `10` case failures.

*Description of changes:*

Before ([full report](https://paste.debian.net/hidden/7cfc487d/))
```
cargo test --package partiql-conformance-tests --features "conformance_test"
...

test result: FAILED. 76 passed; 87 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

After:
```
cargo test --package partiql-conformance-tests --features "conformance_test"
...

test result: FAILED. 86 passed; 77 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

### DIFF
--- a/partiql-ast/src/ast.rs
+++ b/partiql-ast/src/ast.rs
@@ -481,13 +481,13 @@ pub struct Path {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Call {
     pub func_name: SymbolPrimitive,
-    pub args: Vec<Box<Expr>>,
+    pub args: Option<Vec<Box<Expr>>>,
 }
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct CallAgg {
     pub func_name: SymbolPrimitive,
-    pub args: Vec<Box<Expr>>,
+    pub args: Option<Vec<Box<Expr>>>,
 }
 
 #[derive(Clone, Debug, PartialEq)]
--- a/partiql-ast/src/ast.rs
+++ b/partiql-ast/src/ast.rs
@@ -217,29 +217,40 @@ pub struct Expr {
 }
 
 /// Represents an AST Node of type T with BytePosition Location
-type AstBytePos<T> = AstNode<T, BytePosition>;
+pub type AstBytePos<T> = AstNode<T, BytePosition>;
 
-type LitAst = AstBytePos<Lit>;
-type VarRefAst = AstBytePos<VarRef>;
-type ParamAst = AstBytePos<VarRef>;
-type StructAst = AstBytePos<Struct>;
-type BagAst = AstBytePos<Bag>;
-type ListAst = AstBytePos<List>;
-type SexpAst = AstBytePos<Sexp>;
-type BinOpAst = AstBytePos<BinOp>;
-type UniOpAst = AstBytePos<UniOp>;
-type LikeAst = AstBytePos<Like>;
-type BetweenAst = AstBytePos<Between>;
-type InAst = AstBytePos<In>;
-type SimpleCaseAst = AstBytePos<SimpleCase>;
-type SearchCaseAst = AstBytePos<SearchCase>;
-type UnionAst = AstBytePos<Union>;
-type ExceptAst = AstBytePos<Except>;
-type IntersectAst = AstBytePos<Intersect>;
-type PathAst = AstBytePos<Path>;
-type CallAst = AstBytePos<Call>;
-type CallAggAst = AstBytePos<CallAgg>;
-type SelectAst = AstBytePos<Select>;
+pub type LitAst = AstBytePos<Lit>;
+pub type VarRefAst = AstBytePos<VarRef>;
+pub type ParamAst = AstBytePos<VarRef>;
+pub type StructAst = AstBytePos<Struct>;
+pub type BagAst = AstBytePos<Bag>;
+pub type ListAst = AstBytePos<List>;
+pub type SexpAst = AstBytePos<Sexp>;
+pub type BinOpAst = AstBytePos<BinOp>;
+pub type UniOpAst = AstBytePos<UniOp>;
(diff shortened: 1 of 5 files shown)


partiql-ast/src/ast.rs
```diff
@@ -470,13 +470,13 @@ pub struct Path {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Call {
     pub func_name: SymbolPrimitive,
-    pub args: Vec<Box<Expr>>,
+    pub args: Option<Vec<Box<Expr>>>,
```
jpschorr, line 473: Is having an `Option` here actually adding anything over having an empty `Vec` ?

partiql-parser/src/parse/partiql.lalrpop
```diff
@@ -645,8 +645,14 @@ ExprPair: ast::ExprPair = {
...
 #[inline]
 FunctionCall: ast::Call = {
+    <func_name:"Identifier"> "(" ")" =>
+        ast::Call{ func_name: ast::SymbolPrimitive{ value: func_name.to_owned() }, args: None },
+     <func_name:"Identifier"> "(" "*" ")" =>
+         ast::Call{ func_name: ast::SymbolPrimitive{ value: func_name.to_owned() }, args: None },
     <func_name:"Identifier"> "("  <args:CommaSep<ExprQuery>> ")" =>
-        ast::Call{ func_name: ast::SymbolPrimitive{ value: func_name.to_owned() }, args }
+        ast::Call{ func_name: ast::SymbolPrimitive{ value: func_name.to_owned() }, args: Some(args) },
+    <func_name:"Identifier"> "(" "DISTINCT" <args:CommaSep<ExprQuery>> ")" =>
```
jpschorr, line 654: Everywhere that `Distinct` is allowed like this, `All` is too: https://jakewheat.github.io/sql-overview/sql-92-grammar.html#_6_5_set_function_specification.

This likely points to adding a `SetStrategy` or similar AST node enum that we add to `Call` and `CallAgg` (see also `GroupStrategy`) with `All` as default.

partiql-ast/src/ast.rs
```diff
@@ -470,13 +470,13 @@ pub struct Path {
...
 }
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct CallAgg {
     pub func_name: SymbolPrimitive,
-    pub args: Vec<Box<Expr>>,
+    pub args: Option<Vec<Box<Expr>>>,
```
alancai98, line 479: To confirm my understanding, `CallAgg` refers to an aggregation function call (like `SUM`, `COUNT`, `AVG`, etc?)?

From the SQL92 BNF (https://ronsavage.github.io/SQL/sql-92.bnf.html#general%20set%20function, can also see pg. 123 of [sql92 spec](http://www.contrib.andrew.cmu.edu/~shadow/sql/sql1992.txt)), the aggregation functions appear to only take one argument, so should this be

```rust
pub arg: Box<Expr>,
```

partiql-ast/src/ast.rs
am357: At first went with empty Vec but then thought using Option might make more sense in signaling that we can have functions with no args. Happy to change though, do you see a performance or ergonomics concern?

partiql-parser/src/parse/partiql.lalrpop
am357: Makes sense, will look into it further.

alancai98 (commented):
Thanks for providing the before and after for the test results. May be helpful to include a brief description for the fixed tests since there can be some subsequent PRs fixing more conformance tests.

The output of `cargo test` shows the number of passing/failing tests but as part of the conformance report generation #79, we may need to ensure that tests that were previously passing are not now failing. E.g. if `5` more tests pass due to a change, need to ensure that `5` additional tests are passing and not something like `6` tests now pass but `1` more fails.

partiql-ast/src/ast.rs
jpschorr: The problem is that the SQL spec formulates the BNF so that there isn't ambiguity as to what is an aggregation function and what is a scalar function. But, in practice, implementing a parser is implemented that was is impractical (the number of rules is crazy).

Instead, here, we parse everything as a function call and will patch things up later when we do type check/inference.

Also, we need to be able to support the syntax if a future integrator adds their own multivalued aggregation function via a plugin interface.

But also, we need the AST to be able to represent what's parseable, even if it's wrong. I think I'd rather parse text like `SELECT COUNT(foo, bar)` and be able prepare an error about the aggregation call than to just throw an unexpected `,`.

alancai98: I see. Thanks for the explanation.

partiql-ast/src/ast.rs
jpschorr: I think a vec with no args is preferable. Otherwise we have 2 situations that represent 0 args: `None` and `Some(vec![])`

am357: That's a fair point, will make the change.

am357:
> Thanks for providing the before and after for the test results. May be helpful to include a brief description for the fixed tests since there can be some subsequent PRs fixing more conformance tests.
> 
> The output of `cargo test` shows the number of passing/failing tests but as part of the conformance report generation #79, we may need to ensure that tests that were previously passing are not now failing. E.g. if `5` more tests pass due to a change, need to ensure that `5` additional tests are passing and not something like `6` tests now pass but `1` more fails.

That's a fair point, will do.

am357:
Please disregard this PR as it has been polluted by commits from `main`. I'll create a new one and reference this PR for comment history.

am357:
Closing this PR in favor of the following:
https://github.com/partiql/partiql-lang-rust/pull/116

PR #111, opened 2022-04-30
Add Union/Intersect/Except to parser

jpschorr:
#108 

Add Union/Intersect/Except to parser

Makes the the following pass: 
- https://github.com/partiql/partiql-tests/blob/main/partiql-test-data/pass/parser/primitives/union-except-intersect.ion

### DIFF
--- a/partiql-ast/src/ast.rs
+++ b/partiql-ast/src/ast.rs
@@ -64,12 +64,20 @@ impl<T> ToAstNode for T {}
 /// for creating the node. See [ToAstNode] for more details on the usage.
 ///
 /// [1]: https://crates.io/crates/derive_builder
-#[derive(Builder, Clone, Eq, PartialEq, Debug)]
+#[derive(Builder, Clone, Debug)]
 pub struct AstNode<T, Loc: Display> {
     pub node: T,
     #[builder(setter(strip_option), default)]
     pub location: Option<Location<Loc>>,
 }
+
+impl<T: PartialEq, Loc: Display> PartialEq for AstNode<T, Loc> {
+    fn eq(&self, other: &Self) -> bool {
+        self.node == other.node
+    }
+}
+
+impl<T: Eq, Loc: Display> Eq for AstNode<T, Loc> {}
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Item {
@@ -207,11 +215,6 @@ pub enum ConflictAction {
 }
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct Query {
-    pub expr: Box<Expr>,
-}
-
-#[derive(Clone, Debug, PartialEq)]
 pub struct Expr {
     pub kind: ExprKind,
 }
@@ -233,9 +236,7 @@ pub type BetweenAst = AstBytePos<Between>;
 pub type InAst = AstBytePos<In>;
 pub type SimpleCaseAst = AstBytePos<SimpleCase>;
 pub type SearchCaseAst = AstBytePos<SearchCase>;
-pub type UnionAst = AstBytePos<Union>;
-pub type ExceptAst = AstBytePos<Except>;
-pub type IntersectAst = AstBytePos<Intersect>;
+pub type SetExprAst = AstBytePos<SetExpr>;
 pub type PathAst = AstBytePos<Path>;
 pub type CallAst = AstBytePos<Call>;
 pub type CallAggAst = AstBytePos<CallAgg>;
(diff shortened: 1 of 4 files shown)


partiql-parser/src/parse/partiql.lalrpop
```diff
@@ -15,7 +15,44 @@ use partiql_source_map::location::{ByteOffset, BytePosition, ToLocated};
 grammar<'input, 'err>(input: &'input str,
                       errors: &'err mut Vec<ErrorRecovery<ByteOffset, lexer::Token<'input>, ParseError<'input, BytePosition>>>);
 
+
+// ------------------------------------------------------------------------------ //
+//                                                                                //
+//                            UNION/INTERSECT/EXCEPT                              //
+//                                                                                //
```
am357, line 22: Not a blocker: How about adding a note about the higher precendence of `INTERSECT`?

partiql-parser/src/parse/mod.rs
```diff
@@ -372,6 +372,13 @@ mod tests {
             "#;
             parse!(q)
         }
+
+        #[test]
+        fn set_ops() {
+            parse!(
+                r#"SELECT * FROM a LIMIT 10 OFFSET 2 UNION SELECT * FROM b INTERSECT c EXCEPT SELECT * FROM d"#
```
am357, line 379: Not a blocker: Can we also add a comment on what this equivalent to? If not mistaken it should evaluate to the following?
`SELECT * FROM a LIMIT 10 OFFSET 2 UNION (SELECT * FROM b INTERSECT c EXCEPT SELECT * FROM d)`

am357 (approved):
LGTM w/ couple of non-blocking comments.

partiql-parser/src/parse/partiql.lalrpop
```diff
@@ -15,7 +15,44 @@ use partiql_source_map::location::{ByteOffset, BytePosition, ToLocated};
...
+Query01: Box<ast::Expr> = {
+    <lo:@L> <lhs:Query01> "INTERSECT" <setq:SetQuantifier> <rhs:SingleQuery>  <hi:@R> => {
+        Box::new( ast::Expr{ kind: ast::ExprKind::Intersect(ast::Intersect{setq, lhs, rhs}.ast(lo..hi)) } )
+	},
+    <SingleQuery>
+}
+
+SetQuantifier: ast::SetQuantifier = {
+    "DISTINCT" => ast::SetQuantifier::Distinct,
+    "ALL"? => ast::SetQuantifier::All,
```
alancai98, line 47: I'm still trying to understand the syntax/semantics for lalrpop. What does the `?` denote? And should it be the same for both `DISTINCT` and `ALL`?

alancai98 (approved):
Left a question about some LALRPOP syntax. Rest looks good to me

partiql-parser/src/parse/mod.rs
alancai98: On a related note, it may be worth adding a TODO to add more operator precedence tests. The Kotlin implementation has some operator precedence tests: https://github.com/partiql/partiql-lang-kotlin/blob/34625c68dbcbaf7b8ae60df7a4cf65c60b2a3b79/lang/test/org/partiql/lang/syntax/SqlParserPrecedenceTest.kt.

I decided not to add them to the conformance test suite at this time since we don't have any ast checks currently.

jpschorr: We should look into these tests.

Tying it to the structure of the AST isn't really necessary. 

We should be able to assert the "equality of the AST" of two queries for the purposes of precendence testing.
e.g. `a union b union c` == `(a union b) union c`

partiql-parser/src/parse/partiql.lalrpop
jpschorr: LALRPOP refers to it as a macro.

 Four macros are built-in and have regex-like semantics: `?` is optional, `+` is one or more, `*` is zero or more, `(`...`)` is grouping. These correspond to returning the rust types `?` => `Option<T>`, `+`, `*` => `Vec<T>`

http://lalrpop.github.io/lalrpop/tutorial/006_macros.html

jpschorr:
Requesting a re-review on this, as I made significant changes to bring precedence in line with the spec.

For example, the following two should be equivalent:
`SELECT a FROM b UNION SELECT x FROM y LIMIT 10`
`(SELECT a FROM b UNION SELECT x FROM y) LIMIT 10`

In order to apply the `LIMIT` to the rightmost `SELECT`, one would need:
`SELECT a FROM b UNION (SELECT x FROM y LIMIT 10)`

am357 (approved):
LGTM

partiql-parser/src/parse/mod.rs
```diff
@@ -378,7 +380,42 @@ mod tests {
...
+        }
+
+        #[test]
+        fn limit() {
+            let l = parse!(r#"SELECT a FROM b UNION SELECT x FROM y LIMIT 10"#);
+            let r = parse!(r#"(SELECT a FROM b UNION SELECT x FROM y) LIMIT 10"#);
+            assert_eq!(l, r);
+            let r2 = parse!(r#"SELECT a FROM b UNION (SELECT x FROM y LIMIT 10)"#);
+            assert_ne!(l, r2);
+            assert_ne!(r, r2);
```
alancai98, line 418: nit: could also include tests for `ORDER BY`, `OFFSET`, and a combination of `ORDER BY`, `LIMIT`, and `OFFSET` using the set ops

partiql-ast/src/ast.rs
```diff
@@ -251,6 +252,39 @@ pub type GroupByExprAst = AstBytePos<GroupByExpr>;
...
+    pub limit: Option<Box<Expr>>,
+    pub offset: Option<Box<Expr>>,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum QuerySet {
+    SetOp(Box<SetExprAst>),
+    Select(Box<SelectAst>),
+    Expr(Box<Expr>),
+    Query(Box<QueryAst>),
```
alancai98, line 271: What is this last `enum` variant, `Query` used for?

alancai98 (approved):
Looks good. Just left a question and a suggestion for more set op parsing tests

PR #127, opened 2022-06-07
Add PartiQL Playground proof of concept (POC)

am357:
*Issue #, if available:* #128 

*Description of changes:*
This change adds the PartiQL Playground proof of concept and includes:
- Enable JSON serialization on PartiQL AST and Parser result.

- Adds the new `partiql-playground` package.
- Add the required Front-end work for the POC:
  - Putting a layout in place.
  - Use already available ace-editor PartiQL mode.
  - Add collapsable JSON and Graph representation.


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "partiql-conformance-tests/partiql-tests"]
 	path = partiql-conformance-tests/partiql-tests
 	url = https://github.com/partiql/partiql-tests.git
+[submodule "partiql-playground/ace-builds"]
```
am357, line 4: We have an option of using `npm` but I thought this is an overkill at this stage as we might move the front-end code out of this package, hence decided to use `submodule` for now. Happy to discuss this further if `npm` is favorable.

am357:
The `diff` of this PR is large because we're using some `js` and `css` dependencies as HTML assets.

partiql-playground/README.md
```diff
@@ -0,0 +1,42 @@
...
+```bash
+python3 -m http.server
+```
+4. On your browser go to `http://http://localhost:8000/`
+
+## Development
+`PartiQL Playground` uses [WebAssembly (Wasm)](https://webassembly.org/) for integrating the front-end with PartiQL Rust back-end.
+Considering this, upon any changes to the package's Rust dependencies (E.g. `partiql-parser`) or the wasm code under `./src/lib` of this package, you need to rebuild the Wasm package using the following command from the root of this package:
+```bash
+wasm-pack build --target web
```
alancai98, line 28: Could add an installation step for `wasm-pack`. I followed the prerequisites from the repo: https://github.com/rustwasm/wasm-pack#-prerequisities

partiql-playground/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
```
alancai98, line 1: Getting a warning to add `LICENSE` file for this package 

```
[INFO]: License key is set in Cargo.toml but no LICENSE file(s) were found; Please add the LICENSE file(s) to your project directory
```

.gitmodules
alancai98: Just curious what would be involved to use `npm` rather than a git submodule?

partiql-playground/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+1. Pull down the `partiql-rust` package from GitHub:
+```bash
+git clone --recursive https://github.com/partiql/partiql-lang-rust.git
+```
+2. Enter the `partiql-playground` root directory:
+```bash
+cd partiql-lang-rust/partiql-playground
+```
+3. Start a webserver from the root directory, as an example, you can use [Python's SimpleHTTPServer](https://docs.python.org/3.8/library/http.server.html#http.server.SimpleHTTPRequestHandler):
```
alancai98, line 18: Btw, I had to run the `wasm-pack build --target web` step from the "Development" section below to get the query editor to show up. Might be worth including as a step before starting the server.

alancai98 (commented):
Really cool work here! I especially like the syntax highlighting of the editor and the AST output in JSON. The AST graph offers a nice visualization of the parsed query, though I found it a bit hard to navigate (I'm not sure if I was using the correct controls).

Had a couple questions for the features included in the front-end:
1. What does the `SAVE` button do? It didn't appear to be doing anything when I clicked it. Is it supposed to `SAVE` the PartiQL statement(s) to a file? This [PostgreSQL online editor](https://onecompiler.com/postgresql) has a "Download" button that allows exporting the PostgreSQL statements to a file.
2. I noticed there was a version drop-down. Is the intention here to allow users to play around with different PartiQL Rust versions?

Left a few comments for some typos and further clarifications. Otherwise looks great!

.gitmodules
am357: If not mistaken, we need to define the project as a `Node.js` project and then change the entry point to node like `app.js`.

am357:
Thank you

>     1. What does the `SAVE` button do? It didn't appear to be doing anything when I clicked it. Is it supposed to `SAVE` the PartiQL statement(s) to a file? This [PostgreSQL online editor](https://onecompiler.com/postgresql) has a "Download" button that allows exporting the PostgreSQL statements to a file.

You're right, the `Save` button is no-op at the moment, I just put it there to show-case the capability. In terms of functionality I'm thinking of being the same as `Save` in [db-fiddle.com](https://www.db-fiddle.com/)—I think it makes to disable it with a tool-tip to describe the functionality.

>     2. I noticed there was a version drop-down. Is the intention here to allow users to play around with different PartiQL Rust versions?

Correct, again, the same concept as  [db-fiddle.com](https://www.db-fiddle.com/).

partiql-playground/Cargo.toml
am357: Added.

partiql-playground/README.md
am357: Ah, that's a miss from my side. Added.

partiql-playground/README.md
am357: Added.

alancai98 (approved):
Looks good! I just have a minor comment for the playground-demo gif name. Should it be "playground-demo-06-13-2022" rather than "...-2020"?

Also the GitHub actions seems to fail due to a minor rustfmt issue

```
 use crate::location::{ByteOffset, BytePosition, LineAndCharPosition, LineOffset};
+use serde::{Deserialize, Serialize};
 use smallvec::{smallvec, SmallVec};
 use std::ops::Range;
-use serde::{Deserialize, Serialize};
```

PR #134, opened 2022-06-09
Make SELECT conformance tests pass

am357:
*Issue #, if available:* #133 #108 

*Description of changes:*
This PR includes the required changes to ensure the current
SELECT conformance tests pass.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #136, opened 2022-06-11
Experimental REPL PoC

jpschorr:
This is a minimal PoC for the start of a REPL. It should be considered experimental, subject to change, etc.

I wrote this in order to test the parser interface & types from the perspective of an external application. As a result, some minor tweaks were made around error reporting. Potentially a refactor of the mietter::Diagnostic stuff in this PR will be included in the parser crate in the future.

It currently accepts no commands, assuming any/all input is a PartiQL query, which it will attempt to parse. Parse errors are pretty printed to the output.

This uses:
- [miette](https://crates.io/crates/miette) for errors
- [rustyline](https://crates.io/crates/rustyline) for readline/editing/history
- [syntect](https://crates.io/crates/syntect) for syntax highlighting via the included sublime `ion` and `partiql` syntax definitions

Demo:
demo

### DIFF
--- a/partiql-cli/src/bin/partiql-cli.rs
+++ b/partiql-cli/src/bin/partiql-cli.rs
@@ -1,0 +1,260 @@
+#![deny(rustdoc::broken_intra_doc_links)]
+
+use partiql_parser::ParseError;
+use rustyline::completion::Completer;
+use rustyline::config::Configurer;
+use rustyline::highlight::Highlighter;
+use rustyline::hint::{Hinter, HistoryHinter};
+
+use rustyline::validate::{ValidationContext, ValidationResult, Validator};
+use rustyline::{ColorMode, Context, Helper};
+use std::borrow::Cow;
+use std::fs::{File, OpenOptions};
+
+use std::path::Path;
+
+use syntect::easy::HighlightLines;
+use syntect::highlighting::{Style, ThemeSet};
+use syntect::parsing::{SyntaxDefinition, SyntaxSet, SyntaxSetBuilder};
+use syntect::util::as_24_bit_terminal_escaped;
+
+use miette::{Diagnostic, LabeledSpan, Report, SourceCode};
+use owo_colors::OwoColorize;
+use partiql_source_map::location::{BytePosition, Location};
+use thiserror::Error;
+
+static ION_SYNTAX: &str = include_str!("ion.sublime-syntax");
+static PARTIQL_SYNTAX: &str = include_str!("partiql.sublime-syntax");
+
+struct PartiqlHelperConfig {
+    dark_theme: bool,
+}
+
+impl PartiqlHelperConfig {
+    pub fn infer() -> Self {
+        const TERM_TIMEOUT_MILLIS: u64 = 20;
+        let timeout = std::time::Duration::from_millis(TERM_TIMEOUT_MILLIS);
+        let theme = termbg::theme(timeout);
+        let dark_theme = match theme {
+            Ok(termbg::Theme::Light) => false,
+            Ok(termbg::Theme::Dark) => true,
+            _ => true,
+        };
+        PartiqlHelperConfig { dark_theme }
+    }
+}
+struct PartiqlHelper {
+    config: PartiqlHelperConfig,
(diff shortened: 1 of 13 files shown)


partiql-cli/src/bin/partiql-cli.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Debug, Error)]
+pub enum CLIError {
+    #[error("PartiQL syntax error:")]
+    SyntaxError {
+        src: String,
+        msg: String,
+        loc: Location<BytePosition>,
+    },
+    #[error("Internal Compiler Error - please report this.")]
```
am357, line 115: Maybe link to the GitHub repo issue?

partiql-cli/src/bin/partiql-cli.rs
```diff
@@ -0,0 +1,244 @@
...
+                ))))
+            }
+            CLIError::InternalCompilerError { .. } => None,
+        }
+    }
+}
+
+impl CLIError {
+    pub fn from_parser_error(err: ParseError, source: &str) -> CLIError {
+        match err {
```
am357, line 143: Do we need the `InvalidInput` error too?

partiql-cli/src/bin/partiql-cli.rs
```diff
@@ -0,0 +1,244 @@
...
+        let source = ctx.input();
+        let result = parser.parse(source);
+        match result {
+            Ok(_) => Ok(ValidationResult::Valid(None)),
+            Err(e) => {
+                if e.errors
+                    .iter()
+                    .any(|err| matches!(err, ParseError::UnexpectedEndOfInput))
+                {
+                    Ok(ValidationResult::Incomplete)
```
am357, line 196: From couple errors I've seen, this can be as a result of incomplete Grammar, do we want to provide some hints or perhaps link for cutting an issue if the input appears to be correct. For example [before this PR](https://github.com/partiql/partiql-lang-rust/pull/134/files) we were getting `UnexpectedEndOfInput` for  `SELECT * FROM a AS a CROSS JOIN c AS c AT q`.

am357 (commented):
Left some comments, would also be great to add more documentation to `partiql-cli.rs`.

am357 (approved):

partiql-cli/src/bin/partiql-cli.rs
am357: That's fair, maybe as a follow-on we can add a `README` to emphasize on the PoC.

partiql-cli/src/bin/partiql.sublime-syntax
```diff
@@ -0,0 +1,131 @@
+%YAML 1.2
```
alancai98, line 1: I'm curious if there a better place to store these syntax files so they could be used by other programs. Is there a way to publish the sublime syntaxes (i.e. partiql.sublime-syntax and ion.sublime-syntax) somewhere similar to what you had done for the ACE editor (https://github.com/ajaxorg/ace/pull/4630)? Alternatively, we could have a separate repo for the syntax highlighting modes that could be taken as a submodule of this repo.

partiql-cli/src/bin/partiql-cli.rs
alancai98: +1 a short `README` would be helpful.

alancai98 (approved):
Very cool! I especially like the syntax highlighting along with the syntax error underlining, which makes it easier to see where exactly an error occurred.

Left a non-blocking comment related to the long-term location for the sublime syntax files.

partiql-cli/src/bin/partiql-cli.rs
jpschorr: added minimal README

am357 (approved):

partiql-cli/src/bin/partiql.sublime-syntax
jpschorr: Yes. We should likely have all the tooling somewhere else and bring it in via submodule or something else here. The official way add a sublime extension is via https://packagecontrol.io/docs/submitting_a_package

PR #155, opened 2022-07-21
Make usage of `serde` optional behind feature flag `serde`

jpschorr:
- Make usage of `serde` optional behind feature flag `serde`
- exclude CLI and playground from main workspace
- use more broadly compatible version requirement for `serde`

---

### DIFF
--- a/.github/workflows/ci_build_test.yml
+++ b/.github/workflows/ci_build_test.yml
@@ -112,7 +112,7 @@ jobs:
         continue-on-error: true
         run: cargo test --verbose --package partiql-conformance-tests --features "conformance_test" -- -Z unstable-options --format json > cargo_test_results.json
       # Create a conformance report from the `cargo test` json file
-      - run: ./target/debug/generate_cts_report cargo_test_results.json ${GITHUB_SHA} cts_report.json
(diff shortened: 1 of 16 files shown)


am357 (approved):

am357 (approved):

alancai98 (approved):
:shipit: I'm curious what are some of the reasons to make `serde` optional behind a feature flag?

jpschorr:
> I'm curious what are some of the reasons to make `serde` optional behind a feature flag?

Making `serde` optional is rather idiomatic across the ecosystem. Serde is not core to the the library, but rather was added for the playground; it's useful, but not core.

A crate's `pub` types and their `impl`ed traits form the public interface. So, if serde is a default feature, then removing it is a sermver breaking change. Behind a feature flag, there is potentially more leeway (at least it's clear to consumers that it is an optional interface that must be opted into)

Serde derived traits (including transitive) potentially add a fair bit to compile time and binary size if not needed.

See:
- https://lurklurk.org/effective-rust/dep-graph.html
- https://lurklurk.org/effective-rust/features.html

am357 (approved):

Issue #166, opened 2022-08-05
Release v0.1.0


am357:
This issue tracks the effort for publishing `partiql-lang-rust` `v0.1.0`.

### DoD
1. GitHub release is done
2. crates.io release is done

PR #172, opened 2022-08-08
(cargo-release) start next development iteration 0.1.1-alpha.0

am357:
*Issue #, if available:* 

#166 

*Description of changes:*

Prepares the version as pre-release for the next version.

```bash
cargo release --dev-version --no-push --no-publish -x
...
```

Since `partiql-cli` and `partiql-playground` are [excluded](https://github.com/partiql/partiql-lang-rust/blob/f99ce48dc3bb64571d80452c705b662302e6dfce/Cargo.toml#L15) from workspace, had to create a change for them manually. For `partiql-playground` we don't intend to release as there is no entry-point to it at this stage. For `partiql-cli` we probably need to include it as workspace `members` so that it can get included in the release. For now I'm adding the change for both until we finalize the decision for the next release.

### DIFF
--- a/partiql-ast/Cargo.toml
+++ b/partiql-ast/Cargo.toml
@@ -13,7 +13,7 @@ exclude = [
   "**/.github/**",
 ]
 edition = "2021"
-version = "0.1.0"
+version = "0.1.1-alpha.0"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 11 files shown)


partiql-ast/Cargo.toml
```diff
@@ -13,15 +13,15 @@ exclude = [
   "**/.github/**",
 ]
 edition = "2021"
-version = "0.1.0"
+version = "0.1.1-alpha.0"
```
alancai98, line 16: I'm curious is this the normal approach for Rust releases (i.e. to include the `-alpha` marker before the next release)? I saw that `ion-rust` and `ion-schema-rust` don't use this procedure

am357: I don't think we can pinpoint a general/normal approach necessarily for Rust as it's more related to the release lifecycle (not entirely certain though). On the other hand, knowing that using `cargo release` automatically adds the suffix in each post-release and in the next release gets removed automatically by the tool, I see this being beneficial for now. I also don't see this as an irreversible change since we're still at the early stages. This is what the tool does:

```bash
1. remove the suffix and publish—e.g. `0.1.1-alpha.0` -> `0.1.1`
2. bump the version w/ the new suffix `0.1.1` -> `0.1.2-alpha.0`
```

alancai98 (approved):

PR #174, opened 2022-08-12
Fix pre-release

am357:
*Issue #, if available:*
#166 

*Description of changes:*

- Moves the following pacakges to workspace excluded:
   - "partiql-conformance-test-generator"
   - "partiql-conformance-tests"

  For both, users can play around with conformance tests using GitHub code.

- Removes [workspace] attribute from `partiql-cli` `Cargo.toml`. This is
  does not seems to be affect partiql-cli build.

- Changes the version back to "0.1.0". This seems to be more aligned with
  other projects as noted by @alancai98 before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,8 +3,6 @@
 members = [
   "partiql",
   "partiql-ast",
-  "partiql-conformance-test-generator",
-  "partiql-conformance-tests",
   "partiql-source-map",
   "partiql-eval",
   "partiql-ir",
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -14,7 +12,9 @@ members = [
 ]
 
 exclude = [
+  "partiql-conformance-test-generator",
```
alancai98, line 15: Moving the packages out of the workspace results in the following error in the GitHub Actions for the "Cargo Test of the conformance tests (can fail) and save to json file" step

```
error: cannot specify features for packages outside of workspace
```

am357: Fair, this is a non-blocker so can move it back for now.

alancai98 (approved):

Issue #176, opened 2022-08-19
[PartiQL Playground] New features


am357:
This issue is for tracking PartiQL playground new feature additions including the following:

- Node.js server listener
- Remove `ace-builds` submodule.
- Containerization
- Client side URL export
- AST Graph pan and zooming bugfix

#### DoD
All the new features are PR'ed and merged.

PR #177, opened 2022-08-19
Move partiql-playground to Node.js

am357:
*Issue #, if available:*

#176 

*Description of changes:*
This is the first PR as part of a series of PR to new features in PartiQL Playground:

This PR includes the following:

- Moves the PartiQL Playground to a Node.js project. This allows
using the Node features within the project, such as having a
server side listerner instead of an arbitrary web-server.

- Moves static assets under `src` directory for better project structure.

- Removes the `ace-builds` sub-modules as an action from previous reviews:
  - For this, it directly consumes `js` files in `partiql-playground/src/js/ace.js` and `partiql-playground/src/js/mode-partiql.js`. This is probably not the best way to handle this but it's a better option than having the sub-modules with unnecessary files. Another option is to enable client side to use `ace-build` which bring more complexity to the project because in that case we either need to use [RequireJs](https://requirejs.org/) or use [Browserify](https://browserify.org/) as two options available.

#### TODO

- [x] Node.js server listener
- [x] Remove `ace-builds` submodule.
- [ ] Containerization
- [ ] REST API
- [ ] Client side URL export
- [ ] AST Graph pan and zooming bugfix
- [ ] update `CHANGELOG.md`


partiql-playground/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "Playground"
+  ],
+  "author": "partiql-team@amazon.com",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/partiql-lang-rust/partiql-playground/issues"
+  },
+  "homepage": "https://github.com/partiql-lang-rust/partiql-playground#readme",
+  "dependencies": {
+    "body-parser": "^1.20.0",
```
am357, line 24: This is required by the subsequent PRs.

partiql-playground/package-lock.json
```diff
@@ -0,0 +1,1069 @@
+{
```
am357, line 1: This is an auto-generated file, but is [required](https://nodejs.dev/en/learn/the-package-lock-json-file) to get check-in—you can ignore reviewing it:

>The package-lock.json file needs to be committed to your Git repository, so it can be fetched by other people, if the project is public or you have collaborators, or if you use Git as a source for deployments.

partiql-playground/src/index.html
```diff
@@ -101,11 +101,11 @@ <h3>The PartiQL Playground</h3>
         }
     }
 </script>
-<script src="./ace-builds/src-noconflict/ace.js" type="text/javascript" charset="utf-8"></script>
+<script src="./js/ace.js"></script>
```
alancai98, line 104: I'm not that familiar with using `ace-builds` but I think it would be better to still use the git submodule rather than to use a copy of the `ace.js` and `mode-partiql.js` files (that may not be updated regularly).

I'm not too concerned with the unneeded files. But I guess if we're concerned with the unneeded files we could do what's suggested in this stack overflow post: https://stackoverflow.com/questions/20985983/git-add-only-some-files-from-remote-repository-as-submodule. Essentially use a branch or fork with many of the unneeded files stripped out. This approach can let us track + merge new upstream changes.

am357: Thanks for the feedback. Agreed, but I'm no a fan of forking or branching since it introduces another problem for keeping the alternate path in sync with upstream; as discussed offline I added the submodule back.

alancai98 (approved):

PR #181, opened 2022-08-30
feat: adds initial crate for static types

almann:


alancai98 (approved):

partiql-types/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "partiql-types"
```
alancai98, line 2: Do you also need to add this crate to the workspace's `Cargo.toml`: https://github.com/partiql/partiql-lang-rust/blob/5d105335c2b35f7f9fd911db6ebd2ce810641eb3/Cargo.toml?

almann: Oh good catch, will add!

am357 (approved):

PR #194, opened 2022-09-12
Give AST nodes identity & decouple source location from AST node.

jpschorr:
- Move the source offset location that gave rise to each AST node into a `LocationMap`
- Add a unique `NodeId` to each AST node

---

### DIFF
--- a/partiql-ast/src/ast.rs
+++ b/partiql-ast/src/ast.rs
@@ -7,85 +7,24 @@
 // As more changes to this AST are expected, unless explicitly advised, using the structures exposed
 // in this crate directly is not recommended.
 
-use partiql_source_map::location::{ByteOffset, BytePosition, Location};
 use rust_decimal::Decimal as RustDecimal;
 
 use std::fmt;
-use std::fmt::Display;
-use std::ops::Range;
 
 #[cfg(feature = "serde")]
 use serde::{Deserialize, Serialize};
 
-/// Provides the required methods for AstNode conversations.
-pub trait ToAstNode: Sized {
-    /// Wraps the `self` to an [AstNode] and returns an `AstNodeBuilder` for
-    /// further [AstNode] construction.
-    /// ## Example:
-    /// ```
-    /// use partiql_ast::ast;
-    /// use partiql_ast::ast::{SymbolPrimitive, ToAstNode};
-    /// use partiql_ast::ast::CaseSensitivity::CaseInsensitive;
-    /// use partiql_source_map::location::{ByteOffset, BytePosition, Location, ToLocated};
-    ///
-    /// let p = SymbolPrimitive {
-    ///     value: "symbol2".to_string(),
-    ///     case: Some(ast::CaseSensitivity::CaseInsensitive)
-    ///  };
-    ///
-    /// let node = p
-    ///     .to_node()
-    ///     .location((BytePosition::from(12)..BytePosition::from(1)).into())
-    ///     .build()
-    ///     .expect("Could not retrieve ast node");
-    /// ```
-    fn to_node(self) -> AstNodeBuilder<Self, BytePosition>
-    where
-        Self: Clone,
-    {
-        AstNodeBuilder::default().node(self).clone()
-    }
-
-    fn to_ast<Loc, IntoLoc>(self, location: IntoLoc) -> AstNode<Self, Loc>
-    where
-        Loc: Display,
-        IntoLoc: Into<Location<Loc>>,
-    {
-        AstNode {
-            node: self,
-            location: Some(location.into()),
-        }
-    }
-
-    fn ast(self, Range { start, end }: Range<ByteOffset>) -> AstNode<Self, BytePosition> {
-        self.to_ast(start.into()..end.into())
-    }
-}
-
-/// Implements [ToAstNode] for all types within this crate, read further [here][1].
-///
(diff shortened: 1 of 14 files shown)


partiql-parser/src/parse/mod.rs
```diff
@@ -32,42 +36,64 @@ type LalrpopResult<'input> = Result<Box<ast::Expr>, LalrpopError<'input>>;
...
+pub(crate) fn parse_partiql(s: &str) -> AstResult {
+    parse_partiql_with_state(s, ParserState::default())
 }
 
-fn process_errors<'input>(
-    _s: &'input str,
-    _offsets: &LineOffsetTracker,
-    result: Result<Box<ast::Expr>, LalrpopError<'input>>,
-    errors: Vec<LalrpopErrorRecovery<'input>>,
+pub(crate) fn parse_partiql_with_state<'input, Id: IdGenerator>(
```
alancai98, line 63: Is it intentional for this to be public? If so, may need some documentation for the function.

partiql-parser/src/parse/parser_state.rs
```diff
@@ -0,0 +1,83 @@
...
+
+const INIT_LOCATIONS: usize = 50;
+
+pub trait IdGenerator {
+    fn id(&mut self) -> NodeId;
+}
+
+pub(crate) struct NodeIdGenerator {
+    next_id: ast::NodeId,
+}
```
alancai98, line 26: Could use some docs for these public apis (`IdGenerator` and `NodeIdGenerator`). Also for `ParserState` on L42

partiql-ast/src/ast.rs
```diff
@@ -7,86 +7,25 @@
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct NodeId(pub u32);
 
-/// Represents an AST node. [AstNode] uses [derive_builder][1] to expose a Builder
-/// for creating the node. See [ToAstNode] for more details on the usage.
-///
-/// [1]: https://crates.io/crates/derive_builder
-#[derive(Builder, Clone, Debug)]
+/// Represents an AST node. See [ToAstNode] for more details on the usage.
```
alancai98, line 21: Second sentence related to `ToAstNode` isn't relevant anymore

partiql-ast/src/ast.rs
```diff
@@ -246,43 +185,40 @@ pub struct Expr {
...
+pub type QuerySetAst = AstNode<QuerySet>;
+pub type SearchedCaseAst = AstNode<SearchedCase>;
+pub type SelectAst = AstNode<Select>;
+pub type SetExprAst = AstNode<SetExpr>;
+pub type SexpAst = AstNode<Sexp>;
+pub type SimpleCaseAst = AstNode<SimpleCase>;
+pub type SortSpecAst = AstNode<SortSpec>;
+pub type StructAst = AstNode<Struct>;
+pub type UniOpAst = AstNode<UniOp>;
+pub type VarRefAst = AstNode<VarRef>;
```
alancai98, line 221: I'm curious how did you determine which parts of the ast derive `Eq`? Among this list (L188-221), only `VarRef` derives `Eq`.

alancai98 (commented):
Looks good. Left some minor comments and questions. Also probably need to update the CHANGELOG and mention that it's a breaking change to the `AstNode` API

partiql-ast/src/ast.rs
jpschorr: It's an oversight (likely due to some automated rewriting) that is fixed in https://github.com/partiql/partiql-lang-rust/pull/195

am357:
> Looks good. Left some minor comments and questions. Also probably need to update the CHANGELOG and mention that it's a breaking change to the `AstNode` API

Echoing Alan, overall LGTM.

partiql-parser/src/parse/mod.rs
jpschorr: Despite the name `pub(crate)` really means "crate-private" and is thus visible to only other functions in the current crate.

jpschorr: But, also, I think it can be module private; will update.

partiql-parser/src/parse/parser_state.rs
jpschorr: I'll add docs.

But these aren't really public apis. Though marked `pub`, they are in a module that is private included in its `super`, so they are public to the `super`, but not the outside world.

You can see what's exported with `cargo doc --open`.

partiql-parser/src/parse/mod.rs
alancai98: > Despite the name `pub(crate)` really means "crate-private" and is thus visible to only other functions in the current crate.

Ah, my mistake. Mixed up the API visibilities with other languages. Thanks for the reference;

alancai98 (approved):

am357 (approved):

partiql-parser/src/parse/mod.rs
jpschorr: Yeah. Rust's API visibility is somewhat expressive, but can be confusing, as the effects are local, and thus you need a global understanding of the crate's structure to understand the public exports.

PR #197, opened 2022-10-19
PoC interpreting evaluator

jpschorr:
PoC basic (though anemic) value type, logical plan (needs to evolve to a graph model), and sketch of an interpreting evaluator (FROM, FROM_AT, UNPIVOT, WHERE, and SELECT)

### DIFF
--- a/partiql-eval/src/env.rs
+++ b/partiql-eval/src/env.rs
@@ -1,0 +1,65 @@
+use std::fmt::Debug;
+
+use partiql_value::{BindingsName, Tuple, Value};
+use unicase::UniCase;
+
+pub trait Bindings<T> {
+    fn get(&self, name: &BindingsName) -> Option<&T>;
+}
+
+impl Bindings<Value> for Tuple {
+    fn get(&self, name: &BindingsName) -> Option<&Value> {
+        match name {
+            BindingsName::CaseSensitive(s) => self.0.get(s),
+            BindingsName::CaseInsensitive(s) => {
+                //TODO
+                self.0.get(s)
+            }
+        }
+    }
+}
+
+pub mod basic {
+    use super::*;
+    use std::collections::HashMap;
+
+    #[derive(Debug)]
+    pub struct MapBindings<T> {
+        sensitive: HashMap<String, usize>,
+        insensitive: HashMap<UniCase<String>, usize>,
+        values: Vec<T>,
+    }
+
+    impl<T> Default for MapBindings<T> {
+        fn default() -> Self {
+            MapBindings {
+                sensitive: HashMap::new(),
+                insensitive: HashMap::new(),
+                values: vec![],
+            }
+        }
+    }
+
+    impl<T> MapBindings<T> {
+        pub fn insert(&mut self, name: &str, value: T) {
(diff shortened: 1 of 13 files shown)


partiql-eval/src/env.rs
```diff
@@ -0,0 +1,65 @@
...
+pub trait Bindings<T> {
+    fn get(&self, name: &BindingsName) -> Option<&T>;
+}
+
+impl Bindings<Value> for Tuple {
+    fn get(&self, name: &BindingsName) -> Option<&Value> {
+        match name {
+            BindingsName::CaseSensitive(s) => self.0.get(s),
+            BindingsName::CaseInsensitive(s) => {
+                //TODO
```
am357, line 15: Other than the fact that we should perhaps treat the case insensitive differently, what do you expect to be as part of `TODO`?

am357 (approved):
Since this is a PoC that enables us to kick-start the work, I don't see a point in reviewing it in details; I had a cursory review and approving it order to get the code in asap.

partiql-eval/src/env.rs
jpschorr: That's the todo. It's not actually doing case sensitivity currently.

partiql-eval/src/eval.rs
```diff
@@ -0,0 +1,469 @@
...
+    fn push_tuple(&mut self, bindings: Tuple, ctx: &dyn EvalContext) {
+        let is_new = self.seen.insert(bindings.clone());
+        if is_new {
+            self.output.push_tuple(bindings, ctx)
+        }
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct EvalOutputAccumulator {
```
alancai98, line 442: Might be a naive question but what is `EvalOutputAccumulator` for? Also, why does is it's only field a `Bag`? Couldn't there be other output types emitted by PartiQL?

partiql-eval/src/plan.rs
```diff
@@ -0,0 +1,110 @@
...
+                expr,
+                as_key,
+                at_key: _,
+                out,
+            }) => Box::new(eval::EvalFrom::new(
+                self.plan_values(expr),
+                &as_key,
+                self.plan_bindings(*out),
+            )),
+            _ => panic!("Unevaluable bexpr"),
```
alancai98, line 33: To check my understanding, only `BindingsExpr::From` is currently supported/implemented in the POC's evaluator planner?

alancai98 (approved):
Also gave a general review of this POC PR (focusing primarily on the `partiql-eval` crate. Left a couple comments to check my understanding. Further questions about the finer details may come up as we continue the implementation of the logical interpreter.

partiql-eval/src/eval.rs
jpschorr: The PoC only models Bindings -> Bindings relations properly.

We will need to build out Bindings -> Value (i.e., Select Value) and more properly model Value -> Bindings (i.e., From, Unpivot) .

Yes, a Value -> Value expression (e.g. `1+2`) at the head will return a single value, not a bag, but is not modelled currently.

PR #202, opened 2022-10-28
[steel thread] Add query plan and eval using DAG model

### LINKED ISSUE #21: Add Macro for Scanner Test Cases

am357:
*Issue #, if available:* 
Closes #21

#### Description of changes
Adds the required changes for enabling query evaluation using a DAG model.

What is our definition of DAG?
We are only interested in DAGs that can be used as execution plans, which leads to the following definition. A DAG is a directed, cycle-free graph G = (V, E) with a denoted root node v0 ∈ V such that all v ∈ V \{v0} are reachable from v0. Note that this is the definition of trees without the condition |E| = |V | − 1. Hence, all trees are DAGs. Reference: https://link.springer.com/article/10.1007/s00450-009-0061-0

#### In addition
- The steel thread includes only `Scan` and `Project` operators, and passes the test for planning and evaluation of the following query:
```SQL
SELECT a AS b FROM data
```
- The code is _not_ clean, maybe not idiomatic at some instances, and lacks documentation. I'll iterate through the code with a consequent PR when adding new operators.

#### Algorithm
We're loosely using a push model with DAG as called out in the following: https://link.springer.com/article/10.1007/s00450-009-0061-0 (Section 10.3.5. Pushing)

1. Create a DAG logical plan `LogicalPlan` with each node as a logical operator (for now manual until we integrate with AST -> PLAN).
1. Create an evaluation plan `EvalPlan` with the same structure as logical plan.
1. [Toposort](https://en.wikipedia.org/wiki/Topological_sorting) the `EvalPlan` to ensure all dependencies of a node get evaluated before the node itself.
1. For each operator in `EvalPlan`, eval the node.
1. Store the result in the node.
1. Push the output to each consumer—for now, for final result, we use a `Sink` node that is considered as the node that accumulates output; we need to see if there are better alternatives.

#### Yet to be implemented
- [ ] port all the other existing nodes and ensure the current tests pass against them
- [ ] add error handling and validation (incl. DAG cyclic validation).
- [ ] add documentation.

Finally, we're not performing any optimization at this stage (and we won't for extended amount of time).

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -5,7 +5,8 @@ use std::fmt::Debug;
 use std::rc::Rc;
 
 use petgraph::algo::toposort;
-use petgraph::{Graph, Incoming, Outgoing};
+use petgraph::{Directed, Incoming, Outgoing};
+use petgraph::prelude::StableGraph;
 
 use partiql_value::Value::{Boolean, Missing, Null};
 use partiql_value::{Bag, BindingsName, Tuple, Value};
@@ -60,7 +61,7 @@ pub struct Sink {
 }
 
 #[derive(Debug)]
-pub struct EvalPlan(pub Graph<EvalOp, ()>);
+pub struct EvalPlan(pub StableGraph::<EvalOp, (), Directed>);
 
 impl Default for EvalPlan {
     fn default() -> Self {
@@ -70,7 +71,7 @@ impl Default for EvalPlan {
 
 impl EvalPlan {
     fn new() -> Self {
-        EvalPlan(Graph::<EvalOp, ()>::new())
+        EvalPlan(StableGraph::<EvalOp, (), Directed>::new())
     }
 }
 
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -5,7 +5,6 @@ use std::fmt::Debug;
 use std::rc::Rc;
 
 use petgraph::algo::toposort;
-use petgraph::graph::NodeIndex;
 use petgraph::{Graph, Incoming, Outgoing};
 
 use partiql_value::Value::{Boolean, Missing, Null};
@@ -26,16 +25,14 @@ pub struct Scan {
     pub expr: Box<dyn EvalExpr>,
     pub as_key: String,
     pub output: Option<Value>,
-    pub consumers: Vec<NodeIndex>,
 }
 
 impl Scan {
-    pub fn new(expr: Box<dyn EvalExpr>, as_key: &str, consumers: Vec<NodeIndex>) -> Self {
+    pub fn new(expr: Box<dyn EvalExpr>, as_key: &str) -> Self {
         Scan {
             expr,
             as_key: as_key.to_string(),
             output: None,
-            consumers,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


am357:
Adding a summary of yesterday's review discussion w/ @alancai98 and @jpschorr:
- Let's move to Trait object to reduce the code as a result of combinatorial logic (addressing in this PR).
- [Follow-up] removing `petgraph` dependency from `Logical` and `Eval` plans.
- [Follow-up] discussion about batching inputs for `Scan`.

partiql-logical/src/lib.rs
```diff
@@ -1,13 +1,24 @@
 use partiql_value::{BindingsName, Value};
+use petgraph::prelude::StableGraph;
+use petgraph::Directed;
 use std::collections::HashMap;
 
-// TODO we should replace this enum with some identifier that can be looked up in a symtab/funcregistry?
 #[derive(Debug)]
+pub struct LogicalPlan(pub StableGraph<BindingsExpr, (), Directed>);
```
jpschorr, line 7: Not to worry while we iterate, but we'll want to remove petgraph from the exposed logical plan soon and replace something like a simple list of nodes and list of edges.

partiql-eval/src/eval.rs
```diff
@@ -1,12 +1,138 @@
...
+}
+
+pub type EvalResult = Result<Evaluated, EvalErr>;
+
+pub struct Evaluated {
+    pub result: Option<Value>,
+}
+
+pub struct EvalErr {
+    pub errors: Vec<EvaluationError>,
```
jpschorr, line 39: To be useful, this will need, at a minimum to implement `std::error::Error`. Probably best to go with something like thiserror: https://docs.rs/thiserror/latest/thiserror/

No need to block this PR on the above, though.

jpschorr (approved):

partiql-eval/src/lib.rs
```diff
@@ -25,16 +28,34 @@ mod tests {
...
         println!("{:?}", &output);
-
         let result = &output.borrow_mut().output;
         result.clone()
     }
 
+    // TODO: rename once we move to DAG model completely
+    fn evaluate_dag(logical: LogicalPlan, bindings: MapBindings<Value>) -> Value {
+        // TODO remove once we agree on using evaluate output in the following PR:
+        // https://github.com/partiql/partiql-lang-rust/pull/202
```
alancai98, line 43: Since we decided to move forward with the DAG model, should we remove output now or will it be removed in an upcoming PR?

partiql-eval/src/eval.rs
```diff
@@ -1,12 +1,138 @@
...
+impl EvalPlan {
+    fn new() -> Self {
+        EvalPlan(StableGraph::<Box<dyn DagEvaluable>, (), Directed>::new())
+    }
+}
+
+pub type EvalResult = Result<Evaluated, EvalErr>;
+
+pub struct Evaluated {
+    pub result: Option<Value>,
```
alancai98, line 35: What would a `None` value for `result` represent?

alancai98 (approved):

partiql-eval/src/lib.rs
am357: It'll be removed in a subsequent PR.

partiql-eval/src/eval.rs
am357: Good call, perhaps it's a remnant of my own refactoring; will re-visit this and reflect the outcome in the subsequent PR.

PR #203, opened 2022-11-01
Add arithmetic ops to evaluator; add Decimal partiql-value

alancai98:
Implements arithmetic ops (+, -, *, /, %; exponentiation is saved for a TODO) in the evaluator and adds a Decimal PartiQL `Value` using the [rust_decimal](https://crates.io/crates/rust_decimal) crate. Also included are some minor fixes to the following:
- Fix to `Ord` related to `OrderedFloat` behavior
- Add `missing`/`null` propagation in binary expressions

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -105,35 +105,56 @@ pub enum EvalBinop {
     Gteq,
     Lt,
     Lteq,
+
+    // Arithmetic ops
+    Add,
+    Sub,
+    Mul,
+    Div,
+    Mod,
+    Exp,
 }
 
 impl EvalExpr for EvalBinopExpr {
     fn evaluate(&self, bindings: &Tuple, ctx: &dyn EvalContext) -> Value {
         let lhs = self.lhs.evaluate(bindings, ctx);
         let rhs = self.rhs.evaluate(bindings, ctx);
-        match self.op {
-            EvalBinop::And => todo!(),
-            EvalBinop::Or => todo!(),
-            EvalBinop::Concat => {
-                // TODO non-naive concat
-                let lhs = if let Value::String(s) = lhs {
-                    *s
-                } else {
-                    format!("{:?}", lhs)
-                };
-                let rhs = if let Value::String(s) = rhs {
-                    *s
-                } else {
-                    format!("{:?}", lhs)
-                };
-                Value::String(Box::new(format!("{}{}", lhs, rhs)))
+        // Missing and Null propagation. Missing has precedence over Null
+        if lhs == Value::Missing || rhs == Value::Missing {
+            Value::Missing
+        } else if lhs == Value::Null || rhs == Value::Null {
+            Value::Null
+        } else {
+            match self.op {
+                EvalBinop::And => todo!(),
+                EvalBinop::Or => todo!(),
+                EvalBinop::Concat => {
+                    // TODO non-naive concat
+                    let lhs = if let Value::String(s) = lhs {
+                        *s
+                    } else {
+                        format!("{:?}", lhs)
+                    };
+                    let rhs = if let Value::String(s) = rhs {
+                        *s
+                    } else {
+                        format!("{:?}", lhs)
+                    };
+                    Value::String(Box::new(format!("{}{}", lhs, rhs)))
+                }
+                EvalBinop::Eq => todo!(),
+                EvalBinop::Neq => todo!(),
+                EvalBinop::Gt => Boolean(lhs > rhs),
+                EvalBinop::Gteq => Boolean(lhs >= rhs),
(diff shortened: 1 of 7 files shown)


partiql-eval/src/eval.rs
```diff
@@ -105,35 +105,56 @@ pub enum EvalBinop {
...
+                EvalBinop::Gt => Boolean(lhs > rhs),
+                EvalBinop::Gteq => Boolean(lhs >= rhs),
+                EvalBinop::Lt => Boolean(lhs < rhs),
+                EvalBinop::Lteq => Boolean(lhs <= rhs),
+                EvalBinop::Add => lhs + rhs,
+                EvalBinop::Sub => lhs - rhs,
+                EvalBinop::Mul => lhs * rhs,
+                EvalBinop::Div => lhs / rhs,
+                EvalBinop::Mod => lhs % rhs,
+                EvalBinop::Exp => todo!("Exponentiation"),
```
alancai98, line 156: Only doing +, -, *, /, % in this PR. Exponentiation is saved for a future PR.

partiql-eval/src/lib.rs
```diff
@@ -95,6 +114,124 @@ mod tests {
...
+                                vec![PathComponent::Key("a".to_string())],
+                            )),
+                            Box::new(ValueExpr::Lit(Box::new(lit))),
+                        ),
+                    )]),
+                    out: Box::new(BindingsExpr::Output),
+                })),
+            })
+        }
+        // Plan for `select a + <lit> as b from data`
```
alancai98, line 144: This test doesn't actually do any assertions. Currently just printing the evaluated output as a sanity check. I felt like the tests added to `partiql-value/src/lib.rs` were sufficient for this PR.

I can add assertions in an upcoming commit if that's desired.

partiql-value/src/lib.rs
```diff
@@ -23,13 +26,158 @@ pub enum Value {
...
     Blob(Box<Vec<u8>>),
     List(Box<List>),
     Bag(Box<Bag>),
     Tuple(Box<Tuple>),
 }
 
+impl ops::Add for Value {
+    type Output = Self;
+
+    fn add(self, rhs: Self) -> Self::Output {
```
alancai98, line 40: For arithmetic ops, if there are different operand types, coerces the operand in the following way int -> real (if no decimal among operands) -> decimal (if there's a decimal among operands).

partiql-value/src/lib.rs
```diff
@@ -151,12 +300,65 @@ impl Ord for Value {
             (Value::Boolean(_), _) => Ordering::Less,
             (_, Value::Boolean(_)) => Ordering::Greater,
 
+            // TODO: `OrderedFloat`'s implementation of `Ord` slightly differs from what we want in
```
jpschorr, line 303: Since there are a lot of cases here, it likely warrants a unit test, not just relying on the conformance tests that will eventually assert on it.
It's likely as 'simple' as creating a `Vec` of `Value` and then `.sort()`ing it and asserting on the output order.

partiql-eval/src/lib.rs
jpschorr: I think this is ok. Eventually, we'll remove this test that just prints as the evaluator matures. Then various ops will be unit tested via `Value` and integration tested with conformance/etc tests.

jpschorr (approved):

partiql-value/src/lib.rs
```diff
@@ -23,13 +26,158 @@ pub enum Value {
...
+            (_, Value::Null) => Value::Null,
+            (Value::Integer(l), Value::Integer(r)) => Value::Integer(l + r),
+            (Value::Real(l), Value::Real(r)) => Value::Real(*l + *r),
+            (Value::Decimal(l), Value::Decimal(r)) => Value::Decimal(l + r),
+            (Value::Integer(_), Value::Real(_)) => coerce_int_to_real(&self) + rhs,
+            (Value::Integer(_), Value::Decimal(_)) => coerce_int_or_real_to_decimal(&self) + rhs,
+            (Value::Real(_), Value::Decimal(_)) => coerce_int_or_real_to_decimal(&self) + rhs,
+            (Value::Real(_), Value::Integer(_)) => self + coerce_int_to_real(&rhs),
+            (Value::Decimal(_), Value::Integer(_)) => self + coerce_int_or_real_to_decimal(&rhs),
+            (Value::Decimal(_), Value::Real(_)) => self + coerce_int_or_real_to_decimal(&rhs),
```
am357, line 55: Not a blocker, perhaps for future: we could have a method for the BinOps to coerce the `lhs` in case of type mismatch before op overload; in that case we would end up w/ coercion only once.

am357 (approved):

jpschorr (approved):

PR #207, opened 2022-11-02
Remove `petgraph` from logical plan dependency

### LINKED ISSUE #206: Remove Petgraph dependency from `LogicalGraph`
As a follow on action item from #202  this commit removes the `petgraph`
dependency from `LogicalPlan`.

### DoD
- The dependency is removed from `partiql-logical` and the current tests pass.

am357:
*Issue #, if available:* 

Closes #206

*Description of changes:*
    
As a follow on action item from #202  this commit removes the `petgraph`
dependency from `LogicalPlan`.
    
The graph presented in this commit has just enough interfaces for doing
he work for now. We need to iterate on it later when vending as an API
to consumers.

### DIFF
--- a/partiql-logical/src/lib.rs
+++ b/partiql-logical/src/lib.rs
@@ -6,9 +6,57 @@ use std::collections::HashMap;
 #[derive(Debug)]
 pub struct LogicalPlan(pub StableGraph<BindingsExpr, (), Directed>);
 
+#[derive(Debug)]
+pub struct OpId(usize);
+
+impl OpId {
+    fn index(&self) -> usize {
+        self.0
+    }
+}
+
 impl LogicalPlan {
     pub fn new() -> Self {
         LogicalPlan(StableGraph::<BindingsExpr, (), Directed>::new())
+    }
+}
+
+#[derive(Debug)]
+pub struct LgPlan<'a, T> {
+    nodes: Vec<T>,
+    edges: Vec<Vec<&'a OpId>>,
+    node_count: usize,
+}
+
+impl<'b, T> LgPlan<'b, T> {
+    fn new() -> Self {
+        LgPlan {
+            nodes: vec![],
+            edges: vec![],
+            node_count: 0,
+        }
+    }
+
+    fn add_operator(&mut self, op: T) -> OpId {
+        self.nodes.push(op);
+        self.edges.push(vec![]);
+        self.node_count += 1;
+        OpId(self.node_count)
+    }
+    fn add_relation<'a: 'b>(&mut self, src: &'a OpId, dst: &'a OpId) {
+        let src_index = src.index() - 1;
+        let dst_index = dst.index() - 1;
+        self.nodes.get(src_index).expect("No such operator exists");
+        self.nodes.get(dst_index).expect("No such operator exists");
+
+        // let mut edge:&mut Vec<&OpId> = self.edges.get(src_index).expect("some").borrow_mut();
+        self.edges[src_index].extend(vec![dst]);
+    }
+    fn operators(&self) -> &Vec<T> {
+        &self.nodes
+    }
+    fn relations(&self) -> &Vec<Vec<&OpId>> {
+        &self.edges
     }
 }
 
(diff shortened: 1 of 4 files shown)


partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,58 @@
...
+#[derive(Debug, Hash)]
+pub struct OpId(usize);
+
+impl OpId {
+    pub fn index(&self) -> usize {
+        self.0
+    }
+}
+
+impl PartialEq<Self> for OpId {
```
jpschorr, line 13: can probably use `#[derive(PartialEq, Eq)]`

partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,58 @@
 use partiql_value::{BindingsName, Value};
-use petgraph::prelude::StableGraph;
-use petgraph::Directed;
 use std::collections::HashMap;
 
+#[derive(Debug, Hash)]
```
jpschorr, line 4: for wrapping a `usize`, you could also `#derive[(Copy)]` here, which would make dealing with these IDs more pleasant

partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,58 @@
...
+}
+
+impl Eq for OpId {}
+
 #[derive(Debug)]
-pub struct LogicalPlan(pub StableGraph<BindingsExpr, (), Directed>);
+pub struct LogicalPlan<'a, T> {
+    nodes: Vec<T>,
+    edges: Vec<(&'a OpId, &'a OpId)>,
+    node_count: usize,
```
jpschorr, line 25: is this not redundant with `nodes.len()` ?

partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,58 @@
...
+    }
+}
+
+impl Eq for OpId {}
+
 #[derive(Debug)]
-pub struct LogicalPlan(pub StableGraph<BindingsExpr, (), Directed>);
+pub struct LogicalPlan<'a, T> {
+    nodes: Vec<T>,
+    edges: Vec<(&'a OpId, &'a OpId)>,
```
jpschorr, line 24: I don't think it's worth the hassle of references for the ids here. Might as well use `Vec<(OpId, OpId)>` and ensure the invariant that the ids must reference existing nodes in the `add_flow` fn.

partiql-logical/src/lib.rs
jpschorr: e.g. 


```rust

pub fn node_count(&self) -> usize {self.nodes.len()}

```

partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,58 @@
...
+    pub fn add_operator(&mut self, op: T) -> OpId {
+        self.nodes.push(op);
+        self.node_count += 1;
+        OpId(self.node_count)
+    }
+
+    pub fn add_flow<'a: 'b>(&mut self, src: &'a OpId, dst: &'a OpId) {
+        let src_index = src.index() - 1;
+        let dst_index = dst.index() - 1;
+        self.nodes.get(src_index).expect("No such operator exists");
```
jpschorr, line 46: this is essentially 
```rust

assert!(src <= self.node_count())

```

am357: True, will change.

partiql-logical/src/lib.rs
am357: It is actually redundant, will change.

partiql-logical/src/lib.rs
am357: Yeah I think what confused me was a compiler error that I got during dev; let me take another stab at it.

partiql-logical/src/lib.rs
am357: Makes sense; changed in the new commit.

partiql-logical/src/lib.rs
```diff
@@ -1,14 +1,53 @@
...
+
+    pub fn add_operator(&mut self, op: T) -> OpId {
+        self.nodes.push(op);
+        OpId(self.operator_count())
+    }
+
+    pub fn add_flow(&mut self, src: OpId, dst: OpId) {
+        let src_idx = src.index() - 1;
+        let dst_idx = dst.index() - 1;
+        assert!(src_idx <= self.operator_count());
```
jpschorr, line 35: I believe
- created `OpId`s are 1-indexed (they are created after the new node is `push`ed). 
- `operator_count()` is also '1-indexed' 

so either the `-1` is unnecessary or we should replace `<=` with `<`

jpschorr (approved):

partiql-logical/src/lib.rs
am357: Agreed, I overlooked this as I was rushing to push the commit; will change in the PR that I'm putting together for moving the rest of the operators to DAG plan.

PR #209, opened 2022-11-04
Move the plan (logical and eval) and evaluation to DAG model

am357:
*Issue #, if available:*
#211 

*Description of changes:*

- Moves the remaining operators to DAG operators for both `logical` and `evaluation`. 
- Adds a `extend_with_flows` method to `LogicalPlan` to accept array of `flows` (edges).
- Addresses the remaining comments from #202  #207 
  - Moves `add_flow` to `1-indexed`. 
  - Removes optional value from `EvalResult`.
- Renames operators names (E.g. prepends `Eval` to the new ones (E.g. `Scan` to `EvalScan`).
- Removes redundant structures related to previous the tree model.
- Refactors the current tests to ensure they pass w/ the latest changes.


partiql-logical/Cargo.toml
```diff
@@ -21,6 +21,7 @@ edition.workspace = true
 
 [dependencies]
 partiql-value = { path = "../partiql-value" }
+petgraph = "0.6.*"
```
jpschorr, line 24: Doesn't seem to be needed for this crate?

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
+        let ordered = &v.is_ordered();
+        let mut c: i64 = 0;
+        let at_key = &self.at_key;
         for t in v.into_iter() {
-            let out = Tuple(HashMap::from([(self.as_key.clone(), t)]));
-            value.push(Value::Tuple(Box::new(out)));
+            let mut out = HashMap::from([(self.as_key.clone(), t)]);
+            if !at_key.is_empty() {
+                let at_id = if *ordered { c.into() } else { Missing };
+                out.insert(at_key.clone(), at_id.into());
```
alancai98, line 82: nit: (clippy warning) `at_id` is already a `Value` so don't need the `.into()` call

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
 }
 
-impl DagEvaluable for Scan {
+impl Evaluable for EvalScan {
     fn evaluate(&mut self, ctx: &dyn EvalContext) -> Option<Value> {
-        let mut value = partiql_bag!();
+        let mut value = partiql_bag![];
         let v = self.expr.evaluate(&Tuple(HashMap::new()), ctx);
+        let ordered = &v.is_ordered();
+        let mut c: i64 = 0;
```
alancai98, line 76: What does `c` represent? Is it the counter for the `at` index? If so, could have a more descriptive variable name.

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
     }
 }
 
 #[derive(Debug)]
-pub struct Project {
+pub struct EvalUnpivot {
+    pub expr: Box<dyn EvalExpr>,
+    pub as_key: String,
+    pub at_key: String,
+    pub input: Option<Value>,
```
alancai98, line 103: Does `EvalUnpivot` need an `input` field? `EvalScan` doesn't have an `input` node. And the current `impl` for `Evaluable` doesn't make use of it other than in `update_input`.

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
+            input: None,
+            output: None,
+        }
+    }
+
+    #[inline]
+    fn eval_filter(&self, bindings: &Tuple, ctx: &dyn EvalContext) -> bool {
+        let result = self.expr.evaluate(bindings, ctx);
+        match result {
+            Boolean(bool_val) => bool_val,
```
alancai98, line 166: Could we also have a `NULL` or `MISSING` value? E.g. `WHERE` expression evaluates to `NULL` or `MISSING`

partiql-logical/src/lib.rs
```diff
@@ -135,32 +140,25 @@ pub enum BindingsToValueExpr {}
...
-pub struct From {
+pub struct Scan {
     pub expr: ValueExpr,
     pub as_key: String,
-    pub at_key: Option<String>,
-    pub out: Box<BindingsExpr>,
+    pub at_key: String,
 }
 
+/// [`Scan`] bridges from [`ValueExpr`]s to [`BindingExpr`]s
```
alancai98, line 151:
```suggestion
/// [`Unpivot`] bridges from [`ValueExpr`]s to [`BindingExpr`]s
```

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
+    fn evaluate(&mut self, ctx: &dyn EvalContext) -> Option<Value>;
+    fn update_input(&mut self, input: &Value);
+}
+
 #[derive(Debug)]
-pub struct Scan {
+pub struct EvalScan {
     pub expr: Box<dyn EvalExpr>,
     pub as_key: String,
+    pub at_key: String,
```
alancai98, line 56: Any reason this is modeled as a `String` rather than an `Option<String>`? I think modeling it as an optional type is less confusing to represent whether an `at_key` is provided or not (`None` rather than empty string for indicating empty string is used.

partiql-eval/src/eval.rs
```diff
@@ -46,56 +44,172 @@ pub enum EvaluationError {
...
+            value.push(Value::Tuple(Box::new(Tuple(out))));
         }
-        Some(Value::Bag(Box::new(value)))
+
+        self.output = Some(Value::Bag(Box::new(value)));
+        self.output.clone()
     }
+
     fn update_input(&mut self, _input: &Value) {
         todo!("update_input for Scan")
```
alancai98, line 94: Just curious, in what scenario would `update_input` be called on an `EvalScan` node?

partiql-eval/src/eval.rs
am357: TBH I don't know yet; intuitively I'd say they should, in favor of `expr` field (esp. for sub-queries), but can remove for now until it's required.

partiql-eval/src/eval.rs
am357: Yes it is a counter for `at_index`, will make the change.

partiql-eval/src/eval.rs
am357: I've actually tried it with `Option` as well but then found the API as being harder to work with; this is based on the assumption that in evaluation, we're in lower levels and initialization is being done via some other procedure, therefore IMO it's more ergonomic to use plain `String` for the code that wants to generate the plan. On the other hand, I think it makes sense to have it as optional in `LogicalPlan` the same as `Unpivot`. I don't have a strong opinion about this at this stage btw, let me know if you want to discuss about this further.

partiql-eval/src/eval.rs
am357: Good catch, I think spec. specifically discusses about this case; let me look into it.

partiql-eval/src/eval.rs
am357: Referring to what I've mentioned in the other comment, it could be the case of last operator from a sub-query to update the `Scan`'s input, but I think we still need to flesh this out.

partiql-eval/src/eval.rs
alancai98: Ah, got it. I guess to me, `Option` is "lower level" or less ambiguous than using `String`. I don't have a strong opinion on it either. We can always iterate as we learn more and adjust as needed.

partiql-eval/src/eval.rs
alancai98: Ok, so like in the example you gave, a subquery within a `FROM` source?

alancai98 (approved):
Migrated plan to the DAG model looks good to me. CI seems to be failing due to `rustfmt` failures.

partiql-eval/src/eval.rs
jpschorr: I would prefer `Option<String>` for `at_key`, since it is optional.

It should be easy enough to handle the optional:

In `EvalScan`, replace

```rust

if !at_key.is_empty() {
...
}

```

with

```rust

if let Some(at_key) = &self.at_key {
...
}

```


And for creation all the `"".to_string()`s become `None` and any `"at_key_name".to_string()` is `Some("at_key_name".into())`

jpschorr (approved):

partiql-logical/Cargo.toml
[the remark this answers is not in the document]
am357: Whoops! has leaked in from the previous impl, will remove.

partiql-eval/src/eval.rs
am357: Yes.

partiql-eval/src/eval.rs
am357: Fair, I hear you both, will change in the upcoming commit.

partiql-eval/benches/bench_eval.rs
```diff
@@ -61,7 +61,7 @@ fn eval_bench(c: &mut Criterion) {
                 expr: Box::new(EvalVarRef {
```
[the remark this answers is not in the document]
am357: Ack—fixed in the new commit.

jpschorr (approved):

PR #214, opened 2022-11-10
Add Conformance Test Generation of 'Envs' and 'Equivs'

jpschorr:
Adds:
- generation of files containing the ion-formatted test environment data
- generation of tests for equivalence classes

It's probably easiest to review by pulling the changes locally and inspecting the generated test output.

---

Here's a sample of the generation:

input spec (abridged):

```ion

'section-5'::[
    envs::{
        someOrderedTable: [
            {a: 0, b: 0},
            {a: 1, b: 1}
        ],
        someUnorderedTable: $bag::[
            {a: 0, b: 0},
            {a: 1, b: 1}
        ]
    },
    {
        name: "single source FROM with list and AT clause",
        statement: "SELECT * FROM someOrderedTable AS x AT y",
        assert: {
            evalMode: [EvalModeCoerce, EvalModeError],
            result: EvaluationSuccess,
            output: $bag::[{a: 0, b: 0, y: 0}, {a: 1, b: 1, y: 1}]
        }
    },

   ...

]

```

generated tests (abridged):

```rust

#[allow(unused_imports)]
mod r#section_5 {
    use super::*;

    const ENV_ION_TEXT: &'static str = include_str!("spec_tests___section_5.env.ion");
    fn environment() -> Option<TestValue<'static>> {
        Some(ENV_ION_TEXT.into())
    }

    /// Generated test for test named `single source FROM with list and AT clause`
    #[test]
    #[allow(text_direction_codepoint_in_literal)]
    #[ignore = "not yet implemented"]
    fn r#single_source_from_with_list_and_at_clause() {
        let env = environment();

        //**** evaluation success test case(s) ****//
        let expected = "$bag::[
            {
                a: 0,
                b: 0,
                y: 0
            },
            {
                a: 1,
                b: 1,
                y: 1
            }
        ]"
        .into();
        let stmt = "SELECT * FROM someOrderedTable AS x AT y";
        pass_eval(stmt, EvaluationMode::Coerce, &env, &expected);
        pass_eval(stmt, EvaluationMode::Error, &env, &expected);
    }

   ...

}


```

---


partiql-conformance-test-generator/src/reader.rs
```diff
@@ -182,7 +182,24 @@ impl TryFrom<&Struct> for TestCase {
...
+            ),
+            _ => {
+                return Err(miette!(
+                    "unexpected type {:?} when parsing TestCase statement",
+                    statement.ion_type()
+                ))
+            }
+        };
+
+        //let statement = expect_str!(value.get("statement"), "TestCase statement").into();
```
alancai98, line 202: Unused line

partiql-conformance-test-generator/src/generator.rs
```diff
@@ -112,19 +250,53 @@ impl Generator {
...
+        let env_file = self
+            .curr_mod_path
+            .iter()
+            .map(|s| s.escape_path())
+            .collect::<Vec<_>>()
+            .join("___")
+            + ".env.ion";
+
+        scope.raw(
+            quote! {
```
alancai98, line 263: Just curious, was using `quote` easier to perform certain code generation than the `codegen` crate?

alancai98 (approved):

partiql-conformance-test-generator/src/generator.rs
jpschorr: Yes. I originally started using it when trying to create a [OnceCell](https://docs.rs/once_cell/latest/once_cell/) to hold the 'envs' as a lazily implemented global static variable, as it was much clearer to do with `quote`.

It looked something like:

```rust

let tokens = quote! {
    const ENV_ION_TEXT: &'static str = #ion_text_str;
    static ENV_STRUCT: once_cell::unsync::Lazy<Mutex<Struct>> = Lazy::new(|| {
        let element_stream = element_reader().read_all("foo".as_bytes()).expect("ion value");
        let element = element_stream[0];
        element.as_struct().expect("struct")
    });
};

```

But, the currently implementation of ion-rust's `Element` uses an `Rc`, which makes it neither `Send` nor `Sync`, so I was unable to create the static variable.


Another way to think about it is: `codegen` makes it convenient to create and pass around structure (e.g., `scope`s, `module`s, `fn`s), but it doesn't provide any particular functionality for lines of logic; `quote` is nice for this.

am357 (approved):
Had a cursory review as I don't have the prior context; however, as suggested by Josh looked at a generated test (as a sample) and it LGTM.

PR #217, opened 2022-11-11
For Conformance Test Generator: large epectations to file rather than inline string.

jpschorr:
Move large "expected" values to external files similar to "environment" files.
Move all such test data files into `_test_data` directory.

---

### DIFF
--- a/partiql-conformance-test-generator/src/generator.rs
+++ b/partiql-conformance-test-generator/src/generator.rs
@@ -47,7 +47,7 @@ pub enum TestTree {
 #[derive(Debug)]
 pub enum Node {
     Test(TestNode),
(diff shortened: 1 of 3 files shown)


partiql-conformance-test-generator/src/generator.rs
```diff
@@ -251,6 +253,29 @@ impl Generator {
     }
 
     fn gen_envs(&mut self, scope: &mut Scope, envs: &Environments) {
+        let envs = struct_to_string(&envs.envs);
+
+        if envs.lines().count() < 10 {
```
am357, line 258: This and the other constant `25` could go to const.

am357 (approved):

PR #218, opened 2022-11-15
Add `INNER` and `CROSS JOIN`s to planner and evaluator

alancai98:
Adds `INNER` and `CROSS JOIN`s to planner and evaluator, which required adding a branch side tracking to the DAG.

Still have some TODOs to address but wanted to cut this PR before working on those.
- `LATERAL` `JOIN`s
- Cleanup of `.clone()`ing
- `LEFT JOIN`s

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 
 use petgraph::algo::toposort;
 use petgraph::prelude::StableGraph;
-use petgraph::{Directed, Incoming, Outgoing};
+use petgraph::{Directed, Outgoing};
 
 use partiql_value::Value::{Boolean, Missing, Null};
 use partiql_value::{
@@ -18,7 +18,7 @@ use crate::env::basic::MapBindings;
 use crate::env::Bindings;
 
 #[derive(Debug)]
-pub struct EvalPlan(pub StableGraph<Box<dyn Evaluable>, (), Directed>);
+pub struct EvalPlan(pub StableGraph<Box<dyn Evaluable>, u8, Directed>);
 
 impl Default for EvalPlan {
     fn default() -> Self {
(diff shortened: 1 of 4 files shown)


partiql-eval/src/eval.rs
```diff
@@ -108,11 +108,103 @@ impl Evaluable for EvalScan {
...
+            on,
+            input_l: None,
+            input_r: None,
+            output: None,
+        }
+    }
+}
+
+impl Evaluable for EvalJoin {
+    fn evaluate(&mut self, ctx: &dyn EvalContext) -> Option<Value> {
```
alancai98, line 147: Currently, the `JOIN` implementation has quite a few `.clone()`s (partially from my own Rust inexperience). I'll work to remove unnecessary clones when we can agree on this modeling of `JOIN`s.

partiql-eval/src/eval.rs
```diff
@@ -627,39 +719,33 @@ impl Evaluator {
         // that all v ∈ V \{v0} are reachable from v0. Note that this is the definition of trees
         // without the condition |E| = |V | − 1. Hence, all trees are DAGs.
         // Reference: https://link.springer.com/article/10.1007/s00450-009-0061-0
-        match graph.externals(Incoming).exactly_one() {
-            Ok(_) => {
```
alancai98, line 631: This graph invariant no longer applies (e.g. could have two scan nodes without incoming nodes)

am357 (approved):

partiql-eval/src/eval.rs
```diff
@@ -50,7 +50,7 @@ pub enum EvaluationError {
 
 pub trait Evaluable: Debug {
     fn evaluate(&mut self, ctx: &dyn EvalContext) -> Option<Value>;
-    fn update_input(&mut self, input: &Value);
+    fn update_input(&mut self, input: &Value, branch_num: u8);
```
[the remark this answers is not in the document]
alancai98: Agreed that this approach can be more resilient to future changes. I'll leave it as is for now. If we modify the interface further (e.g. adding more arguments or modifying the arguments), can use this approach.

jpschorr (approved):

PR #221, opened 2022-11-18
Adds non-reserved keywords

johnedquinn:
## Relevant Issues
- None

## Description
- This is a proof of concept (POC) for adding non-reserved keywords to the Rust implementation
- I've added several keywords to the NonReservedKeywords list in LALRPOP
  - `ACYCLIC`, `BOTH`, `DOMAIN`, `LEADING`, `TRAIL`, `TRAILING`, and `USER`
- I also modified the preprocessor so that it can parse the non-reserved keywords appropriately.
- If they can be parsed as VarRefExprs, we convert them to SymbolPrimitives (case-insensitive), so we don't even need to know what their original case is. I didn't add qualified/unqualified to this PR, as it's a quick POC.
- I've added tests to show several scenarios of using non-reserved keywords.
- I don't intend for this to be merged, as I am not familiar with the codebase -- however I wanted to show how it can possiblly be done.
- Note: I'm unfamiliar with this language, so please excuse mistakes made along the way


partiql-parser/src/parse/mod.rs
```diff
@@ -679,6 +679,55 @@ mod tests {
         }
     }
 
+    // PROOF OF CONCEPT: NON-RESERVED KEYWORDS
+    mod non_reserved {
```
jpschorr, line 683: These tests all pass without *any* of the above parser changes, except for `with_order` (thought I think the gpml tests would fail when the gpml parsing rules are included).

I'm also not sure that i feel treating `order` as an unreserved keyword is a good idea...

PR #222, opened 2022-11-18
Fix comparison ops, add between op

alancai98:
This PR:
- Fixes comparison ops (`<`, `>`, `<=`, `>=`). Previously didn't properly handle data type mismatches (returned boolean rather than missing)
- Adds `BETWEEN` op (e.g. `foo BETWEEN 1 AND 2`)
- Changes some PartiQL `Value` extension methods to operate on reference to `Value`s rather than `Value`s
- Some minor bug fixes from previous comparison implementation

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -556,8 +556,8 @@ impl EvalExpr for EvalBinOpExpr {
 
         let rhs = self.rhs.evaluate(bindings, ctx);
         match self.op {
-            EvalBinOp::And => lhs.and(rhs),
-            EvalBinOp::Or => lhs.or(rhs),
+            EvalBinOp::And => lhs.and(&rhs),
+            EvalBinOp::Or => lhs.or(&rhs),
             EvalBinOp::Concat => {
                 // TODO non-naive concat. Also doesn't properly propagate MISSING and NULL
                 let lhs = if let Value::String(s) = lhs {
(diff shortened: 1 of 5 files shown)


partiql-eval/src/eval.rs
```diff
@@ -588,6 +588,22 @@ impl EvalExpr for EvalBinOpExpr {
...
+    pub from: Box<dyn EvalExpr>,
+    pub to: Box<dyn EvalExpr>,
+}
+
+impl EvalExpr for EvalBetweenExpr {
+    fn evaluate(&self, bindings: &Tuple, ctx: &dyn EvalContext) -> Value {
+        let value = self.value.evaluate(bindings, ctx);
+        let from = self.from.evaluate(bindings, ctx);
+        let to = self.to.evaluate(bindings, ctx);
+        value.gteq(&from).and(&value.lteq(&to))
```
alancai98, line 603: Previously, these `Value` methods (`gteq` and `lteq` and `and`) used `Value`s as parameters, which took ownership of the `Value`s. Changing the methods to use references allowed me to reuse the `Value`s without cloning. Not sure if this was the best way to reuse the `Value`s.

partiql-eval/src/plan.rs
```diff
@@ -120,7 +120,7 @@ impl EvaluatorPlanner {
                     BinaryOp::Gt => EvalBinOp::Gt,
                     BinaryOp::Gteq => EvalBinOp::Gteq,
                     BinaryOp::Lt => EvalBinOp::Lt,
-                    BinaryOp::Lteq => EvalBinOp::Gteq,
+                    BinaryOp::Lteq => EvalBinOp::Lteq,
```
alancai98, line 123: Previously committed bug

partiql-value/src/lib.rs
```diff
@@ -512,6 +573,7 @@ impl Ord for Value {
                 }
             }
             (Value::Integer(l), Value::Integer(r)) => l.cmp(r),
+            (Value::Decimal(l), Value::Decimal(r)) => l.cmp(r),
```
alancai98, line 576: Line was accidentally omitted from this `match`

jpschorr (approved):

am357 (approved):

PR #220, opened 2022-11-16
Implement `IN` expr

am357:
*Issue #, if available:*

Closes #219

*Description of changes:*
This commit adds the required changes to implement `IN` expression. The following test passes:

```SQL
SELECT a AS b FROM [{'a': 1}, {'a': 2}, {'a': 3}] WHERE a IN [1];
-- Expected << {'b': 1} >>
```

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #227, opened 2022-11-23
Implement subquery eval using sub-plan approach

am357:
*Description of changes:*
This PR includes changes to enable subquery evaluation. This is by no means a final solution and the intention is to put a model as reference so that we can discuss and iterate on it.

The change also includes an implementation of a LATERAL JOIN. As discussed with Josh and Alan, putting the changes here w/ a TODO to revisit our current JOIN implementation to cater for LATERAL functionality.

The following tests are currently passing:
- `SELECT t.a, s FROM data AS t, (SELECT v.a*2 AS u FROM t AS v) AS s;`
- `SELECT t.a, (SELECT v.a*2 AS u FROM t AS v) AS s FROM data AS t;`

**_Note: The diff is based on [feat-value-iters](https://github.com/partiql/partiql-lang-rust/tree/feat-value-iters) which hasn't been merged, therefore it has some unrelated changes related to `IN` expr and tests; once the other branch gets merged I'll update this PR._**


partiql-eval/src/eval.rs
```diff
@@ -614,6 +720,9 @@ pub enum EvalBinOp {
     Div,
     Mod,
     Exp,
+
+    // Boolean ops
+    In,
```
alancai98, line 725: fyi, the `feat-value-iter` doesn't have the `IN` operator implementation (https://github.com/partiql/partiql-lang-rust/pull/220), so that PR's diff is included in this PR

partiql-eval/src/eval.rs
```diff
@@ -164,6 +227,21 @@ impl Evaluable for EvalJoin {
             })
         }
 
+        #[inline]
+        fn cross_lateral<'a>(
+            left: impl Iterator<Item = &'a Value> + Clone + 'a,
+            right: impl Iterator<Item = &'a Value> + Clone + 'a,
+        ) -> impl Iterator<Item = Tuple> + 'a {
+            left.zip(right).map(|(l_tuple, r_tuple)| {
```
alancai98, line 235: Interesting that we can `.zip` the two iterators together to perform the lateral cross join rather than `.cartesian_product` for the non-lateral cross join. Guess it makes sense based on my current understanding. I'll be looking more into lateral joins this week to see if this implementation would need altered.

alancai98 (approved):
Looks good overall. Had some questions about the refactoring of the `evaluate` function. Also left a comment about the lateral join implementation, which may change as I look more into implementing lateral joins this week.

partiql-eval/src/eval.rs
alancai98: Looking at this now with some other test cases, I don't think `.zip` is the right behavior here. It might have worked for the `subquery_in_from` example you had in `lib.rs` since the `lhs` and `rhs` have the same length of elements (i.e. they both use `data`). But if the `lhs` and `rhs` have different lengths, then some rows may be ignored.

Consider a test like the following taken from [spec example 10](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.5.3):
```Rust
#[test]
fn select_with_cross_join_sensors() {
    let mut lg = LogicalPlan::new();

    let from_lhs = lg.add_operator(scan("sensors", "s"));
    let from_rhs = lg.add_operator(BindingsExpr::Scan(logical::Scan {
        expr: path_var("s", "readings"),
        as_key: "r".to_string(),
        at_key: None,
    }));

    let project = lg.add_operator(Project(logical::Project {
        exprs: HashMap::from([
            ("v".to_string(), path_var("r", "v")),
        ]),
    }));

    let join = lg.add_operator(BindingsExpr::Join(logical::Join {
        kind: JoinKind::CrossLateral,
        on: None,
    }));

    let sink = lg.add_operator(BindingsExpr::Sink);
    lg.add_flow_with_branch_num(from_lhs, from_rhs, 0);
    lg.add_flow_with_branch_num(from_lhs, join, 0);
    lg.add_flow_with_branch_num(from_rhs, join, 1);
    lg.add_flow_with_branch_num(join, project, 0);
    lg.add_flow_with_branch_num(project, sink, 0);

    let out = evaluate(lg, join_data_sensors());
    println!("{:?}", &out);
}
```
Using `join_data_sensors()` specified as follows:
```Rust
fn join_data_sensors() -> MapBindings<Value> {
    let sensors = partiql_list![
        partiql_tuple![
            ("readings", partiql_list![
                partiql_tuple![("v", 1.3)],
                partiql_tuple![("v", 2)],
            ])
        ],
        partiql_tuple![
            ("readings", partiql_list![
                partiql_tuple![("v", 0.7)],
                partiql_tuple![("v", 0.8)],
                partiql_tuple![("v", 0.9)],
            ])
        ],
    ];
    let mut bindings = MapBindings::default();
    bindings.insert("sensors", sensors.into());
    bindings
}
```

The above outputs:
```
<<[Tuple { v: 1.3 }, Tuple { v: 2 }]>>
```

Rather than the `v` for every `reading` (i.e.

```
<<[Tuple { v: 1.3 }, Tuple { v: 2 }, Tuple { v: 0.7 }, Tuple { v: 0.8 }, Tuple { v: 0.9 }]>>
```

partiql-eval/src/eval.rs
```diff
@@ -31,6 +32,53 @@ impl EvalPlan {
     fn new() -> Self {
         EvalPlan(StableGraph::<Box<dyn Evaluable>, u8, Directed>::new())
     }
+
+    pub fn execute_mut(&mut self, bindings: MapBindings<Value>) -> Result<Evaluated, EvalErr> {
```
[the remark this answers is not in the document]
am357: > Any reason the Evaluator struct was deleted and for its execute function to be moved under EvalPlan?

Sorry I had to provide the reason as part of PR.

The main issue is that b/c we're calling `Evaluator`'s `execute` from within the node (which is specific to subquery) we need to move `self.plan` to `execute` but we can't b/c we're using `&self` as part of `evaluate`. Here we have an option of cloning but as far as my current understanding with Rust goes we can't  clone `dyn Evaluable` unless we use libraries like [dyn-clone](https://crates.io/crates/dyn-clone) b/c it's not sized (happy to discuss the `dyn-clone` option further). Here we end up with using references. For this, there are options like using mutable reference as `execute`'s parameter but those will also end up mutating the plan through references (shared or exclusive). In that case, I found the current approach making sense i.e. mutating the plan in place through its own method. I might be too tunneled visioned in this so please let me know if you have an idea for another approach.

BTW: I haven't actually been able to successfully implement the other approach using references in `Evaluator` yet.

> Does self need to be a mutable reference?

Yes, this is b/c we modify the `self.plan` through execution to update the nodes.

partiql-eval/src/eval.rs
am357: Nice clarification! honestly I had my doubts about the current `CrossLateral` that's why I mentioned in the PR description that it needs a revisit. The main objective of this PR is to implement subqueries but not joins. I know as a follow-on we're already looking into `LATERAL JOIN` implementation in general.

PR #228, opened 2022-11-30
Add simple and searched case when expressions to evaluator

alancai98:
Note: target branch is `feat-value-iters`. Will create a new PR targeting `main` once that branch's PR is merged to `main`.

*Issue #, if available:* None.

*Description of changes:* Adds simple and searched case when expressions to the evaluator. Also fixes projection behavior when a tuple value is `MISSING` (previously included those tuple values; now omits).

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -5,6 +5,7 @@ use std::fmt::Debug;
 use thiserror::Error;
 
 use petgraph::algo::toposort;
+use petgraph::data::DataMapMut;
 use petgraph::prelude::StableGraph;
 use petgraph::{Directed, Outgoing};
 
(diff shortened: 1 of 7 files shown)


partiql-eval/src/eval.rs
```diff
@@ -326,7 +326,11 @@ impl Evaluable for EvalProject {
             let mut t = Tuple::new();
 
             self.exprs.iter().for_each(|(alias, expr)| {
-                t.insert(alias.as_str(), expr.evaluate(&v_as_tuple, ctx));
+                let evaluated_val = expr.evaluate(&v_as_tuple, ctx);
+                if evaluated_val != Missing {
+                    // Per section 2 of PartiQL spec: "value MISSING may not appear as an attribute value
+                    t.insert(alias.as_str(), evaluated_val);
+                }
```
alancai98, line 333: Previous behavior kept projected tuple values that evaluated to `MISSING`. Changed to now omit those attribute pairs.

Perhaps a better modeling, would be as the spec mentions in [section 6.3](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.6.3) -- that is modeling is a `SELECT VALUE`. This modeling would likely involve changing the logical to evaluation plan in `plan.rs`.

partiql-eval/src/plan.rs
```diff
@@ -187,6 +187,40 @@ impl EvaluatorPlanner {
...
+                            )),
+                            self.plan_values(*case.1),
+                        )
+                    })
+                    .collect();
+                let default = e
+                    .default
+                    .as_ref()
+                    .map(|default| self.plan_values(*default.clone()));
+                // Here, rewrite `SimpleCaseExpr`s as `SearchedCaseExpr`s
```
alancai98, line 209: Modeling simple and searched case expressions separately in the logical plan but reducing simple case expressions to searched case expressions in the evaluator.

partiql-eval/src/plan.rs
```diff
@@ -187,6 +187,40 @@ impl EvaluatorPlanner {
...
+                let cases = e
+                    .cases
+                    .into_iter()
+                    .map(|case| (self.plan_values(*case.0), self.plan_values(*case.1)))
+                    .collect();
+                let default = e
+                    .default
+                    .as_ref()
+                    .map(|default| self.plan_values(*default.clone()));
+                Box::new(EvalSearchedCaseExpr { cases, default })
```
alancai98, line 222: Could also make the `default` a non-option type by following what's mentioned in the SQL-92 spec (section 6.9 <case expression> pg 141):

```
 6) If an <else clause> is not specified, then ELSE NULL is implicit.
```

alancai98: Updated in 7de9988dd95e286ba8e7d92198b675f7d2ea4168 to use implicit `ELSE NULL` when no `ELSE` clause is provided.

jpschorr (approved):

alancai98:
Pretty cool that GitHub changed the target branch (originally targeted `feat-value-iters`, which was merged to `main`), so a separate PR isn't needed.
image

PR #229, opened 2022-12-01
Implements `IS` (unknown types only), `NULLIF`, and `COALESCE`

alancai98:
Note: target branch is `case-expressions`. Will create a new PR targeting `main` once that branch's PR (#228) is merged to `main`.

*Issue #, if available:* None.

*Description of changes:* Implements `IS` (unknown types only -- i.e. `IS [NOT] NULL` and `IS [NOT] MISSING`), `NULLIF`, and `COALESCE`. `NULLIF` and `COALESCE` are implemented using the rewrites mentioned in the SQL-92 spec (section 6.9, pg 142).

### DIFF
--- a/partiql-logical/src/lib.rs
+++ b/partiql-logical/src/lib.rs
@@ -147,6 +147,17 @@ pub enum Type {
 }
 
 #[derive(Clone, Debug)]
+pub struct NullIfExpr {
+    pub lhs: Box<ValueExpr>,
+    pub rhs: Box<ValueExpr>,
+}
+
+#[derive(Clone, Debug)]
+pub struct CoalesceExpr {
+    pub elements: Vec<ValueExpr>,
+}
+
+#[derive(Clone, Debug)]
 #[allow(dead_code)] // TODO remove once out of PoC
 pub enum ValueExpr {
     // TODO other variants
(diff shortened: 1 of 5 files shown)


partiql-value/src/lib.rs
```diff
@@ -317,6 +317,8 @@ pub trait NullableOrd {
 impl NullableEq for Value {
     type Output = Self;
 
+    // TODO: `eq` and `neq` are not quite right as implemented. Should be able to assert equality between
+    //  different comparable types (e.g. numerics)
```
alancai98, line 321: I found that equality wasn't implemented exactly right for the edge case of comparing different comparable types (e.g. `1` = `1.0`). Marked as a TODO here.

partiql-logical/src/lib.rs
```diff
@@ -108,6 +108,55 @@ pub enum PathComponent {
...
 
+#[derive(Clone, Debug)]
+pub struct IsTypeExpr {
+    pub not: bool,
+    pub expr: Box<ValueExpr>,
+    pub is_type: Type,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Type {
```
alancai98, line 119: `Type` copied from `ast.rs`. Currently used in both the logical plan and physical plan. Perhaps a better place would be the `partiql-types` crate?

jpschorr (approved):

PR #230, opened 2022-12-02
Add AST visitor 

jpschorr:
Add AST visitor

- Add `Visit` and `Visitor` Traits for visiting AST
- Add AST node `Visit` impls via `proc_macro`s
- Refactor some AST nodes (including `FROM`,`WHERE`, and `HAVING` clauses) for better visitation

---


partiql-ast/src/ast.rs
```diff
@@ -64,33 +66,38 @@ pub enum DdlOp {
     DropIndex(DropIndex),
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Visit, Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct CreateTable {
+    #[visit(skip)]
```
alancai98, line 72: Out of curiosity, why are some `ast` nodes skipped?

partiql-ast/partiql-ast-macros/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    );
+    let exit_fn_name = Ident::new(
+        &format!("exit_{}", visit_fn_name),
+        proc_macro2::Span::call_site(),
+    );
+
+    let visit_children = (!should_skip_recurse(ast)).then(|| impl_visit_children(&ast));
+
+    let ast_name = &ast.ident;
+    quote! {
```
alancai98, line 64: I'm still new to procedure macros in Rust. Is there a way to see what the procedural macro generates in `cargo` or CLion?

alancai98 (approved):
Looks good. Left a nit and a couple non-blocking questions

partiql-ast/src/ast.rs
```diff
@@ -262,16 +427,30 @@ pub enum CollectionLit {
...
     pub qualifier: ScopeQualifier,
 }
 
-#[derive(Clone, Debug, PartialEq)]
+/// Indicates scope search order when resolving variables.
+/// Has no effect except within `FROM` sources.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ScopeQualifier {
+    /// Use the default search order.
```
am357, line 444: Perhaps adding more info (or spec. reference to default order would be good).

am357 (approved):

partiql-ast/partiql-ast-macros/src/lib.rs
jpschorr: `cargo expand` in the directory of the crate using the macro will expand *all* macro usages.

partiql-ast/src/ast.rs
jpschorr: Importantly, it's not really 'ast nodes' that are skipped, but rather data items that inhere to the current node.

In some sense, we recursively visit 'children'. Things that aren't 'children' get 'skip'ed by the visitor macros, as it wouldn't really make sense for your visitor to have a `visit_symbol_primitive` method that is divorced from what node that symbol is appearing in.

PR #232, opened 2022-12-07
`Value` arithmatic by reference rather than taking value ownership

jpschorr:
`Value` arithmatic by reference rather than taking value ownership
---

### DIFF
--- a/partiql-value/src/lib.rs
+++ b/partiql-value/src/lib.rs
@@ -37,8 +37,8 @@ pub enum Value {
     // TODO: add other supported PartiQL values -- timestamp, date, time, sexp
 }
 
-impl ops::Add for Value {
-    type Output = Self;
+impl ops::Add for &Value {
+    type Output = Value;
 
     fn add(self, rhs: Self) -> Self::Output {
         match (&self, &rhs) {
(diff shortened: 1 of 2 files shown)


partiql-value/src/lib.rs
```diff
@@ -247,11 +263,23 @@ impl BinaryOr for Value {
...
+    fn not(self) -> Self::Output {
+        match self {
+            Value::Boolean(b) => Value::from(!b),
+            Value::Null | Value::Missing => Value::Null,
+            _ => Value::Missing, // data type mismatch => Missing
+        }
+    }
+}
+
 impl ops::Not for Value {
```
alancai98, line 278: Should we delete the `ops::Not` trait that works on `Value` (not reference)? Or should both be kept?

partiql-value/src/lib.rs
```diff
@@ -191,6 +191,22 @@ impl ops::Neg for Value {
...
+            // TODO: handle overflow for negation
+            Value::Null => self,
+            Value::Missing => self,
+            Value::Integer(i) => Value::from(-i),
+            Value::Real(f) => Value::Real(-f),
+            Value::Decimal(d) => Value::from(-d),
+            _ => Value::Missing, // data type mismatch => Missing
+        }
+    }
+}
```
alancai98, line 208: Do we need the `ops::Not` trait implemented for both `Value` and for `&Value`?

alancai98 (approved):
Looks good. Just left a couple questions about whether the `Not` and `Neg` trait should be implemented for both `Value` and `&Value`

am357 (approved):

PR #233, opened 2022-12-09
[WIP] Implements LATERAL JOINs using existing DAG flows and eval node structure

alancai98:
[WIP] -- still testing if `LEFT` JOINs can be modeled using the existing DAG flows and evaluation node structure.

Models `CROSS` and `INNER` JOINs as lateral JOINs (i.e. make LHS JOIN bindings available to RHS JOIN expression), which is how the PartiQL spec defines `CROSS` and `INNER` JOINs. This allows us to remove the `CrossLateral` JOIN type specified in #227 and resolves the [comment](https://github.com/partiql/partiql-lang-rust/pull/227#discussion_r1038661582) mentioned in that PR. The current approach in this PR moves the "JOIN"-ing/chaining of tuples together to within the RHS scan, which keeps the previous DAG modeling.

This is still a POC as I'm still seeing how easily other JOINs can be implemented with this approach. I'm also experimenting with another approach that refactors how JOINs are modeled in the eval nodes and DAG.

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -145,24 +145,27 @@ impl Evaluable for EvalScan {
         let mut value = partiql_bag![];
         bindings.iter().for_each(|binding| {
             let v = self.expr.evaluate(&binding.as_tuple_ref(), ctx);
-            let ordered = &v.is_ordered();
-            let mut at_index_counter: i64 = 0;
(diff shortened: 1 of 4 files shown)


alancai98:
Closing as #236 is how we're currently implementing `LEFT` and `LATERAL` JOINs.

PR #236, opened 2022-12-15
Implements JOINs as LATERAL, implements LEFT JOINs, refactors JOIN implementation

alancai98:
*Issue #, if available:* None

*Description of changes:* Implements JOINs as being LATERAL JOINs (i.e. RHS of JOIN can reference variables from proceeding LHS) and adds implementation for LEFT JOINs. This PR refactors the JOIN `BindingOp`s quite a bit and no longer models the LEFT and RIGHT side inputs as separate DAG nodes. The new JOIN node now contains a left and right `BindingOp` (e.g. `Scan`, or `Unpivot`) that it will evaluate.

I believe the previous modeling was suitable for non-lateral JOINs (can probably do something similar for FULL JOINs), but it was tricky to adapt to LATERAL and LEFT JOINs. PoC of modeling LATERAL JOINs using the previous JOIN node structure #233.


partiql-eval/src/plan.rs
```diff
@@ -92,19 +92,27 @@ impl EvaluatorPlanner {
...
+                on,
+            }) => {
                 let kind = match kind {
-                    JoinKind::Inner => EvalJoinKind::Inner,
+                    JoinKind::Cross | JoinKind::Inner => EvalJoinKind::Inner,
                     JoinKind::Left => EvalJoinKind::Left,
                     JoinKind::Right => EvalJoinKind::Right,
                     JoinKind::Full => EvalJoinKind::Full,
-                    JoinKind::Cross => EvalJoinKind::Cross,
-                    JoinKind::CrossLateral => EvalJoinKind::CrossLateral,
```
alancai98, line 102: Model CROSS JOINs as INNER JOINs as mentioned by equivalence mentioned in [section 5.3 of spec](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.5.3).

partiql-eval/src/eval.rs
```diff
@@ -101,6 +103,9 @@ pub enum EvaluationError {
 pub trait Evaluable: Debug {
     fn evaluate(&mut self, ctx: &dyn EvalContext) -> Option<Value>;
     fn update_input(&mut self, input: &Value, branch_num: u8);
+    fn get_vars(&self) -> Vec<String> {
```
alancai98, line 106: This is used for `LEFT JOIN`s when the evaluation of the RHS returns an empty bag of bindings. From [section 5.4 of the spec](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.5.4), when this case occurs, we create binding with every variable defined from the RHS `BindingOp` (e.g. the `AS` and `AT` bindings for `Scan` and `Unpivot`). Not sure if there was a better way to get the variables defined by a `BindingOp`.

partiql-logical/src/lib.rs
```diff
@@ -86,8 +86,11 @@ where
         self.edges.push((src, dst, 0));
     }
 
-    /// Adds a data flow with a branch number. One application of branch number is in data flows to
-    /// [`Join`] operator to distinguish between LHS and RHS of a JOIN.
+    /// Adds a data flow with a branch number.
+    /// TODO: decide if `branch_num` is necessary within the current implementation. JOINs were
```
alancai98, line 90: Since I refactored the `EvalJoin` node to not require a left and right `input`, I'm not too sure if `branch_num`s are useful currently. They're probably useful for some other use case we'll have in the future. Updated the documentation to make that point clearer.

partiql-eval/src/eval.rs
```diff
@@ -182,96 +195,161 @@ pub enum EvalJoinKind {
...
 
 #[derive(Debug)]
 pub struct EvalJoin {
     pub kind: EvalJoinKind,
     pub on: Option<Box<dyn EvalExpr>>,
-    pub input_l: Option<Value>,
-    pub input_r: Option<Value>,
+    pub input: Option<Value>,
+    pub left: Box<dyn Evaluable>,
+    pub right: Box<dyn Evaluable>,
```
alancai98, line 206: This refactoring made it a lot easier to loop over the left bindings and right bindings (evaluated with respect to each left binding) as [spec 5.6](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.5.6) defines.

partiql-eval/src/plan.rs
am357: Also as a comment in the code?

am357 (approved):
Nice remodeling of the `Join` w/ _left_ and _right_ `BindingOp`; overall LGTM left some minor comments.

partiql-eval/src/plan.rs
alancai98: Added comment to the code

partiql-logical/src/lib.rs
alancai98: I'm also generally inclined to remove it and add back later if it's needed. I'll cut a separate issue (https://github.com/partiql/partiql-lang-rust/issues/237) to track and possibly remove so this PR's scope isn't increased too much.

partiql-value/src/lib.rs
```diff
@@ -1118,12 +1118,32 @@ impl Tuple {
...
         self.attrs.push(attr.to_string());
         self.vals.push(val);
     }
 
+    #[inline]
+    /// Creates a `Tuple` containing all the attributes and values provided by `self` and `other`.
+    /// TODO: This isn't quite the correct behavior for binding tuple concatenation specified in
+    ///  the spec section 3.4 (https://partiql.org/assets/PartiQL-Specification.pdf#subsection.3.4).
+    ///  Duplicates should be removed but currently aren't.
+    pub fn tuple_concat(&self, other: &Tuple) -> Self {
```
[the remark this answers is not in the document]
alancai98: Yeah, adding/combining two tuples with and without duplicates could have use cases for the `Tuple` API. I'll add a clarification comment that both those methods should probably exist. Or some other `Tuple` data structure filtering duplicates should exist.

partiql-logical/src/lib.rs
am357: Great, would be great to have it before release!

partiql-value/src/lib.rs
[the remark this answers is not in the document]
am357: In that case how about changing the name of this one?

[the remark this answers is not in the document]
alancai98: Okay, I'll change it to `add`. I'll move the TODO to the call site (within `EvalJoin`) in that case since the comment is more relevant for the usage.

jpschorr (approved):

partiql-value/src/lib.rs
```diff
@@ -1118,12 +1118,32 @@ impl Tuple {
...
     pub fn insert(&mut self, attr: &str, val: Value) {
         self.attrs.push(attr.to_string());
         self.vals.push(val);
     }
 
+    #[inline]
+    /// Creates a `Tuple` containing all the attributes and values provided by `self` and `other`.
+    /// TODO: This isn't quite the correct behavior for binding tuple concatenation specified in
+    ///  the spec section 3.4 (https://partiql.org/assets/PartiQL-Specification.pdf#subsection.3.4).
+    ///  Duplicates should be removed but currently aren't.
```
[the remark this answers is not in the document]
alancai98: Yeah, that works. Updated in [ba02ffa](https://github.com/partiql/partiql-lang-rust/pull/236/commits/ba02ffa39046d78cdfb6ab4dfaa3b04b17cb7799) to use `unique_by`. Had to also make some slight modifications to check `other` before `self` to comply with the spec behavior (i.e. keeps duplicate attributes from rhs).

Also changed the usages to use the corrected `tuple_concat` method.

partiql-value/src/lib.rs
```diff
@@ -1118,12 +1118,32 @@ impl Tuple {
         }
     }
 
+    /// Creates a `Tuple` with attributes `attrs`, each with value `default_val`.
+    pub fn new_with_default_val(attrs: Vec<String>, default_val: Value) -> Self {
```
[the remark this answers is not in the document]
alancai98: Agree on not seeing much use outside of the use in `EvalJoin`. Moved to a `fn` within `EvalJoin`'s `evaluate` method.

partiql-value/src/lib.rs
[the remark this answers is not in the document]
alancai98: Ended up fixing `tuple_concat` as per Josh's [comment](https://github.com/partiql/partiql-lang-rust/pull/236#discussion_r1049109515). Also deleted `add` as there's no current usage and we can always add it back in the future.

PR #243, opened 2022-12-20
Address binding case insensitivity TODOs; make PathComponent's Key BindingName rather than String

alancai98:
- Fixes some TODO's related to `Bindings` and `Tuple` case insensitivity
- Make `PathComponent`'s `Key` `BindingName` rather than a `String`
- Updates some benchmark code
- Makes `Tuple`'s `vals` field private

### DIFF
--- a/partiql-value/src/lib.rs
+++ b/partiql-value/src/lib.rs
@@ -11,6 +11,7 @@ use std::{ops, slice, vec};
 
 use rust_decimal::prelude::FromPrimitive;
 use rust_decimal::{Decimal as RustDecimal, Decimal};
+use unicase::UniCase;
 
 mod ion;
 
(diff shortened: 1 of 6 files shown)


partiql-value/src/lib.rs
```diff
@@ -1144,7 +1145,7 @@ impl Hash for Bag {
 #[derive(Default, Eq, Clone)]
 pub struct Tuple {
     attrs: Vec<String>,
-    pub vals: Vec<Value>,
+    vals: Vec<Value>,
```
alancai98, line 1148: Accidently made public in #236. Didn't need to be public

am357 (approved):

jpschorr (approved):

PR #252, opened 2022-12-21
Add ability to use `EvalExpr` in path components

jpschorr:
e.g. `[1,2,3][1+1]` <=> `[1,2,3][2]` <=> `3`

### DIFF
--- a/partiql-eval/src/lib.rs
+++ b/partiql-eval/src/lib.rs
@@ -1364,6 +1364,51 @@ mod tests {
     }
 
     #[test]
+    fn paths() {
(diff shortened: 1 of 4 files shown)


partiql-eval/src/lib.rs
```diff
@@ -1363,6 +1363,38 @@ mod tests {
         assert_matches!(out, Value::Integer(42));
     }
 
+    #[test]
+    fn paths() {
```
alancai98, line 1367: nit: could also add a test using `KeyExpr`

alancai98 (approved):

PR #253, opened 2022-12-21
PoC AST to Plan compilation and query evalution of conformance tests

jpschorr:
Provide a PoC end-to-end [query text + environment] -> [evaluated results]

- Adds a compilation from AST to Logical plan for a large subset of SELECT/WHERE/FROM, expressions, and some PartiQL pathing.
- Adds conformance test evaluation (currently assumes COERCE mode) and assertions.
- Adds *dynamic lookup* for runtime name resolution of variable references.

---


partiql-logical/src/lib.rs
```diff
@@ -139,6 +139,10 @@ where
         &self.edges
     }
 
+    pub fn operator(&self, id: OpId) -> Option<&T> {
+        self.nodes.get(id.0)
```
alancai98, line 143: This came about when I was testing some queries with JOINs using the AST to logical plan conversion. Don't think this is entirely correct. The `OpId`s are 1-indexed as far as I can tell. (Can look at how `add_operator(...)` which uses the number of operators https://github.com/partiql/partiql-lang-rust/blob/d3a4b46a55274368653166fed9555ab37ed36cab/partiql-logical/src/lib.rs#L74-L78). While the `nodes` list is a vec that is 0-indexed. I guess a simple fix could be to subtract 1:

```suggestion
        self.nodes.get(id.0 - 1)
```

Alternatively, could change `OpId` to be 0-indexed when adding to the logical plan (not sure how much refactoring would be required though).

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+        self.enter_env();
+    }
+
+    fn exit_order_by_expr(&mut self, _order_by_expr: &'ast OrderByExpr) {
+        let _env = self.exit_env();
+        todo!("order by clause");
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
```
alancai98, line 992: Could include some further documentation on how the name resolution and AST to Plan compilation works. Not sure if we want to include documentation before or after the 0.2.0 release, but having some supporting documentation could help with understanding the complicated parts of PartiQL's name resolution and how it's implemented.

alancai98 (commented):
Really great work getting the end-to-end query evaluation working! Left a lot of minor comments and questions to better understand how the AST to Plan compilation is implemented.

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+
+use std::collections::{HashMap, HashSet};
+
+use std::hash::Hash;
+
+use std::sync::atomic::{AtomicU32, Ordering};
+
+type FnvIndexSet<T> = IndexSet<T, FnvBuildHasher>;
+
+type FnvIndexMap<K, V> = IndexMap<K, V, FnvBuildHasher>;
```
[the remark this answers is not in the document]
jpschorr: "[IndexMap](https://docs.rs/indexmap/latest/indexmap/map/struct.IndexMap.html) is a hash table where the iteration order of the key-value pairs is independent of the hash values of the keys." (from https://docs.rs/indexmap/latest/indexmap/)

IndexMap/Set incur a small performance overhead for the guarantee that when we iterate over their contents, they will always iterate in the order that we inserted them. This is not necessarily important for correctness, but can be a great aid in debugging and testing. 

For example, in the expression `(1+2)*(3+4)`, it doesn't really matter whether we do `1+2` first or `3+4` first as long as both are available when we want to perform the `*`. But when debugging it is much easier if the order of the sub-expressions is stable between executions. This is also important if we want to write tests that assert that a given query text or AST compiles to an expected plan, as we don't need to worry about order-independent equality matching, etc.

For a discussion of what/why FNV, see https://docs.rs/fnv/latest/fnv/

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1391 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+                return ValueExpr::DynamicLookup(Box::new(lookups));
+            }
+        }
+
+        // TODO error?
```
am357, line 298: Would be great to have more context in this comment for future.

am357 (commented):
Thank you for all the effort! It's great to see this coming through; some of my comments were just echoing Alan's so I did not include them.

.github/workflows/ci_build_test.yml
```diff
@@ -2,13 +2,17 @@ name: CI Build
...
+  RUST_TEST_TIME_INTEGRATION: 150,5000
+
 jobs:
   build:
     name: Build and Test
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-20.04, windows-latest, macos-latest]
```
alancai98, line 15: Should also change the `ubuntu-latest` usage in https://github.com/partiql/partiql-lang-rust/blob/6c3bb697bde5152efeff60bd32f595879f26e70b/.github/workflows/coverage.yml#L18 to `ubuntu-20.04` to make the "Coverage" step succeed

partiql-logical-planner/src/lib.rs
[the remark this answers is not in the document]
alancai98: Got it. Thanks for the explanation and links!

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+        self.enter_env()
+    }
+
+    fn exit_sexp(&mut self, _sexp: &'ast Sexp) {
+        todo!("exit_sexp")
+    }
+
+    fn enter_call(&mut self, _call: &'ast Call) {
+        todo!("call")
+    }
```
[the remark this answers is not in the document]
jpschorr: Not necessary. The `enter` & `exit` calls are defined with a default implementation in https://github.com/partiql/partiql-lang-rust/blob/main/partiql-ast/src/visit.rs#L76

The `todo` here is just to make it clear during PoC why all function calls are failing to evaluate. I have a fast-follow PR that implements lowering of calls.

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+
+    fn enter_call_agg(&mut self, _call_agg: &'ast CallAgg) {
+        todo!("call_agg")
+    }
+
+    fn exit_call_agg(&mut self, _call_agg: &'ast CallAgg) {}
```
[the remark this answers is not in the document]
jpschorr: Not necessary. The `enter` & `exit` calls are defined with a default implementation in https://github.com/partiql/partiql-lang-rust/blob/main/partiql-ast/src/visit.rs#L76

The `todo` here is just to make it clear during PoC why all aggregation calls are failing to evaluate.

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+    pub consume: NameRefs,
+    pub produce_required: Names,
+    pub produce_optional: NameOptions,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum QNode {
+    FromLet,
+    Query,
+}
```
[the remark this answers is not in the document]
jpschorr: I'll add a comment to the code:
```
// The enclosing clause; used, in part, to track whether a name is a 'from path' reference
```

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,1397 @@
...
+        todo!("call")
+    }
+
+    fn enter_call_agg(&mut self, _call_agg: &'ast CallAgg) {
+        todo!("call_agg")
+    }
+
+    fn exit_call_agg(&mut self, _call_agg: &'ast CallAgg) {}
+
+    fn enter_var_ref(&mut self, _var_ref: &'ast VarRef) {
```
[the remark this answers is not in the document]
jpschorr: SQL (and therefore PartiQL) text (and therefore AST) is not lexically-scoped as is the case with most programming languages with which we are familiar.

In order to properly process expressions (especially the name references), this essentially processes the AST in a kind of post-order traversal, where most node processing is performed after that node's children. The `exit_<x>` calls correspond to the post-order processing. Often it is necessary to do a little bit of work (preparing data structures into which children will collect) in the `enter_<x>` methods. The children of the node are visited in between the `enter_<x>` and `exit_<x>` calls.

For 'leaf' nodes of the tree (e.g. variable references, etc.), the node doesn't have any children, so there is nothing done between the `enter_<x>` and `exit_<x>` calls. By convention, processing for them is done in the `enter_<x>` calls here.

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+}
+
+// The enclosing clause; used, in part, to track whether a name is a 'from path' reference
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+enum EnclosingClause {
+    FromLet,
+    Query,
+}
+
+/// Resolves which clauses in a query produce & consume variable references by walking the
```
am357, line 80:
```suggestion
/// Represents clauses in a query that produce & consume variable references by walking the
```

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        self.push_consume_name(name);
+    }
+
+    fn exit_project_expr(&mut self, project_expr: &'ast ast::ProjectExpr) {
+        let id = self.current_node();
+        // get the "as" alias
+        // 1. if explictly given
+        // 2. else try to infer if a simple variable reference or path
+        // 3. else it is currently 'Unknonw'
```
am357, line 315:
```suggestion
        // 3. else it is currently 'Unknown'
```

partiql-logical-planner/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+indexmap = "1.9"
+petgraph = "0.6.*"
+num = "0.4"
+fnv = "1"
+assert_matches = "1.5.*"
+
+[dev-dependencies]
+partiql-eval = { path = "../partiql-eval" }
+
+
```
am357, line 37:
```suggestion
```

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+
+    fn exit_from_let(&mut self, from_let: &'ast ast::FromLet) {
+        self.exit_child_stack();
+        let id = *self.current_node();
+        let KeyRefs { consume, .. } = self.exit_keyref();
+
+        // get the "as" alias
+        // 1. if explictly given
+        // 2. else try to infer if a simple variable reference or path
+        // 3. else it is currently 'Unknonw'
```
am357, line 265:
```suggestion
        // 3. else it is currently 'Unknown'
```

am357 (approved):

partiql-logical-planner/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        if let Ok(out) = plan.execute_mut(bindings) {
+            out.result
+        } else {
+            Value::Missing
+        }
+    }
+
+    #[track_caller]
+    fn evalute_query(query: &str) -> Value {
```
alancai98, line 62: nit: typo `evalute_query` -> `evaluate_query`. Also in the call sites on L109 and L116

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+
+        // Scopes above this `FROM` in the AST are in-scope to use variables defined by this from
+        for in_scope in self.id_path_to_root.iter().rev().skip(1) {
+            self.in_scope
+                .entry(*in_scope)
+                .or_insert_with(Vec::new)
+                .push(id);
+        }
+
+        // This `FROM` item is in-scope of variables defined by any preceeding items in this `FROM` (e.g., lateral joins)
```
alancai98, line 246:
```suggestion
        // This `FROM` item is in-scope of variables defined by any preceding items in this `FROM` (e.g., lateral joins)
```

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.lateral_stack.last_mut().unwrap().push(id);
+    }
+
+    fn exit_from_let(&mut self, from_let: &'ast ast::FromLet) {
+        self.exit_child_stack();
+        let id = *self.current_node();
+        let KeyRefs { consume, .. } = self.exit_keyref();
+
+        // get the "as" alias
+        // 1. if explictly given
```
alancai98, line 263:
```suggestion
        // 1. if explicitly given
```

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+        };
+
+        self.push_consume_name(name);
+    }
+
+    fn exit_project_expr(&mut self, project_expr: &'ast ast::ProjectExpr) {
+        let id = self.current_node();
+        // get the "as" alias
+        // 1. if explictly given
```
alancai98, line 313:
```suggestion
        // 1. if explicitly given
```

partiql-logical-planner/src/lower.rs
```diff
@@ -0,0 +1,1005 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+                return ValueExpr::DynamicLookup(Box::new(lookups));
+            }
+        }
+
+        // TODO in the presence of schema, error if the varialbe reference doesn't correspond to a data table
```
alancai98, line 304:
```suggestion
        // TODO in the presence of schema, error if the variable reference doesn't correspond to a data table
```

partiql-logical-planner/src/lib.rs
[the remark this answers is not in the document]
alancai98: Thanks for the explanation. That makes some more sense

alancai98 (approved):
Thanks for addressing feedback from the previous review. Looks good. Only left minor comments related to some typos.

PR #255, opened 2023-01-02
Do proper null-propagation for concat

jpschorr:
Do proper null-propagation for concat

---

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -945,10 +945,10 @@ impl EvalExpr for EvalBinOpExpr {
         fn short_circuit(op: &EvalBinOp, value: &Value) -> Option<Value> {
             match (op, value) {
                 (EvalBinOp::And, Boolean(false)) => Some(false.into()),
-                (EvalBinOp::And, Null) | (EvalBinOp::And, Missing) => Some(Null),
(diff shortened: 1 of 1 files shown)


partiql-eval/src/eval.rs
```diff
@@ -949,6 +949,7 @@ impl EvalExpr for EvalBinOpExpr {
                 (EvalBinOp::Or, Boolean(true)) => Some(true.into()),
                 (EvalBinOp::Or, Null) | (EvalBinOp::Or, Missing) => Some(Null),
                 (EvalBinOp::In, Null) | (EvalBinOp::In, Missing) => Some(Null),
+                (EvalBinOp::Concat, Null) | (EvalBinOp::Concat, Missing) => Some(Null),
```
alancai98, line 952: I don't think concat should be propagating `NULL` over `MISSING`. Based on the [conformance tests for concat](https://github.com/partiql/partiql-tests/blob/main/partiql-tests-data/eval/primitives/operators/concat.ion#L87-L145), `MISSING` is propagated over `NULL`. I think only the logical operators (i.e. `AND`, `OR`, `IN`) will propagate `NULL` over `MISSING`.

alancai98 (approved):

PR #258, opened 2023-01-05
Refactor evaluation to perform fewer clones of data between stages

jpschorr:
Refactor evaluation to  perform fewer clones of data between stages.

Locally this results in using only 30% of the memory and 40% of run-time when running the conformance tests suite (when the build is already cached):

Command:
```
/usr/bin/time -l cargo test --package partiql-conformance-tests --features "conformance_test" --release
```

Before:
```
       95.57 real       118.32 user        23.23 sys
          8482541568  maximum resident set size         
```
            
After:
```
       37.80 real        64.96 user         3.35 sys
          2491486208  maximum resident set size      
```

---


partiql-eval/src/eval.rs
```diff
@@ -397,12 +400,11 @@ impl Evaluable for EvalUnpivot {
             out.push(Value::Tuple(Box::new(t)));
         }
 
-        self.output = Some(Value::Bag(Box::new(Bag::from(out))));
-        self.output.clone()
```
alancai98, line 401: Just realized `self.output` wasn't being used at all for all of these nodes. Nice that we can remove the clones.

alancai98 (approved):
Nice performance improvements!

PR #266, opened 2023-01-09
Support optional `AT` for `UNPIVOT`

jpschorr:
Support optional AT for UNPIVOT

---

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -373,16 +373,16 @@ impl Evaluable for EvalJoin {
 pub struct EvalUnpivot {
     pub expr: Box<dyn EvalExpr>,
     pub as_key: String,
-    pub at_key: String,
+    pub at_key: Option<String>,
     pub input: Option<Value>,
 }
 
 impl EvalUnpivot {
-    pub fn new(expr: Box<dyn EvalExpr>, as_key: &str, at_key: &str) -> Self {
+    pub fn new(expr: Box<dyn EvalExpr>, as_key: &str, at_key: Option<String>) -> Self {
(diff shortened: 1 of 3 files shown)


partiql-eval/src/eval.rs
```diff
@@ -373,38 +373,42 @@ impl Evaluable for EvalJoin {
...
-            Tuple::from([(self.as_key.as_str(), v), (self.at_key.as_str(), k.into())])
-        });
-
-        for t in unpivoted {
-            out.push(Value::Tuple(Box::new(t)));
-        }
+        let out = tuple
+            .into_iter()
+            .filter_map(|(k, v)| match v {
+                Missing => None, // if Value is missing, don't add it to output tuple
```
alancai98, line 401: For this line, think the spec ([section 3.2](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.3.2)) says we shouldn’t filter `MISSING` from binding tuples
> Also important, is that when we represent binding tuples we explicitly represent a variable with a `MISSING` value, as opposed to omitting it because the lack of a variable name is distinct from a variable whose value is `MISSING`. For example, we write `<x : 1, y : MISSING>`, instead of `<x : 1>`.

That missing attribute will be removed in the `SELECT` clause. See [section 6.1.4](https://partiql.org/assets/PartiQL-Specification.pdf#subsubsection.6.1.4):
> Whenever during tuple construction an attribute value evaluates to `MISSING`, then
the particular attribute/value is omitted from the constructed tuple.

alancai98 (commented):
Looks good. Left one comment about filtering `MISSING` from the binding tuple created in `EvalUnpivot`

PR #272, opened 2023-01-10
Release v0.2.0

alancai98:
Releases v0.2.0. Also makes edits to the CHANGELOG and README.

### DIFF
--- a/partiql-eval/Cargo.toml
+++ b/partiql-eval/Cargo.toml
@@ -20,8 +20,8 @@ edition.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-partiql-logical = { path = "../partiql-logical" }
-partiql-value = { path = "../partiql-value" }
+partiql-logical = { path = "../partiql-logical", version = "*" }
+partiql-value = { path = "../partiql-value", version = "*" }
 petgraph = "0.6.*"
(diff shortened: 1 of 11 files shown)


partiql-eval/Cargo.toml
```diff
@@ -20,8 +20,8 @@ edition.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-partiql-logical = { path = "../partiql-logical" }
-partiql-value = { path = "../partiql-value" }
+partiql-logical = { path = "../partiql-logical", version = "*" }
```
alancai98, line 23: Would it be better to set it to the
- current workspace version ("0.2.0") <- would require updating for every release version
- or "*"?

alancai98: Nvm, we cannot use "*" when publishing to crates.io 

>error: failed to publish to registry at https://crates.io
Caused by:
  the remote server responded with an error: wildcard (`*`) dependency constraints are not allowed on crates.io. See https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies for more information

alancai98: Will change to use explicit version with wildcard -- e.g. 0.2.*

am357 (approved):

am357 (approved):

PR #275, opened 2023-01-17
Adds performance benchmarking to GitHub actions

jpschorr:
Uses https://github.com/benchmark-action/github-action-benchmark

Cf. [The action's rust example](https://github.com/benchmark-action/github-action-benchmark/tree/master/examples/rust)
Cf. [A similar PR for partiql-lang-kotlin](https://github.com/partiql/partiql-lang-kotlin/pull/921)

---

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,0 +1,53 @@
+# Do not run this workflow on pull request since this workflow has permission to modify contents.
+on:
+  push:
+    branches:
+      - main
+
+env:
+  AUTHOR: "@${{ github.event.commits[0].author.username }}"
+  REPO: "@${{ github.repository }}"
+  REPO_OWNER: "@${{ github.repository_owner }}"
+  ACTOR: "@${{ github.actor }}"
+
+permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,53 @@
+# Do not run this workflow on pull request since this workflow has permission to modify contents.
+on:
+  push:
+    branches:
+      - master
```
alancai98, line 5: Should this be `main` rather than `master`?

Cargo.toml
```diff
@@ -31,3 +31,6 @@ opt-level = 3
 [profile.release.build-override]
 opt-level = 3
 
+[profile.bench]
+lto = true
+
```
am357, line 36:
```suggestion
```

am357 (approved):

.github/workflows/benchmark.yml
jpschorr: yep. copy/paste error from example

alancai98 (approved):

PR #286, opened 2023-01-25
Implement `PIVOT` in evaluator

alancai98:
Implements `PIVOT` operator in evaluator. Defined in section [6.2 of PartiQL spec](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.6.2).

---

### DIFF
--- a/partiql-eval/src/eval.rs
+++ b/partiql-eval/src/eval.rs
@@ -383,6 +383,48 @@ impl Evaluable for EvalJoin {
     }
 }
 
+/// Represents an evaluation `Pivot` operator; the `Pivot` enables turning a collection into a
+/// tuple. For `Pivot` operational semantics, see section `6.2` of
+/// [PartiQL Specification — August 1, 2019](https://partiql.org/assets/PartiQL-Specification.pdf).
+#[derive(Debug)]
+pub struct EvalPivot {
+    pub input: Option<Value>,
+    pub key: Box<dyn EvalExpr>,
+    pub value: Box<dyn EvalExpr>,
+}
+
+impl EvalPivot {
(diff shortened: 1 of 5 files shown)


partiql-logical/src/lib.rs
```diff
@@ -187,6 +188,14 @@ pub struct Scan {
     pub at_key: Option<String>,
 }
 
+/// [`Pivot`] represents a PIVOT operator, e.g. `PIVOT sp.price AT sp."symbol` in
+/// `PIVOT sp.price AT sp."symbol" FROM todaysStockPrices sp`.
```
am357, line 192: A reference to the spec would also be helpful here as well.

am357 (approved):

partiql-eval/src/eval.rs
```diff
@@ -383,6 +383,48 @@ impl Evaluable for EvalJoin {
...
+
+        let mut out: Tuple = partiql_tuple![];
+        for binding in input_value.into_iter() {
+            let binding = binding.coerce_to_tuple();
+            let key = self.key.evaluate(&binding, ctx);
+            let value = self.value.evaluate(&binding, ctx);
+
+            if let Value::String(s) = key {
+                out.insert(&s, value)
+            }
```
jpschorr, line 418:
```suggestion
            if let Value::String(s) = key {
                let value = self.value.evaluate(&binding, ctx);
                out.insert(&s, value)
            }
```

jpschorr (approved):

partiql-eval/src/eval.rs
alancai98: Adjusted to your suggestion

PR #297, opened 2023-02-10
Add some convenience methods and cleanup of `Value`

jpschorr:
- Add `Extend` implementations for `List` and `Bag`
- Add methods to iterate a `Tuple`'s values without zipping its names
- Allow `collect()` into a `Tuple` with any `Into<String>`
- other minor cleanup

---

### DIFF
--- a/partiql-value/src/lib.rs
+++ b/partiql-value/src/lib.rs
@@ -498,6 +498,11 @@ impl Value {
     }
 
     #[inline]
+    pub fn is_tuple(&self) -> bool {
+        matches!(self, Value::Tuple(_))
+    }
+
+    #[inline]
     pub fn is_list(&self) -> bool {
         matches!(self, Value::List(_))
     }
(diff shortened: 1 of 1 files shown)


partiql-value/src/lib.rs
```diff
@@ -887,6 +892,21 @@ impl List {
     pub fn iter(&self) -> ListIter {
         ListIter(self.0.iter())
     }
+
+    #[inline]
+    pub fn reserve(&mut self, additional: usize) {
```
alancai98, line 897: Just curious, is `reserve` used for some performance improvements?

partiql-value/src/lib.rs
```diff
@@ -497,6 +497,11 @@ impl Value {
         ion::parse_ion(ion)
     }
 
+    #[inline]
+    pub fn is_tuple(&self) -> bool {
```
alancai98, line 501: Could we add these new convenience methods to the CHANGELOG?

alancai98 (approved):

partiql-value/src/lib.rs
jpschorr: The usage in `extend` is idiomatic. Yes, it can be more performant to perform any allocations/moves prior to inserting a sequence of values rather than in the midst.

PR #304, opened 2023-02-16
Build Aggregate function recognizer only once, using OnceCell

jpschorr:
Should fix the parsing performance benchmark regression.

Cf. http://partiql.org/partiql-lang-rust/dev/bench/
Cf. https://github.com/partiql/partiql-lang-rust/commit/320c6297315465f6a97297b6d356a55bbadc4894#commitcomment-100912787

---


alancai98 (approved):
Nice to see the performance regression CI find the regression!

am357 (approved):

PR #319, opened 2023-03-16
Add timestamp value and parsing from ion

jpschorr:
- Adds  a `Time` type to `Value`
- Adds test value parsing of timestamps
- Updates `Value` Ion parsing to be the same as "test value" parsing
   - Perhaps we should just remove "test value" parsing?
- Updates to an unreleased version of `ion-rs` that includes timestamp accessors 

---


partiql-value/src/time.rs
```diff
@@ -1,191 +1,138 @@
...
+                write!(f, "TIME '{:?}'", t)
+            }
+            TimeValue::TimeWithTz(t, tz) => {
+                write!(f, "TIME WITH TIME ZONE '{:?} {:?}'", t, tz)
+            }
+            TimeValue::DateTime(dt) => {
+                write!(f, "TIMESTAMP '{:?}'", dt)
+            }
+            TimeValue::DateTimeWithTz(dt) => {
+                write!(f, "TIMESTAMP WITH TIME ZONE '{:?}'", dt)
```
alancai98, line 97: As a side note, even though the SQL92 spec defines timestamps this way, I don't think the spec does this formatting:

Found [page 8 figure 2](https://partiql.org/assets/PartiQL-Specification.pdf#page=8) writes a timestamp with the Ion value backticks:
```
{
    ...
    ‘configurationItemCaptureTime’: `2016-08-03T08:56:52.415Z`,
    ...
}
```
while later in the spec [section 11.3](https://partiql.org/assets/PartiQL-Specification.pdf#subsection.11.3) it includes a “timestamp” (really is a time value)
```
logs: [{‘sensor’:1, ‘co’:0.4, ‘timestamp’:04:05:06},
       {‘sensor’:1, ‘co’:0.2, ‘timestamp’:04:05:07},
       {‘sensor’:1, ‘co’:0.5, ‘timestamp’:04:05:10},
       {‘sensor’:2, ‘co’:0.3}
]
```

Also the Kotlin reference implementation doesn't support timestamps entered using the SQL92 syntax. And functions outputting timestamps output using the Ion value backticks.

Perhaps we can cut issues to `partiql-lang-kotlin` and/or `partiql-spec` to support the SQL-92 syntax.

deny.toml
alancai98, line 1: Can we update the CHANGELOG to account for the new TIMESTAMP value?

alancai98 (commented):
Looks good overall. Left a minor comment about some naming and some other nits/observations.

partiql-conformance-tests/tests/test_value.rs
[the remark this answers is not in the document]
jpschorr: There's no trickiness. There's just no specification for exactly how to parse embedded ion.

The `test_value` stuff is parsing as per the `partiql-tests`. Currently, the embedded ion parsing does the same because the only example it has to parse are the tests...

partiql-value/src/lib.rs
```diff
@@ -46,6 +48,7 @@ pub enum Value {
     Decimal(RustDecimal),
     String(Box<String>),
     Blob(Box<Vec<u8>>),
+    Time(Box<DateTime>),
```
alancai98, line 51: Could we also rename the `Value` enum to be `DateTime` rather than `Time`?

partiql-value/src/time.rs
alancai98, line 1: nit: could also rename the file from `time.rs` to `datetime.rs`

partiql-conformance-tests/tests/test_value.rs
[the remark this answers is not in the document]
alancai98: > There's just no specification for exactly how to parse embedded ion.

That's true. I couldn't find anywhere in the spec regarding how embedded Ion was to be parsed. Created an issue to track: https://github.com/partiql/partiql-spec/issues/47

alancai98 (approved):

PR #337, opened 2023-04-05
Implements `ORDER BY`

jpschorr:
Implements lowering & evaluation of `ORDER BY` clause.

Some `ORDER BY` conformance tests still fail due to a lack of proper name resolution and data flow analysis, which will be addressed separately.

---


am357 (approved):
There is couple clippy warnings but overall LGTM

partiql-logical-planner/src/lower.rs
```diff
@@ -34,6 +36,7 @@ type FnvIndexMap<K, V> = IndexMap<K, V, FnvBuildHasher>;
 enum QueryContext {
     FromLet,
     Path,
+    Order,
```
alancai98, line 39: I see `QueryContext::Order` is only used in `enter_sort` but not used anywhere else. Is this intentional? Or part of some upcoming data flow analysis work?

alancai98 (commented):
Looks good. Other than the clippy warnings, left a question about `QueryContext::Order` in lower.rs

alancai98 (approved):

PR #340, opened 2023-04-13
Add EXTRACT builtin function; update `partiql-tests`

alancai98:
Implements `EXTRACT` builtin function and updates to the latest `partiql-tests`. Also fixes some parsing of the datetime part for `YEAR`, `TIMEZONE_HOUR`, and `TIMEZONE_MINUTE`. Makes the following assumptions:
- `EXTRACT` with a date component (year, month, day) on a time -> returns missing
- `EXTRACT` with a time component (hour, minute, second) -> returns missing
- a timezone part (timezone_hour, timezone_minute) with a datetime value without a timezone?

All of the above assumptions aren't specified in any SQL spec to my knowledge. The second assumption differs from what `partiql-lang-kotlin` currently does (which defines a default of `0`). Created a spec issue to track: https://github.com/partiql/partiql-spec/issues/53 along with a conformance test issue: https://github.com/partiql/partiql-tests/issues/83.

Another point that may differ from `partiql-lang-kotlin` is the data type returned by the `EXTRACT` function. `partiql-lang-kotlin` always returns a decimal, while this PR returns a decimal for seconds and integer otherwise. The SQL spec says that this is implementation-defined. Add a followup comment in the above spec issue: https://github.com/partiql/partiql-spec/issues/53#issuecomment-1511949884

---

### DIFF
--- a/partiql-eval/src/eval/expr/mod.rs
+++ b/partiql-eval/src/eval/expr/mod.rs
@@ -5,7 +5,8 @@ use itertools::Itertools;
 use partiql_logical::Type;
 use partiql_value::Value::{Boolean, Missing, Null};
 use partiql_value::{
-    Bag, BinaryAnd, BinaryOr, BindingsName, List, NullableEq, NullableOrd, Tuple, UnaryPlus, Value,
+    Bag, BinaryAnd, BinaryOr, BindingsName, DateTime, List, NullableEq, NullableOrd, Tuple,
+    UnaryPlus, Value,
 };
 use regex::{Regex, RegexBuilder};
 use std::borrow::{Borrow, Cow};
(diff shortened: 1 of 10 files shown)


partiql-logical-planner/src/call_defs.rs
```diff
@@ -441,6 +441,118 @@ fn function_call_def_cardinality() -> CallDef {
...
+    CallDef {
+        names: vec!["extract"],
+        overloads: vec![
+            CallSpec {
+                input: vec![
+                    CallSpecArg::Named("year".into()),
+                    CallSpecArg::Named("from".into()),
+                ],
+                output: Box::new(|mut args| {
+                    args.remove(0); // remove first default synthesized argument
```
am357, line 454: Where is the default _arg_ coming from?

am357 (commented):
Overall LGTM; left couple of minor comments and one question?

partiql-logical-planner/src/call_defs.rs
alancai98: The default arg is from the preprocessor step that names the arguments:

https://github.com/partiql/partiql-lang-rust/blob/22379b9599d5551d6ec564052adc4ab4e77c624c/partiql-parser/src/preprocessor.rs#L92-L102

So for `EXTRACT(YEAR FROM x)`, the preprocessor would name the arguments: `YEAR: true` and `FROM: x`. Since we're looking at the `EXTRACT` function with a first `CallSpecArg` of "year" here, we can drop the first argument for `args`.

am357: Gotcha, maybe we can add the context to the comments for future?

am357 (approved):

partiql-logical-planner/src/call_defs.rs
alancai98: Sure, I can add that that arg comes from the parser preprocessor.

partiql-eval/src/eval/expr/mod.rs
```diff
@@ -935,3 +937,213 @@ impl EvalExpr for EvalFnCardinality {
         Cow::Owned(result)
     }
 }
+
+/// Represents a year `EXTRACT` function, e.g. `extract(YEAR FROM t)`.
+#[derive(Debug)]
+pub struct EvalFnExtractYear {
```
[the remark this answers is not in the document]
alancai98: Good callout. I believe we could make all the `EvalExpr` types crate public and most of the `Evaluable` types. There are some usages of `Evaluable` structs that would need a public visibility, but perhaps we could change the usage.

Created a separate issue to track: https://github.com/partiql/partiql-lang-rust/issues/341 just so this PR isn't too cluttered. I'll take a look at this soon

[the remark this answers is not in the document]
am357: Great—thank you!

partiql-value/src/datetime.rs
```diff
@@ -48,34 +49,46 @@ impl DateTime {
...
 
-    pub fn from_ymdhms(
+    pub fn from_ymdhms_nano_offset_minutes(
         year: i32,
         month: NonZeroU8,
         day: u8,
         hour: u8,
         minute: u8,
-        second: f64,
+        second: u8,
```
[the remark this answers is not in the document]
alancai98: I think it would be fine to leave it as a warning for now. We might need to adjust some of these datetime functions when we sort out the date/time modeling in the conformance tests https://github.com/partiql/partiql-tests/issues/82 and also support the date time literals.

jpschorr (approved):
My only comment was the extraneous `% 60` in `EvalFnExtractTimezoneMinute` which was removed in the latest commit.

PR #360, opened 2023-05-08
Add SQL aggregates ANY, SOME, EVERY and their COLL_ versions

### LINKED ISSUE #419: Implement Strict mode errors for `COLL_*` functions
In strict mode, if the element is not a number and not null/missing, this should error? Similarly for the other `COLL_` functions that expect a number

_Originally posted by @alancai98 in https://github.com/partiql/partiql-lang-rust/pull/413#discussion_r1278145811_

alancai98:
Adds the SQL aggregate functions `ANY`, `SOME`, and `EVERY` along with their `COLL_` versions (i.e. `COLL_ANY`, `COLL_SOME`, `COLL_EVERY`). 

This PR also
- fixes a slight bug related to the previously implemented SQL aggregates when an aggregation result is `NULL` or `MISSING`.
- closes #419 with some fixes to the strict mode type checking behavior of the `COLL_` functions.
- adds a check to logical plan lowering for `HAVING` without `GROUP BY`

### DIFF
--- a/partiql-logical-planner/src/lower.rs
+++ b/partiql-logical-planner/src/lower.rs
@@ -707,8 +707,13 @@ impl<'a, 'ast> Visitor<'ast> for AstToLogical<'a> {
         Traverse::Continue
     }
 
-    fn enter_select(&mut self, _select: &'ast Select) -> Traverse {
-        Traverse::Continue
+    fn enter_select(&mut self, select: &'ast Select) -> Traverse {
+        if select.having.is_some() && select.group_by.is_none() {
+            self.errors.push(AstTransformError::HavingWithoutGroupBy);
+            Traverse::Stop
+        } else {
+            Traverse::Continue
+        }
     }
 
     fn exit_select(&mut self, _select: &'ast Select) -> Traverse {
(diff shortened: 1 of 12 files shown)


alancai98:
Marking PR as draft. Will rebase on #413 once that PR is merged in to account for evaluation function binding changes.

alancai98:
Test regression is partially due to a test misclassification that is corrected in: https://github.com/partiql/partiql-tests/pull/107.

Basically, the test (`SELECT rep, SUM(total_sales) as total FROM sales_report HAVING rep = \"Meg\"`) expects an evaluation error, but was previously giving the wrong error. It was previously giving an error that the aggregation was empty when the error should be that a `HAVING` clause was provided without a `GROUP BY`. This error should be caught during lowering, hence the conformance test PR to reclassify the error as a `StaticAnalysisFail` rather than `EvaluationFail`.

partiql-logical-planner/src/lower.rs
```diff
@@ -707,8 +707,13 @@ impl<'a, 'ast> Visitor<'ast> for AstToLogical<'a> {
...
 
-    fn enter_select(&mut self, _select: &'ast Select) -> Traverse {
-        Traverse::Continue
+    fn enter_select(&mut self, select: &'ast Select) -> Traverse {
+        if select.having.is_some() && select.group_by.is_none() {
+            self.errors.push(AstTransformError::HavingWithoutGroupBy);
+            Traverse::Stop
+        } else {
+            Traverse::Continue
+        }
```
alancai98, line 716: Some context for why this change is in this PR provided in this comment: https://github.com/partiql/partiql-lang-rust/pull/360#issuecomment-1666275251

am357 (approved):

jpschorr (approved):

Issue #349, opened 2023-04-28
Better error handling ergonomics in evaluation


alancai98:
Currently in [evaluable.rs](https://github.com/partiql/partiql-lang-rust/blob/aa107b605f1a61054a989a5fc4bc3211e54d77b6/partiql-eval/src/eval/evaluable.rs) and in other places (e.g. AST lowering), there are a lot of possible panic locations through calling `unwrap` and `expect`. We should likely change such functions to use [`Result`](https://doc.rust-lang.org/std/result/) types.

This came up when I was upgrading the partiql-rust-cli to v0.3.0 and adding evaluation to the REPL. Some queries would panic and cause the REPL to exit. To catch the panic, I would need to use [`catch_unwind`](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) which requires the provided closure to be `UnwindSafe`. The `evaluate` function takes in a mutable reference which will make the closure not `UnwindSafe`. There is a workaround using [`AssertUnwindSafe`](https://doc.rust-lang.org/std/panic/struct.AssertUnwindSafe.html), but I'm not sure this is the best long-term approach.

PR #361, opened 2023-05-11
Change `lower` to return `Result` rather than `panic`

alancai98:
Partially addresses #349.

Changes the AST to logical plan `lower` function to return a `Result` type. This is part of an effort to reduce the `panic`s within the APIs. Also changes the `Visitor` API provided by `partiql-ast` to return a `Recurse` enum indicating whether visitation into the child nodes should continue or stop.

Conformance test changes are due to updating the `partiql-tests` commit.

### DIFF
--- a/partiql-logical-planner/src/lower.rs
+++ b/partiql-logical-planner/src/lower.rs
@@ -164,6 +164,25 @@ fn infer_id(expr: &ValueExpr) -> Option<SymbolPrimitive> {
     }
 }
 
+fn eq_or_error<V>(l: V, r: V, msg: &str) -> Result<(), LowerError>
+where
+    V: PartialEq,
+{
+    if l != r {
+        Err(LowerError::IllegalState(msg.to_string()))
+    } else {
+        Ok(())
+    }
+}
+
+fn true_or_error(b: bool, msg: &str) -> Result<(), LowerError> {
+    if !b {
+        Err(LowerError::IllegalState(msg.to_string()))
+    } else {
+        Ok(())
+    }
+}
+
 impl AstToLogical {
     pub fn new(registry: name_resolver::KeyRegistry) -> Self {
         let fnsym_tab: &FnSymTab = &FN_SYM_TAB;
@@ -200,9 +219,9 @@ impl AstToLogical {
     pub fn lower_query(
         mut self,
         query: &ast::AstNode<ast::Query>,
-    ) -> logical::LogicalPlan<logical::BindingsOp> {
-        query.visit(&mut self).expect("todo error handling");
-        self.plan
+    ) -> Result<logical::LogicalPlan<logical::BindingsOp>, LowerError> {
+        query.visit(&mut self)?;
+        Ok(self.plan)
     }
 
     #[inline]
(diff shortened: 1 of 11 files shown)


am357 (approved):
Thanks adding `Recurse`! seemingly it gives a good flexibility to AST visitation.

partiql-logical-planner/src/name_resolver.rs
```diff
@@ -140,8 +152,8 @@ impl NameResolver {
     }
 
     #[inline]
-    fn exit_lateral(&mut self) -> Vec<ast::NodeId> {
-        self.lateral_stack.pop().expect("lateral level")
+    fn exit_lateral(&mut self) -> Option<Vec<ast::NodeId>> {
```
jpschorr, line 155: I think it might be cleaner to construct error messages here than at each call site.

E.g., 
```rust
    fn exit_lateral(&mut self) -> Result<Vec<ast::NodeId>, LowerError> {
        self.lateral_stack
            .pop()
            .ok_or_else(|| LowerError::IllegalState("Expected non-empty lateral stack".to_string()))
    }
```
Then each call site could be simplified to:
```rust
        if let Err(e) = self.exit_lateral() {
            self.errors.push(e);
            return Recurse::Stop;
        }
```

Similar for the other state stack methods.

partiql-logical-planner/src/lower.rs
```diff
@@ -543,118 +613,192 @@ impl<'ast> Visitor<'ast> for AstToLogical {
...
     }
 
-    fn enter_query_set(&mut self, _query_set: &'ast QuerySet) {
+    fn enter_query_set(&mut self, _query_set: &'ast QuerySet) -> Recurse {
         self.enter_env();
 
         match _query_set {
-            QuerySet::SetOp(_) => todo!("QuerySet::SetOp"),
+            QuerySet::SetOp(_) => {
+                self.errors
```
jpschorr, line 632: Might be worth a private helper method like your `true_or_error` method to encapsulate the creation of the `NotYetImplemented` error at these call sites.

partiql-logical-planner/src/lower.rs
```diff
@@ -543,118 +613,192 @@ impl<'ast> Visitor<'ast> for AstToLogical {
...
+        }
         let env = self.exit_env();
 
         let select: BindingsOp = match _projection_kind {
             ProjectionKind::ProjectStar => logical::BindingsOp::ProjectAll,
             ProjectionKind::ProjectList(_) => {
+                if self.true_or_error(env.len().is_even(), "env.len() is not even") == Recurse::Stop
+                {
+                    return Recurse::Stop;
+                }
```
jpschorr, line 749: It may be worth considering turning `true_or_error` and `eq_or_error` into macros, as a macro could return from the current `fn` without needing the `if ... = Stop { return Stop; }` pattern.  

You may need both a 'critical' and a 'recoverable' version of each; with the 'critical' one immediately returning `Stop`, and the 'recoverable' just pushing to the `errors` state. Having 'critical_err' vs 'recoverable_err' (or something like that) in the name would also go a long way towards making it clear what errors are 'stop the world' and which we allow lowering to continue, if only to build up more error state; This is accomplished now with comments like `// Report error but allow visitor to continue` or `// Include as an error but allow lowering to proceed for multiple error reporting`

alancai98: Agree on the clarity that the macro would provide. Almost ended up creating them but wasn't sure if it was overkill. I'll change these call sites to use a macro in an upcoming commit.

partiql-logical-planner/src/lower.rs
alancai98: Yeah, there's some redundant code there. Probably will create a macro for the `NotYetImplemented` similar to the other errors.

jpschorr (approved):

partiql-logical-planner/src/lower.rs
```diff
@@ -842,21 +974,39 @@ impl<'ast> Visitor<'ast> for AstToLogical {
             Lit::IonStringLit(s) => parse_embedded_ion_str(s),
```
jpschorr, line 974: You'll need to fix `parse_embedded_ion_str` to not expect and to return a `Result`, I think

alancai98: Resolved in https://github.com/partiql/partiql-lang-rust/pull/366

Issue #356, opened 2023-05-08
Write `partiql_value::Value` using `ion_rs::IonWriter`


dlurton:
The `partiql_value::parse_ion` function currently exists to parse an Ion value.

However, nothing exists to go the other direction--once I've run a query, which yields a `partiql_value::Value`, I need to convert the result to Ion.  

Releated to #357

PR #362, opened 2023-05-16
Add `partiql-extension-ion` extension for encoding/decoding Ion from/into PartiQL `Value`s

### LINKED ISSUE #352: When fed invalid Ion text, partiql_value::parse_ion panics instead of returning an Err
In my partiql application, we would like to be able to gracefully handle the case where our user has fed us malformed ion, but the [`parse_ion` function](https://github.com/partiql/partiql-lang-rust/blob/aa107b605f1a61054a989a5fc4bc3211e54d77b6/partiql-value/src/ion.rs#LL16C1-L16C1) panics instead.

jpschorr:
Fixes #352 and #356

- Extracts ion decoding into `Value` from the various places it was in the codebase and centralizes it into `partiql-extension-ion`
- Updates the decoding to remove panics and instead return `Result` types
- Adds encoding of `Value` back into Ion
- Supports both 'unlifted' Ion to Value as well as the 'partiql encoded in ion' used by both the conformance tests and the partiql-lang-kotlin implementation
- Abstracts encoding/decoding to/from IonWriter/IonReader

---


extension/partiql-extension-ion/src/decode.rs
```diff
@@ -0,0 +1,512 @@
...
+    #[inline]
+    fn decode_struct(&self, reader: &mut Reader) -> IonDecodeResult {
+        if has_annotation(reader, TIME_ANNOT) {
+            self.decode_time(reader)
+        } else {
+            decode_struct(self, reader)
+        }
+    }
+
+    delegate! {
```
alancai98, line 499: Oh that's a cool macro. Definitely reduces a lot of boilerplate

extension/partiql-extension-ion/src/encode.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// Create the builder from 'config'
+    pub fn new(config: IonEncoderConfig) -> Self {
+        Self { config }
+    }
+
+    /// Create a encoder given the previously specified config and an ion [`IonWriter`].
+    pub fn build<'a, W, I>(
+        self,
+        writer: &'a mut I,
+    ) -> Result<Box<dyn ValueEncoder<W, I> + 'a>, IonEncodeError>
```
alancai98, line 78: Similar to comment about the Ion decoder, is the result type necessary?

extension/partiql-extension-ion/src/encode.rs
```diff
@@ -0,0 +1,417 @@
...
+/// An encoder which will write [`Value`]s as Ion stream values.
+pub trait ValueEncoder<W, I>
+where
+    W: Write,
+    I: IonWriter<Output = W>,
+{
+    /// A reference to the writer used by this encoder.
+    fn writer(&mut self) -> &mut I;
+
+    /// Write a Ion stream value from the given [`Value`]
```
alancai98, line 110:
```suggestion
    /// Write an Ion stream value from the given [`Value`]
```

extension/partiql-extension-ion/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    use super::*;
+    use crate::decode::{IonDecodeResult, IonDecoderBuilder, IonDecoderConfig};
+    use crate::encode::{IonEncodeError, IonEncoderBuilder, IonEncoderConfig};
+    use itertools::Itertools;
+
+    use partiql_value::{partiql_bag, partiql_list, partiql_tuple, DateTime, Value};
+    use rust_decimal_macros::dec;
+    use std::num::NonZeroU8;
+
+    pub fn decode_ion(contents: &str, encoding: Encoding) -> IonDecodeResult {
```
alancai98, line 18: Doesn't need to be `pub`

```suggestion
   fn decode_ion(contents: &str, encoding: Encoding) -> IonDecodeResult {
```

alancai98 (approved):
Looks good overall. I think the factoring for decoding/encoding Ion into its own crate is pretty clean and makes sense as its own crate.

Left some minor comments related to some typos and a question about whether the decoder/encoder `build` functions would need a `Result` type.

partiql-logical-planner/src/lower.rs
```diff
@@ -1429,3 +1431,21 @@ impl<'ast> Visitor<'ast> for AstToLogical {
         }))
     }
 }
+
+// TODO should this support partiql encoded in ion or only straight ion
```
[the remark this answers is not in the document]
am357: To @alancai98's note: maybe we can have this support as a Cargo feature?

am357 (commented):
Having this extension package is great! left just one comment regarding the PartiQL encoding in Ion.

extension/partiql-extension-ion/src/encode.rs
jpschorr: You're right in that it is currently infallible. I modeled it with a result because I'm not certain that it will remain so if additional config options are added...

partiql-logical-planner/src/lower.rs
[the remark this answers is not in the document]
jpschorr: I changed it to be 'straight' ion, as i think that's most reasonable, given the 'partiql-encoded' ion is not specified anywhere; it is a de-facto serialization from partiql-tests and created in the partiql-lang-kotlin library.

I'd rather not have it configurable at compile time. If we need it, I think we create a function that can read such strings (or a custom type, etc).

partiql-logical-planner/src/lower.rs
```diff
@@ -1429,3 +1431,19 @@ impl<'ast> Visitor<'ast> for AstToLogical {
...
```
am357, line 1435: Maybe we can remove this TODO now.

am357 (approved):

partiql-logical-planner/src/lower.rs
```diff
@@ -1429,3 +1431,21 @@ impl<'ast> Visitor<'ast> for AstToLogical {
...
+// TODO remove expects
```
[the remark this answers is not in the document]
jpschorr: Done in #366

partiql-logical-planner/src/lower.rs
jpschorr: Fixed in #366

PR #363, opened 2023-05-17
Changes logical plan to eval plan API to return `Result` type

### LINKED ISSUE #349: Better error handling ergonomics in evaluation

alancai98:
Part of #349.

Changes the logical to eval plan lowering API to use a `Result` type rather than panicking on errors. Other evaluation APIs will be changed in an upcoming PR.

### DIFF
--- a/partiql-eval/src/error.rs
+++ b/partiql-eval/src/error.rs
@@ -1,0 +1,61 @@
+use crate::eval::evaluable::Evaluable;
+use crate::eval::expr::EvalExpr;
+use crate::eval::EvalContext;
+use partiql_value::{Tuple, Value};
+use std::borrow::Cow;
+use thiserror::Error;
+
+/// All errors that occur during evaluation.
+#[derive(Debug)]
+pub struct EvalErr {
+    pub errors: Vec<EvaluationError>,
+}
+
+/// An error that can happen during evaluation.
+/// TODO: decide on splitting out logical to eval plan to separate error enum
+#[derive(Error, Debug, Clone, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum EvaluationError {
+    /// Malformed evaluation plan with graph containing cycle.
+    #[error("Evaluation Error: invalid evaluation plan detected `{0}`")]
+    InvalidEvaluationPlan(String),
+    /// Feature has not yet been implemented.
+    #[error("Not yet implemented: {0}")]
+    NotYetImplemented(String),
+    /// Internal error that was not due to user input or API violation.
+    #[error("Illegal State: {0}")]
+    IllegalState(String),
+    /// Invalid number of arguments to the function call.
+    #[error("Invalid number of arguments: {0}")]
+    InvalidNumberOfArguments(String),
+    /// Invalid escape pattern.
+    #[error("Invalid LIKE expression pattern: {0}")]
+    InvalidLikeEscape(String),
+}
+
+/// Used when an error occurs during the the logical to eval plan conversion. Allows the conversion
+/// to continue in order to report multiple errors.
+#[derive(Debug)]
+pub(crate) struct ErrorNode {}
+
+impl ErrorNode {
+    pub(crate) fn new() -> ErrorNode {
+        ErrorNode {}
+    }
+}
+
+impl Evaluable for ErrorNode {
+    fn evaluate(&mut self, _ctx: &dyn EvalContext) -> Option<Value> {
+        panic!("ErrorNode will not be evaluated")
+    }
+
+    fn update_input(&mut self, _input: Value, _branch_num: u8) {
+        panic!("ErrorNode will not be evaluated")
+    }
+}
+
+impl EvalExpr for ErrorNode {
+    fn evaluate<'a>(&'a self, _bindings: &'a Tuple, _ctx: &'a dyn EvalContext) -> Cow<'a, Value> {
+        panic!("ErrorNode will not be evaluated")
+    }
+}
--- a/partiql-eval/src/error.rs
+++ b/partiql-eval/src/error.rs
@@ -25,9 +25,6 @@ pub enum EvaluationError {
     /// Internal error that was not due to user input or API violation.
     #[error("Illegal State: {0}")]
     IllegalState(String),
-    /// Invalid number of arguments to the function call.
-    #[error("Invalid number of arguments: {0}")]
-    InvalidNumberOfArguments(String),
     /// Invalid escape pattern.
     #[error("Invalid LIKE expression pattern: {0}")]
     InvalidLikeEscape(String),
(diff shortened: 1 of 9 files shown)


dlurton:
My opinions, for what they're worth:

> Should evaluate return a Result type or Option type?

`Result<T, E>`!  Not sure what the use of returning an `Option<T>` would be.  If evaluation fails, we should be able to provide helpful error messaging that explains why, and you can't easily do that with an `Option<T>`.

> Should evaluate support multiple errors?

Probably not since evaluation should stop at the first error.

alancai98:
> There's enough difference between planning and evaluating, that it's probably better to deal with each phase as it's own PR.

Got it. I'll break out the other API(s) into separate PR(s).

partiql-logical-planner/src/error.rs
```diff
@@ -0,0 +1,25 @@
+use thiserror::Error;
+
+/// Contains the errors that occur during AST to logical plan conversion
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct LoweringError {
+    pub errors: Vec<LowerError>,
+}
```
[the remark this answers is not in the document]
alancai98: Currently the line/col numbers aren't plumbed through yet. They likely will be plumbed through to evaluation at some point like they are included with the lexer/parser errors.

Is this something you need soon?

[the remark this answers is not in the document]
dlurton: I don't need it soon.  I just want to make sure you're thinking about this since we'll definitely want it in the future.

partiql-eval/src/plan.rs
```diff
@@ -29,17 +30,49 @@ use crate::eval::expr::{
...
 
 impl EvaluatorPlanner {
+    pub fn new() -> EvaluatorPlanner {
+        EvaluatorPlanner { errors: vec![] }
+    }
+
     #[inline]
-    pub fn compile(&self, plan: &LogicalPlan<BindingsOp>) -> EvalPlan {
-        self.plan_eval(plan)
+    pub fn compile(&mut self, plan: &LogicalPlan<BindingsOp>) -> Result<EvalPlan, EvalErr> {
```
alancai98, line 65: Something I was less sure about was whether these logical plan to eval plan errors should be their own category (e.g. `CompileError` or `PlanningError`). Since the planning code currently resides in the `partiql-eval` crate, thought it was reasonable for now to use the same error type as the other eval errors. I'm open to creating a separate error type though.

partiql-eval/src/error.rs
```diff
@@ -0,0 +1,58 @@
...
+use thiserror::Error;
+
+/// All errors that occur during evaluation.
+#[derive(Debug)]
+pub struct EvalErr {
+    pub errors: Vec<EvaluationError>,
+}
+
+/// An error that can happen during evaluation.
+/// TODO: decide on splitting out logical to eval plan to separate error enum
```
jpschorr, line 15: Planning and Evaluation are distinct phases with differing types of errors. They should be disjoint error types.

partiql-eval/src/plan.rs
```diff
@@ -29,17 +30,49 @@ use crate::eval::expr::{
...
+        if !self.errors.is_empty() {
+            return Err(EvalErr {
+                errors: self.errors.clone(),
+            });
+        }
+        Ok(self.plan_eval(plan))
```
jpschorr, line 71: Planning is happening after the empty error check, but errors won't be populated until after plan.

```suggestion
        let plan = self.plan_eval(plan);
        if !self.errors.is_empty() {
            return Err(EvalErr {
                errors: std::mem::take(&mut self.errors),
            });
        }
        Ok(plan)
```

partiql-eval/src/plan.rs
dlurton: From the user's perspective, I think it's worthwhile to differentiate between *semantic* errors that happen while preparing a query for evaluation vs runtime errors that can only happen at evaluation time.  IMHO, it's ideal to have one `*Error` enum for every distinct step that the client application has to manually invoke.  i.e. if compilation, planning, and evaluation are three distinct steps, then one enum for each.

All such errors should have a line & column number, where applicable.

partiql-eval/src/plan.rs
```diff
@@ -334,7 +371,10 @@ impl EvaluatorPlanner {
                 match pattern {
                     Pattern::Like(logical::LikeMatch { pattern, escape }) => {
                         // TODO statically assert escape length
-                        assert!(escape.chars().count() <= 1);
+                        if escape.chars().count() > 1 {
+                            self.errors
+                                .push(EvaluationError::InvalidLikeEscape(escape.to_string()));
```
jpschorr, line 376: I'm not sure this is worth its own error variant. Perhaps just use `IllegalState` like coalesce

jpschorr (changes requested):

partiql-eval/src/plan.rs
jpschorr: Planning and evaluating are distinct phases. They currently live in the same crate today, but that is an implementation detail.

alancai98: Agree on the points to make planning and evaluating separate error categories. Updated the PR to have different errors:
- `PlanningError` and `PlanErr` (for multiple planning errors)
- `EvaluatorError` and `EvalErr` (for multiple evaluator errors) -- these might get adjusted further as I change the evaluation APIs to use `Result` types in an upcoming PR
---
> All such errors should have a line & column number, where applicable.

Yeah, we will get to that at some point for all applicable errors. I'll cut an issue to track, so we won't forget.

jpschorr (approved):

jpschorr:
> > Should evaluate support multiple errors?
> 
> Probably not since evaluation should stop at the first error. Multiple errors can be returned by the planning and semantic passes, tho.

Counter-argument: 

It is straightforward to join an array of errors for display/reporting. 

On the other hand, if the need arises for multiple messages, re-structuring the return types, de-tangling control-flow from error state accumulation, and pushing breaking API changes to all consumers is decidedly less straightforward.

PR #367, opened 2023-05-18
Change AST to logical plan function argument lowering to return Result

alancai98:
AST to logical plan lowering would panic when an incorrect number of arguments was provided to the function. Changed in this PR to return a `Result` type for this case.

Also changes the recovery behavior to add a `Missing` literal to the environment stack to prevent some downstream irrelevant errors.

### DIFF
--- a/partiql-logical-planner/src/call_defs.rs
+++ b/partiql-logical-planner/src/call_defs.rs
@@ -6,6 +6,7 @@ use partiql_value::Value;
 use std::collections::HashMap;
 use std::fmt::{Debug, Formatter};
 
+use crate::error::LowerError;
 use unicase::UniCase;
 
 #[derive(Debug, Eq, PartialEq)]
@@ -21,7 +22,11 @@ pub struct CallDef {
 }
 
 impl CallDef {
-    pub(crate) fn lookup(&self, args: &Vec<CallArgument>) -> ValueExpr {
+    pub(crate) fn lookup(
+        &self,
+        args: &Vec<CallArgument>,
+        name: String,
+    ) -> Result<ValueExpr, LowerError> {
         'overload: for overload in &self.overloads {
             let formals = &overload.input;
             if formals.len() != args.len() {
@@ -39,10 +44,9 @@ impl CallDef {
                 }
             }
 
-            return (overload.output)(actuals);
+            return Ok((overload.output)(actuals));
         }
-
-        todo!("mismatched formal/actual arguments to {}", &self.names[0])
+        Err(LowerError::InvalidNumberOfArguments(name))
     }
 }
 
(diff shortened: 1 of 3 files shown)


partiql-logical-planner/src/call_defs.rs
```diff
@@ -39,10 +44,9 @@ impl CallDef {
                 }
             }
 
-            return (overload.output)(actuals);
+            return Ok((overload.output)(actuals));
         }
-
-        todo!("mismatched formal/actual arguments to {}", &self.names[0])
```
alancai98, line 45: The line that would previously cause a panic if wrong number of function arguments provided.

partiql-logical-planner/src/lower.rs
```diff
@@ -916,10 +917,18 @@ impl<'ast> Visitor<'ast> for AstToLogical {
         let name = _call.func_name.value.to_lowercase();
 
         if let Some(call_def) = self.fnsym_tab.lookup(name.as_str()) {
-            self.push_vexpr(call_def.lookup(&env));
+            match call_def.lookup(&env, name) {
+                Ok(lookup) => self.push_vexpr(lookup),
+                Err(err) => {
+                    // Include as error but allow lowering to proceed for multiple error reporting
+                    self.errors.push(err);
+                    self.push_vexpr(ValueExpr::Lit(Box::new(Value::Missing)))
```
alancai98, line 925: Push some `ValueExpr` on the stack (`Missing` literal) otherwise some later code that expects something to be on the `vexpr` stack will give a somewhat irrelevant error.

partiql-logical-planner/src/lower.rs
```diff
@@ -916,10 +917,18 @@ impl<'ast> Visitor<'ast> for AstToLogical {
         let name = _call.func_name.value.to_lowercase();
 
         if let Some(call_def) = self.fnsym_tab.lookup(name.as_str()) {
```
jpschorr, line 919: Maybe: 

```rust
    fn exit_call(&mut self, call: &'ast Call) -> Traverse {
        // TODO better argument validation/error messaging
        let args = self.exit_call();
        let name = call.func_name.value.to_lowercase();

        let call_expr = self
            .fnsym_tab
            .lookup(name.as_ref())
            .ok_or_else(|| LowerError::UnsupportedFunction(name.clone()))
            .and_then(|c| c.lookup(&args, name)).;
        let expr = match call_expr {
            Ok(expr) => expr,
            Err(err) => {
                self.errors.push(e);
                ValueExpr::Lit(Box::new(Value::Missing)) // dummy expression to allow lowering to continue
            }
        };
        self.push_vexpr(expr);

        Traverse::Continue
    }
```

jpschorr (approved):

PR #380, opened 2023-05-31
Translate collection literals to Value when lowering

jpschorr:
- Tracks synthesized attributes during parse (currently only whether sub-tree is all literal values)
- Translates literal sub-trees to `Value` literals during lowering

---

### DIFF
--- a/partiql-logical-planner/src/lower.rs
+++ b/partiql-logical-planner/src/lower.rs
@@ -6,8 +6,8 @@ use partiql_ast::ast;
 use partiql_ast::ast::{
     Assignment, Bag, Between, BinOp, BinOpKind, Call, CallAgg, CallArg, CallArgNamed,
     CaseSensitivity, CreateIndex, CreateTable, Ddl, DdlOp, Delete, Dml, DmlOp, DropIndex,
-    DropTable, FromClause, FromLet, FromLetKind, GroupByExpr, GroupKey, GroupingStrategy, Insert,
-    InsertValue, Item, Join, JoinKind, JoinSpec, Like, List, Lit, NodeId, NullOrderingSpec,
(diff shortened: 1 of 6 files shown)


partiql-parser/src/parse/partiql.lalrpop
```diff
@@ -1251,6 +1315,24 @@ CommaTermPlus<T>: Vec<T> = {
...
+}
+
+// Comma as terminator (i.e. "<T>, <T>, <T>," where final comma is optional); at least 1 arg
+// This is a macro; see http://lalrpop.github.io/lalrpop/tutorial/006_macros.html
+SynthCommaTermPlus<T>: Synth<Vec<T>> = {
+    <v:(<T> ",")+> <e:T?> => match e {
+        None => v.into_iter().collect(),
+        Some(e) => v.into_iter().chain( std::iter::once(e) ).collect(),
+    }
+}
```
alancai98, line 1334: I'm still reading up on lalrpop macros but these don't seem to be used in the grammar?

partiql-logical-planner/src/lower.rs
alancai98, line 1: For some context, is this change along with #382, what enables the rendering of literals in the logical plan through the `partiql-rust-cli` (https://github.com/partiql/partiql-rust-cli/pull/21)?

alancai98 (approved):
Change looks good. Left some clarification questions about these changes to get more context. And left some comments about minor documentation updates.

partiql-logical-planner/src/lower.rs
jpschorr: It does enable that, yes. But, more generally, identifying constants prior to evaluation is necessary for several classes of optimizations.

PR #391, opened 2023-06-12
Add encoding/decoding to/from Ion Element

### LINKED ISSUE #351: Conversion from ion_rs::element::Element to partiql_value::Value.
In my partiql application I need to convert `ion_rs::element::Element` to a `partiql_value::Value`, but no such functionality is directly provided for this.

The closest thing today is `partiql_value::parse_ion`, but that function only accepts a `&str`, so I am left with converting the `Element` to a string and then calling `parse_ion`.  This works, but is sub-optimal for obvious reasons.

This API should *not* panic and should always return `Result::Err` when possible.

### LINKED ISSUE #357: Add API to convert `partiql_value::Value` to `ion_rs::element::Element`
This API serves the case for customers who need to work with the result of a query in memory instead of writing it to a stream.

Related to #356.

jpschorr:
Fixes #357
Fixes #351 

Add encoding/decoding to/from ion Element

---


extension/partiql-extension-ion/src/lib.rs
```diff
@@ -114,45 +202,94 @@ mod tests {
...
-        assert_partiql_encoded_ion("null", Value::Null);
-        assert_partiql_encoded_ion("$missing::null", Value::Missing);
+        /*
+        assert_partiql_encoded_ion(
+            "null",
+            ion_rs::element::Value::Null(IonType::Null),
+            Value::Null,
+        );
+
+         */
```
am357, line 248: Do we need to keep this?

jpschorr: Whoops. Yes, it shouldn't be commented out; fixing

am357 (approved):

PR #394, opened 2023-06-14
Use `partiql_ast::ast::AstTypeMap` for `LocationMap`

am357:
*Description of changes:*
Addressing the comment in PR #389, this PR re-uses the `AstTypeMap` for `LocationMap` which removes a dependency to `HashMap` as `AstTypeMap` uses `IndexMap`.

### DIFF
--- a/partiql-ast/Cargo.toml
+++ b/partiql-ast/Cargo.toml
@@ -20,7 +20,7 @@ path = "src/lib.rs"
 bench = false
 
 [dependencies]
-indexmap = { version = "1.9", default-features = false, features = ["serde"] }
+indexmap = { version = "1.9", default-features = false }
 rust_decimal = { version = "1.25.0", default-features = false, features = ["std"] }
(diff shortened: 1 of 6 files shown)


partiql-ast/Cargo.toml
```diff
@@ -20,7 +20,7 @@ path = "src/lib.rs"
 bench = false
 
 [dependencies]
-indexmap = "1.9"
+indexmap = { version = "1.9", default-features = false, features = ["serde"] }
```
jpschorr, line 23: You'll want to remove the unconditional `features = ["serde"]`, since you turn it on below conditionally via `serde = [... "indexmap/serde"]`

jpschorr (approved):

partiql-ast/Cargo.toml
am357: Whoops, let fix that.

PR #399, opened 2023-06-26
Type Simple SFW queries using current `LogicalPlan` as an `IR`

am357:
*Description of changes:*
Relates to: https://github.com/partiql/partiql-spec/issues/49

Adds the code for typing a query using the current `LogicalPlan` as an `IR`. This PR includes the changes to enable simple SFW queries as the following in both `Strict` and `Permissive` typing modes with `Open` and `Closed` schemas.

#### Closed schema with `Strict` typing mode.
```SQL
-- customers: <<{'id': INT, 'name': STRING, age: ANY}>>
SELECT customers.id, customers.name FROM customers;

-- Output schema: <<{'id': INT, 'name': STRING}>>
```

#### Open schema with `Strict` typing mode and `age` non-existent projection.
```SQL

-- customers: <<{'id': INT, 'name': STRING}>>
SELECT customers.id, customers.name, customers.age FROM customers;

-- Output schema: <<{'id': INT, 'name': STRING, 'age': ANY}>>
```

#### Closed Schema with `Permissive` typing mode and `age` non-existent projection.
```SQL
-- customers: <<{'id': INT, 'name': STRING>>
SELECT customers.id, customers.name, customers.age FROM customers;


-- Output schema: <<{'id': INT, 'name': STRING, 'age': NULL}>>
```

See: https://github.com/partiql/partiql-spec/discussions/64

#### Open Schema with `Strict` typing mode and `age` in nested attribute.
```SQL
-- customers: <<{'id': INT, 'name': STRING, 'details': {'age': INT}}>>
SELECT customers.id, customers.name, customers.details.age FROM customers;


-- Output schema: <<{'id': INT, 'name': STRING, 'age': INT}>>
```

#### Open Schema (`customers and details`) with `Strict` typing mode.
```SQL
-- customers: <<{'id': INT, 'name': STRING, 'details': {'age': INT}}>>
SELECT customers.id, customers.name, customers.details.age, customers.details.foo.bar FROM customers;


-- Output schema: <<{'id': INT, 'name': STRING, 'age': INT, 'bar': ANY}>>
```

#### Open Schema (`customers and details`)  with `Strict` typing mode and `age` in nested attribute with Path in `FROM` with alias.
```SQL
-- customers: <<{'id': INT, 'name': STRING, 'details': {'age': INT}}>>
SELECT d.age FROM customers.details AS d;


-- Output schema: <<{'age': INT}>>
```

See: https://github.com/partiql/partiql-spec/discussions/65

#### Closed Schema with `Strict` typing mode with `FROM` and `Projection` aliases.
```SQL
-- customers: <<{'id': INT, 'name': STRING, 'age': ANY}>>
SELECT c.id AS my_id, customers.name AS my_name FROM customers AS c;


-- Output schema: <<{'my_id': INT, 'my_name': STRING}>>
```

In addition:
- fixes some issues with the current `partiql_types` macros, changes the model for some of the types, and adds some helper methods to `PartiqlType`.
- moves constraints to set
- uses ` BTreeSet` for constraints and `AnyOf` to be able to support ordering on the these types; since we're not expecting large number of elements for these types the performance hit seems negligible. 

Upon merging this PR we need to add:
- Support for the rest of the Binding Operators (E.g., `Filter`, `OrderBy`).
- Support for operators (E.g., unary, binary).
- Support for functions.
- Related Conformance Tests.


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `partiql_ast_passes::static_typer` for type annotating the AST.
```
am357, line 22: I will remove this line, after we decide to remove the current `ast` pass or not.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,757 @@
...
+                                let type_ctx = ty_ctx![(&new_type_env, &new_ty)];
+                                self.type_env_stack.push(type_ctx);
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+            ValueExpr::Path(v, components) => {
+                self.type_vexpr(v, LookupOrder::LocalGlobal);
```
alancai98, line 312: If path is in a `FROM` source, lookup order could be globals first.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,757 @@
...
+                    Value::Boolean(_) => TypeKind::Bool,
+                    Value::String(_) => TypeKind::String,
+                    Value::Tuple(_) => TypeKind::Struct(StructType::new_any()),
+                    Value::List(_) => TypeKind::Array(ArrayType::new_any()),
+                    Value::Bag(_) => TypeKind::Bag(BagType::new_any()),
+                    _ => {
+                        self.errors.push(TypingError::NotYetImplemented(
+                            "Unsupported Literal".to_string(),
+                        ));
+                        todo!()
```
alancai98, line 353: What's the todo for?

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,757 @@
...
+    GlobalLocal,
+    LocalGlobal,
+}
+
+/// Represents a type environment context
+#[derive(Debug, Clone)]
+struct TypeEnvContext {
+    env: LocalTypeEnv,
+    /// Represents the type that is used for creating the `env` in the [TypeEnvContext]
+    derived_type: PartiqlType,
```
alancai98, line 71: Could you give more explanation for what `derived_type` is used for?

CHANGELOG.md
am357: I am hesitating to remove it just now, but can add a `TODO` to revisit; we should just make a final decision very soon.

partiql-types/src/lib.rs
```diff
@@ -156,76 +170,129 @@ pub enum TypeKind {
     Struct(StructType),
     Bag(BagType),
     Array(ArrayType),
+    // Serves as Bottom Type
+    Unknown,
```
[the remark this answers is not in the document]
am357: The original intention was to have a type as [Bottom Type](https://en.wikipedia.org/wiki/Bottom_type) as the comment suggests. I foresee this coming up in the operator typing in which we end up with an `Unknown` type for one of the operands so that we can call it an _ill-typed_ expression in order to be able to decide what to do next. I don't have a strong opinion for the name and can change.

partiql-types/src/lib.rs
```diff
@@ -156,76 +170,129 @@ pub enum TypeKind {
...
+            || matches!(*self, PartiqlType(TypeKind::Array(_)))
+    }
+
+    pub fn is_unordered_collection(&self) -> bool {
+        !self.is_ordered_collection()
+    }
+
+    pub fn is_ordered_collection(&self) -> bool {
+        // TODO Add Sexp when added
+        matches!(*self, PartiqlType(TypeKind::Array(_)))
```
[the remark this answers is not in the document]
am357: Yeah that is a good point. As there was a lot of ambiguity around this at the time of constructing the model (still is) I went with a designated constraint but I can see perhaps that is not a good idea. Let me revisit this.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,757 @@
...
+                                let mut new_type_env = LocalTypeEnv::new();
+                                if let TypeKind::Struct(s) = ty.kind() {
+                                    // for field in s.fields() {
+                                    //     let sym = SymbolPrimitive {
+                                    //         value: field.name().to_string(),
+                                    //         case: CaseSensitivity::CaseInsensitive,
+                                    //     };
+                                    //     new_type_env.insert(sym, field.ty().clone());
+                                    // }
+                                    to_bindings(&s).into_iter().for_each(|b| {
```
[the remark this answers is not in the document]
am357: Whoops! comments should not be there!

So this is a common model for creating bindings mostly as `projection schema`.

E.g.,
```rust
// Given: {'a': INT, 'b': DECIMAL} 

// Output: [{'a': INT}, {'b': DECIMAL}]
```

The binding being ordinal allows us to extracts the types with indices.

partiql-logical-planner/src/typer.rs
am357: This should not be here, will clean it up.

partiql-logical-planner/src/typer.rs
am357: Derived type is the type that is used for building the local typing environment in the stack.

alancai98 (commented):
(publishing some comments I had before the offline discussion)

partiql-logical-planner/src/typer.rs
am357: With the current model this is required esp. to know if we are in an open or closed schema during type-checking. E.g., when requiring to type-check `d` in `a.b.c.d` with `a` being `{'b': {'c': INT}}` where `b` is closed and `c` is open, we need to know somehow have access to the type that `d` _may_ have been resulted from to make a decision about its type i.e. being `any` in this case.

partiql-types/src/lib.rs
```diff
@@ -315,98 +353,151 @@ impl FromIterator<PartiqlType> for AnyOf {
...
 
-#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
-enum CollectionConstraint {
-    Ordered(bool),
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
+#[non_exhaustive]
+pub enum OrderedCollectionConstraint {
+    // Each value in an ordered collection is expected to be valid against the type in the corresponding position of the specified types list
+    OrderedElements(Vec<PartiqlType>),
 }
```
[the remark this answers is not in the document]
am357: Experimentally I am thinking of having at least two constraints for Arrays like the following:

```rust
    // Each value in an ordered collection is expected to be valid against the type in the corresponding position of the specified types list.
    // For example for an Array Type with OrderedElements Constraint [INT, DECIMAL] the following data are valid:
    // - [1, 1.3, 10], [102, 29.509, 11, 30], [-200, 415.789]
    // And the following data are invalid:
    // - ['hello', 20.5], [3.14, 223, true]
    OrderedElements(Vec<PartiqlType>),

    // Indicates whether an Array type is having fixed number of elements. For example the following are valid for an Array Type with FixedElements(2..2)
    // [1, 3], ['hello', 3303]
    FixedElements(u64..u64)
```

I defer having these for now and will had an empty `Enum` with `#[non_exhaustive]`

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,783 @@
...
+                    self.type_env_stack.push(ty_ctx![(
+                        &ty_env![(name.clone(), derived_type.clone())],
+                        ty
+                    )]);
+
+                    if !as_key.is_empty() {
+                        self.type_env_stack.push(ty_ctx![(
+                            &ty_env![(string_to_sym(as_key.as_str()), derived_type)],
+                            ty
+                        )]);
```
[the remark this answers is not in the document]
am357: As this works based off of the lowered plan, I want be defensive here and cater for `as_key` emptiness b/c its type `String` allows it to be empty today.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,757 @@
...
+                            let type_ctx = ty_ctx![(&ty_env![(sym, self.element_type(ty))], ty)];
+
+                            self.type_env_stack.push(type_ctx);
+                        } else {
+                            self.errors.push(TypingError::NotYetImplemented(
+                                "Local Lookup after unsuccessful Global lookup".to_string(),
+                            ));
+                        }
+                    }
+                    LookupOrder::LocalGlobal => {
```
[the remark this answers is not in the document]
am357: Fixed this in the new revision.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,783 @@
...
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+            ValueExpr::Path(v, components) => {
+                if let Some(op) = self.current_bop() {
+                    match op {
+                        BindingsOp::Scan(_) => self.type_vexpr(v, LookupOrder::GlobalLocal),
```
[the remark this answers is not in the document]
am357: This should be resolved by introducing `LookupOrder::Delegate` in the new revision.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,897 @@
...
+            Ok(self.output.clone().unwrap_or(undefined!()))
+        } else {
+            Err(TypeErr {
+                errors: self.errors.clone(),
+            })
+        }
+    }
+
+    fn type_bindings_op(&mut self, op: &BindingsOp) {
+        self.bop_stack.push(op.clone());
```
dlurton, line 174: Seems like cloning the entire `BindingsOp` could get expensive with large plans.

am357: `BindingOp` is not nested so I wouldn't  expect this to become expensive; good call-out though as made me to revisit the stack and see if we actually need it now. I removed it altogether.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,893 @@
...
+                "SELECT customers.id, customers.name, customers.details.bar FROM customers",
+                create_customer_schema(
+                    false,
+                    vec![
+                        StructField::new("id", int!()),
+                        StructField::new("name", str!()),
+                        StructField::new("details", details),
+                    ],
+                ),
+                vec![],
```
alancai98, line 787: In the error case + strict mode, should we still emit some type info? Like for instance, stating `id` resolved to `int`, `name` resolved to `string`, and `details.bar` resolved to `undefined`?

alancai98 (commented):
I think I'm starting to understand better the typing and need for derived type. Left some questions for some clarification mostly related to the current variable resolution and error messaging.

partiql-logical-planner/src/typer.rs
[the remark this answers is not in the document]
am357: The `LocalTypeEnv` is an `IndexedMap` which requires to have non-empty string keys. What would be the value of having a type in `LocalTypeEnv` with `""` as key? 

> Shouldn't the as_alias be defined by this point?

I believe we should not make this assumption.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,893 @@
...
+        }
+
+        match v {
+            ValueExpr::VarRef(binding_name, _) => {
+                let sym = binding_to_sym(binding_name);
+
+                match lookup_order {
+                    LookupOrder::GlobalLocal => {
+                        let ty = self.resolve_global_then_local(&sym);
+                        let ty = self.element_type(&ty);
```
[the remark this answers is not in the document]
am357: I don't think the unnesting should be part of the `Scan` mainly b/c of the possibility of having local lookups in the `Scan`. However perhaps it makes more sense to push this down to `resolve_global` since this is the only place that requires to unnest.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,893 @@
...
+                ),
+                vec![],
+            ),
+            vec![
+                TypingError::TypeCheck(
+                    "No Typing Information for SymbolPrimitive { value: \"details\", case: CaseInsensitive }".to_string()
+                ),
+                TypingError::TypeCheck(
+                    "No Typing Information for SymbolPrimitive { value: \"bar\", case: CaseInsensitive }".to_string()
+                )
```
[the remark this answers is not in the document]
am357: These errors get populated as we traverse through the plan and collect them. In this case `bar` is part of the path that we have traversed and faced an error for it. I still see it being useful to give give this error—this is something like an stack trace in my view. If we decide to fine tune the error we can certainly iterate on it.

partiql-logical-planner/src/typer.rs
am357: I don't believe we should, as it is ill-typed. However we can add the last type in the stack to the error message which should be the `output_schema`.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,893 @@
...
+                create_customer_schema(
+                    false,
+                    vec![
+                        StructField::new("id", int!()),
+                        StructField::new("name", str!()),
+                    ],
+                ),
+                vec![],
+            ),
+            vec![TypingError::IllegalState(format!(
```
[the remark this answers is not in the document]
am357: This could very well be a `TypeCheck` error in addition to this error. This is coming from line L465 as opposed to the other two L476 that are erroring out b/c the derive type is valid (struct) but schema is closed. I can make the error messages more clear.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,893 @@
...
+                                let type_ctx = ty_ctx![(&new_type_env, &ty)];
+                                self.type_env_stack.push(type_ctx);
+                            }
+                        }
+                    }
+                    LookupOrder::LocalGlobal => {
+                        let ty = self.resolve_local_then_global(&sym);
+
+                        if ty.is_undefined() {
+                            self.type_with_undefined(&sym);
```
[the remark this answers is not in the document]
am357: The difference is that the `resolve` methods return the types whereas `type_with_undefined` pushes the type in case of undefined to the stack. However I realized there is a simplification that can be made. Which is removing the `TypingMode` in the resolve methods. In the previous versions I was emitting an error in this code path and that one was removed but these where remaining. Thanks for pointing this out!

partiql-types/src/lib.rs
```diff
@@ -315,98 +357,148 @@ impl FromIterator<PartiqlType> for AnyOf {
...
 
-#[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Ord, PartialOrd)]
 #[allow(dead_code)]
 pub struct ArrayType {
-    pub element_type: Box<PartiqlType>,
-    constraints: Vec<CollectionConstraint>,
+    element_type: Box<PartiqlType>,
+    constraints: Vec<ArrayConstraint>,
```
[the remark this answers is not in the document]
am357: No strong opinions—removed.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,917 @@
...
+            ),
+            vec![StructField::new("age", int!())],
+        )
+        .expect("Type");
+
+        assert_query_typing(
+            TypingMode::Strict,
+            "SELECT c.id AS my_id, customers.name AS my_name FROM customers AS c",
+            create_customer_schema(
+                false,
```
alancai98, line 732: nit: test could have a comment since this test is different than the comment from L709. This test is w/ closed schema + strict typing mode

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,917 @@
...
+            ],
+        )
+        .expect("Type");
+    }
+
+    #[test]
+    fn simple_sfw_err() {
+        // Closed Schema with `Strict` typing mode and `age` non-existent projection.
+        let err1 = r#"No Typing Information for SymbolPrimitive { value: "age", case: CaseInsensitive } in closed Schema PartiqlType(Struct(StructType { constraints: {Open(false), Fields([StructField { name: "id", ty: PartiqlType(Int) }, StructField { name: "name", ty: PartiqlType(String) }])} }))"#;
+        let err2 = r#"Unexpected Typing Environment; expected typing environment with only one type but found 2 types"#;
```
alancai98, line 751: May be resolved in an upcoming PR, but I feel like this second error is somewhat confusing. I was mostly confused by the wording of the error when it states it found 2 types in the typing environment when 1 was expected. Shouldn't there only be the first error, which indicates that `age` cannot be found in the typing environment?

It's classified as an `IllegalState` error which I thought was not due to any user input error.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,917 @@
...
+            ),
+            vec![
+                TypingError::TypeCheck(err1.to_string()),
+                TypingError::TypeCheck(err2.to_string()),
+            ],
+            Some(bag![r#struct![BTreeSet::from([StructConstraint::Fields(
+                vec![
+                    StructField::new("id", int!()),
+                    StructField::new("name", str!()),
+                    StructField::new("bar", int!()),
```
alancai98, line 811: How is `bar` resolved to `int` here? Shouldn't this be `missing` since `customers.details` only has field `age`?

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,917 @@
...
+
+    fn type_with_undefined(&mut self, key: &SymbolPrimitive) {
+        if let TypingMode::Permissive = &self.typing_mode {
+            let type_ctx = ty_ctx![(&ty_env![(key.clone(), missing!())], &undefined!())];
+
+            self.type_env_stack.push(type_ctx);
+        }
+    }
+
+    fn get_singleton_type_from_env(&mut self) -> PartiqlType {
```
alancai98, line 521: Sorry think I missed this from a prior review. In what contexts should this `fn` be called? And why does it expect just one type in the typing environment (is it related to duplicates at all?)? Seems like it's currently called when extracting the final output schema from the typed plan and when creating the projection list schema.

alancai98 (commented):
I had a few more clarification questions. Otherwise should be ready to approve

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,917 @@
...
+            let mut new_type_env = LocalTypeEnv::new();
+            if let TypeKind::Struct(s) = ty.kind() {
+                to_bindings(s).into_iter().for_each(|b| {
+                    new_type_env.insert(b.0, b.1);
+                });
+
+                let type_ctx = ty_ctx![(&new_type_env, ty)];
+                self.type_env_stack.push(type_ctx);
+            } else {
+                // new_type_env.insert(sym, self.element_type(&ty));
```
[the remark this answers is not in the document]
am357: Oops! will remove.

partiql-logical-planner/src/typer.rs
am357: Let's discuss more as I still believe we need to have the list of all the errors we have encountered during typing (like an StackTrace). It's an `IllegalState` b/c we were expecting one type in the environment at some point during typing but found one. I can make changes in subsequent PRs if we agreed to.

partiql-logical-planner/src/typer.rs
am357: It's not related to duplicates, it is a helper function to extract one type out of the environment when we expect it. As far as I can tell today, this is mainly in projections, when we expect to infer one type from the project list items.

partiql-logical-planner/src/typer.rs
am357: This is in error state, so the schema is expected to be incorrect, however the more correct erroneous type for this is `UNDEFINED` which can translate to `MISSING` in permissive mode. Let me look further into this.

partiql-logical-planner/src/lower.rs
```diff
@@ -360,6 +361,19 @@ impl<'a> AstToLogical<'a> {
...
+                                                .resolve_type(name_ref.sym.value.as_ref())
+                                            {
+                                                let expr = ValueExpr::VarRef(
+                                                    var_binding.clone(),
+                                                    VarRefType::Global,
+                                                );
+                                                if !lookups.contains(&expr) {
+                                                    lookups.push(expr);
+                                                }
                                                 continue;
```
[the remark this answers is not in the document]
am357: Good catch! this has cropped up as part of the merge will remove.

partiql-logical-planner/src/typer.rs
```diff
@@ -0,0 +1,921 @@
...
+                    vec![
+                        StructField::new("id", int!()),
+                        StructField::new("name", str!()),
+                    ],
+                ),
+                vec![],
+            ),
+            vec![
+                TypingError::TypeCheck(err1.to_string()),
+                // TypingError::IllegalState(err2.to_string()),
```
alancai98, line 772: commented out code

```suggestion
                // TypingError::IllegalState(err2.to_string()),
```

alancai98 (approved):
Thanks for clarifying some things. Looks good and we can always adjust as we implement typing for other binding ops and expressions. Noticed one remaining typo.

alancai98 (approved):

alancai98:
Seems like a rustfmt check is still failing: https://github.com/partiql/partiql-lang-rust/actions/runs/6055593593/job/16434699935

jpschorr:
> `-- Output schema: <<{'id': INT, 'name': STRING, 'age': MISSING}>>`

Should this perhaps use `NULL` instead of `MISSING`? For values, `Missing` is not allowed as the value of a struct's field, as `Missing` implies the field is not present.

jpschorr:
> ```
> -- customers: <<{'id': INT, 'name': STRING, 'details': {'age': INT}}>>
> SELECT d.age FROM customers.details AS d;
> 
> 
> -- Output schema: <<{'age': INT}>>```

The `age` in the output is 'inferred' from final pathing in the `SELECT d.age`, which matches the type of column name inference that many RDBMSs implement. We'll have to think about the inference rules for, e.g., `SELECT d.age +1`

The PartiQL spec 6.3.1 is relevant here: https://partiql.org/assets/PartiQL-Specification.pdf#subsection.6.3

jpschorr (approved):

am357:
> > `-- Output schema: <<{'id': INT, 'name': STRING, 'age': MISSING}>>`
> 
> Should this perhaps use `NULL` instead of `MISSING`? For values, `Missing` is not allowed as the value of a struct's field, as `Missing` implies the field is not present.

That's a good call-out. Another option is to elide the attribute from the schema in permissive mode. As discussed offline will add a TODO in the code and spec. to get back to this later.

am357:
> SELECT d.age +1

am357:
> > `-- Output schema: <<{'id': INT, 'name': STRING, 'age': MISSING}>>`
> 
> Should this perhaps use `NULL` instead of `MISSING`? For values, `Missing` is not allowed as the value of a struct's field, as `Missing` implies the field is not present.

I created the following spec. discussion that aims to address this:
https://github.com/partiql/partiql-spec/discussions/64

am357:
> > ```
> > -- customers: <<{'id': INT, 'name': STRING, 'details': {'age': INT}}>>
> > SELECT d.age FROM customers.details AS d;
> > 
> > 
> > -- Output schema: <<{'age': INT}>>```
> > ```
> 
> The `age` in the output is 'inferred' from final pathing in the `SELECT d.age`, which matches the type of column name inference that many RDBMSs implement. We'll have to think about the inference rules for, e.g., `SELECT d.age +1`
> 
> The PartiQL spec 6.3.1 is relevant here: https://partiql.org/assets/PartiQL-Specification.pdf#subsection.6.3

I created the following GitHub discussion that aims to address this:
https://github.com/partiql/partiql-spec/discussions/65

am357:
> > > `-- Output schema: <<{'id': INT, 'name': STRING, 'age': MISSING}>>`
> > 
> > 
> > Should this perhaps use `NULL` instead of `MISSING`? For values, `Missing` is not allowed as the value of a struct's field, as `Missing` implies the field is not present.
> 
> That's a good call-out. Another option is to elide the attribute from the schema in permissive mode. As discussed offline will add a TODO in the code and spec. to get back to this later.

I went with the `NULL` for now pending https://github.com/partiql/partiql-spec/discussions/64 also added a TODO in the code.

Issue #410, opened 2023-07-17
Variable resolution and dynamic lookup evaluation incorrect or performing too many lookups


alancai98:
Variable resolution currently checks the local bindings and then searches the global context for each lookup:

https://github.com/partiql/partiql-lang-rust/blob/d1bdfb05bf5578dab60de91c5d8b526f8ca70b99/partiql-eval/src/eval/expr/mod.rs#L189

There is currently no way to perform the global lookup first and then the local binding lookup that is required when a variable or path in the `FROM` source is unqualified as specified in [section 10 of the spec](https://partiql.org/assets/PartiQL-Specification.pdf#section.10). This is apparent in a test case such as:

```
-- assuming a global table is defined as `foo`
SELECT a FROM foo, foo AS t2
```
The LHS of the cross join will be correct in pulling `foo` from the globals since the local bindings are empty. This will put `foo` in the local bindings. However the RHS of the cross join should check globals first (from section 10 of the spec) but will check the locals bindings first due to the current variable reference implementation.

---

Another consequence of the current modeling of variable references is that dynamic lookup will check the global context for every local variable lookup. For instance if we have some dynamic lookup of [path(local(a)), path(local(b)), path(local(c)), varref(global(d))], the actual lookup will do something like:
1. local lookup of a; if nothing, then global lookup of a
2. if nothing, local lookup of b; if nothing, then global lookup of b
3. if nothing, local lookup of c; if nothing, then global lookup of c
4. if nothing, global lookup of d
5. else Missing

when it should be something like:
1. local lookup of a
2. if nothing, local lookup of b
3. if nothing, local lookup of c
4. if nothing, global lookup of d
5. else Missing

PR #415, opened 2023-07-26
Change modeling of project lists in logical plan

alancai98:
Changes the modeling of partiql-logical's `project` to be a `Vec` rather than a `HashMap`. Previously, couldn't include multiple project aliases with the same name.

For some context, some conformance tests that will be fixed in an upcoming PR are partially failing because the select list included multiple expressions with the same alias. 


alancai98:
This was the conformance test that referenced multiple SELECT items that have the same alias: https://github.com/partiql/partiql-tests/blob/main/partiql-tests-data/eval/query/select/select-mysql.ion#L13776. This PR along with the fix for variable resolution in #410 will fix that test.

jpschorr (approved):

am357 (approved):

PR #411, opened 2023-07-18
Box Decimals in `Value` to assure `Value` fits in 16 bytes

jpschorr:
---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #413, opened 2023-07-18
Separate binding from evaluation of expression & implement strict eval

jpschorr:
First step towards:
- Putting built-in functions into the catalog so as to treat them like 'plug-in' functions
- `Strict` mode evalution

Still TODO in this PR:
- [X] Add doc comments where necessary
- [X] Update changelog

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #416, opened 2023-07-26
Variable resolution and dynamic lookup fixes

alancai98:
Closes #411.

Variable resolution and dynamic lookup would sometimes perform the incorrect lookup order or do too many lookups (e.g. do a global lookup on every local lookup). This PR
- modifies the variable resolution node in logical plan to distinguish between the local and global lookup
- fixes the group by data flow in name resolution


.github/workflows/ci_build_test.yml
```diff
@@ -6,8 +6,6 @@ env:
   CARGO_TEST_RESULT_NAME: cargo_test_results.json
   CONFORMANCE_REPORT_NAME: cts_report.json
   COMPARISON_REPORT_NAME: cts-comparison-report.md
-  RUST_TEST_TIME_UNIT: 150,5000
-  RUST_TEST_TIME_INTEGRATION: 150,5000
```
alancai98, line 10: A couple of the MYSQL tests were failing because they were taking a bit over these time thresholds.

partiql-conformance-tests/tests/mod.rs
```diff
@@ -126,16 +126,12 @@ pub(crate) fn fail_eval(statement: &str, mode: EvaluationMode, env: &Option<Test
...
-        Ok(plan) => {
-            let out = evaluate(plan, bindings);
-
-            assert!(
-                out.is_err(),
-                "When evaluating (mode = {mode:#?}) `{statement}`, expected `Err(_)`, but was `{out:#?}`"
-            );
-        }
-        Err(_) => {}
+    if let Ok(plan) = plan {
```
alancai98, line 129: Applying some clippy suggestions.

alancai98:
Conformance test regressions (https://github.com/partiql/partiql-lang-rust/pull/416#issuecomment-1652454328) are due to updated `partiql-tests` submodule, which corrected the behavior in some `MOD` builtin function tests. The behavior is corrected in #413.

partiql-eval/src/eval/expr/path.rs
```diff
@@ -96,13 +96,17 @@ impl EvalExpr for EvalDynamicLookup {
...
+    Global(BindingsName),
 }
 
 impl EvalExpr for EvalVarRef {
     fn evaluate<'a>(&'a self, bindings: &'a Tuple, ctx: &'a dyn EvalContext) -> Cow<'a, Value> {
-        let value = Bindings::get(bindings, &self.name).or_else(|| ctx.bindings().get(&self.name));
+        let value = match self {
+            EvalVarRef::Global(name) => ctx.bindings().get(name),
+            EvalVarRef::Local(name) => Bindings::get(bindings, name),
+        };
```
jpschorr, line 109: As the variable ref type is known at compile-time, it would be better to not make every lookup execute a conditional.

Perhaps something like:
```rust
/// Represents a local variable reference in a (sub)query, e.g. `b` in `SELECT t.b as a FROM T as t`.
#[derive(Debug, Clone)]
pub(crate) enum EvalVarRef {
    Local(BindingsName),
    Global(BindingsName),
}

impl BindEvalExpr for EvalVarRef {
    fn bind<const STRICT: bool>(
        &self,
        _: Vec<Box<dyn EvalExpr>>,
    ) -> Result<Box<dyn EvalExpr>, BindError> {
        Ok(match self {
            EvalVarRef::Global(name) => Box::new(EvalGlobalVarRef { name: name.clone() }),
            EvalVarRef::Local(name) => Box::new(EvalLocalVarRef { name: name.clone() }),
        })
    }
}

#[inline]
fn borrow_or_missing<'a>(value: Option<&'a Value>) -> Cow<'a, Value> {
    value.map_or_else(|| Cow::Owned(Value::Missing), Cow::Borrowed)
}

/// Represents a local variable reference in a (sub)query, e.g. `b` in `SELECT t.b as a FROM T as t`.
#[derive(Debug)]
pub(crate) struct EvalLocalVarRef {
    pub(crate) name: BindingsName,
}

impl EvalExpr for EvalLocalVarRef {
    fn evaluate<'a>(&'a self, bindings: &'a Tuple, _: &'a dyn EvalContext) -> Cow<'a, Value> {
        borrow_or_missing(Bindings::get(bindings, &self.name))
    }
}

/// Represents a global variable reference in a (sub)query, e.g. `T` in `SELECT t.b as a FROM T as t`.
#[derive(Debug)]
pub(crate) struct EvalGlobalVarRef {
    pub(crate) name: BindingsName,
}

impl EvalExpr for EvalGlobalVarRef {
    fn evaluate<'a>(&'a self, _: &'a Tuple, ctx: &'a dyn EvalContext) -> Cow<'a, Value> {
        borrow_or_missing(ctx.bindings().get(&self.name))
    }
}
```

and planning might look like:

```rust
            ValueExpr::VarRef(name, var_ref_type) => (
                "var ref",
                match var_ref_type {
                    VarRefType::Global => EvalVarRef::Global(name.clone()),
                    VarRefType::Local => EvalVarRef::Local(name.clone()),
                }
                .bind::<{ STRICT }>(vec![]),
            ),
```

alancai98: Thanks for the suggestion. Updated to your suggestion in https://github.com/partiql/partiql-lang-rust/pull/416/commits/1748f5bd668a2e0096f29a72a24698973b34a11a.

jpschorr (approved):

PR #418, opened 2023-08-02
Fixes to ORDER BY name resolution and null ordering bug

alancai98:
(PR targets `change-var-resolution` branch. Once https://github.com/partiql/partiql-lang-rust/pull/416 is merged in, this PR will target `main`)

Fixes some previous bugs to get more ORDER BY tests to pass:
- Some name resolution fixes. Previous dataflow issue remains related to select aliases referenced in order by
- Null/missing ordering for nested collections (kudos to @jpschorr on the initial factoring there)
  - For some context, the previous `Ord` trait implementation always assumed `NULLS FIRST`. So even though the ORDER BY allowed configuring `NULLS FIRST` or `NULLS LAST`, nested collections would not sort according to the specified sort order. E.g. comparing `[null]` with `[1]` with `NULLS LAST` should return `Ordering::Greater` but instead returned `Ordering::Less`


partiql-eval/src/eval/evaluable.rs
```diff
@@ -1043,6 +1051,7 @@ impl EvalOrderBy {
 impl Evaluable for EvalOrderBy {
     fn evaluate(&mut self, ctx: &dyn EvalContext) -> Value {
         let input_value = take_input!(self.input.take(), ctx);
+        println!("order by self: {:?}", self);
```
jpschorr, line 1054:
```suggestion
```

partiql-eval/src/eval/evaluable.rs
```diff
@@ -1009,30 +1009,38 @@ impl EvalOrderBy {
...
-                    },
-                    EvalOrderBySortSpec::DescNullsLast => r.as_ref().cmp(l.as_ref()),
+                    EvalOrderBySortSpec::AscNullsFirst => {
+                        NullSortedValue::<true, Value>(l.as_ref()).cmp(&NullSortedValue::<
+                            true,
+                            Value,
+                        >(
+                            r.as_ref()
+                        ))
+                    }
```
jpschorr, line 1019: might be more readable as:

```rust
let wrap = NullSortedValue::<true, Value>;
let (l,r) = (wrap(l), wrap(r));
l.cmp(&r)
```

jpschorr (approved):

partiql-eval/src/eval/evaluable.rs
alancai98: Thanks, applied the suggestion. Guess that original line was so long that the `rustfmt` made the code hard to read

partiql-eval/src/eval/evaluable.rs
```diff
@@ -1009,30 +1009,26 @@ impl EvalOrderBy {
...
+                    EvalOrderBySortSpec::DescNullsFirst => {
+                        let wrap = NullSortedValue::<false, Value>;
+                        let (l, r) = (wrap(l.as_ref()), wrap(r.as_ref()));
+                        r.cmp(&l)
+                    }
+                    EvalOrderBySortSpec::DescNullsLast => {
+                        let wrap = NullSortedValue::<true, Value>;
+                        let (l, r) = (wrap(l.as_ref()), wrap(r.as_ref()));
+                        r.cmp(&l)
```
alancai98, line 1030: These are doing the correct ordering with respect to null/missing values.

For `DescNullsFirst`, it'll do the full ordering with nulls assigned to be `Ordering::Greater` (i.e. the false). Since we do `r` compared to `l`, the result is flipped, so this will order in descending order with the null/missing values first.

For `DescNullsLast`, it'll do the full ordering with nulls assigned to be `Ordering::Less` (i.e. the true). And since we do `r` compared to `l`, the result is flipped. This will order in descending order with the null/missing values last.

I noticed these were not part of the conformance tests, so added some tests in this PR: https://github.com/partiql/partiql-tests/pull/106.

PR #435, opened 2023-10-23
Switch `.gitmodules` to use `https`

rajivr:
*Issue #, if available:*

*Description of changes:*

I would like to be able to use PartiQL crates directly from git. When I tried to do the following in my `Cargo.toml`

```toml
[dependencies]
partiql-parser = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-catalog = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-source-map = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-ast = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-logical-planner = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-logical = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-value = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
partiql-eval = { git = "https://github.com/partiql/partiql-lang-rust.git", rev = "fbaac0fe03"}
```

`cargo update` gives the following error.

```sh
$ cargo update
    Updating crates.io index
    Updating git repository `https://github.com/partiql/partiql-lang-rust.git`
error: failed to get `partiql-ast` as a dependency of package `fdb-rl v0.1.0 (/home/rajiv/zfs/foundation-db/dev-env/fdb-rl/fdb-rl)`

Caused by:
  failed to load source for dependency `partiql-ast`

Caused by:
  Unable to update https://github.com/partiql/partiql-lang-rust.git?rev=fbaac0fe03

Caused by:
  failed to update submodule `partiql-conformance-tests/partiql-tests`

Caused by:
  failed to parse url for submodule `partiql-conformance-tests/partiql-tests`: `git@github.com:partiql/partiql-tests.git`

Caused by:
  relative URL without a base
```

By switching the git protocol to use `https` this error can be resolved. [Here](https://github.com/rajivr/fdb-rl-wip/blob/9422d4ed9687893160891a2440f09411c0fff231/fdb-rl/Cargo.toml#L24-L31) is a working version using a repository where this commit is available.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #436, opened 2023-10-24
Fix clippy warnings/errors

alancai98:
Seems like GH Actions CI has been failing since a recent clippy change turned a warning into an error. PR also fixes a different clippy warning.

This should fix the GitHub Actions failure seen in https://github.com/partiql/partiql-lang-rust/pull/435.

### DIFF
--- a/partiql-value/src/list.rs
+++ b/partiql-value/src/list.rs
@@ -183,22 +183,7 @@ impl PartialEq for List {
 
 impl PartialOrd for List {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        let mut l = self.0.iter();
-        let mut r = other.0.iter();
-
-        loop {
-            match (l.next(), r.next()) {
-                (None, None) => return Some(Ordering::Equal),
-                (Some(_), None) => return Some(Ordering::Greater),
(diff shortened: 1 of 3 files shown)


partiql-value/src/list.rs
```diff
@@ -183,22 +183,7 @@ impl PartialEq for List {
...
-                (Some(_), None) => return Some(Ordering::Greater),
-                (None, Some(_)) => return Some(Ordering::Less),
-                (Some(lv), Some(rv)) => match lv.partial_cmp(rv) {
-                    None => return None,
-                    Some(Ordering::Less) => return Some(Ordering::Less),
-                    Some(Ordering::Greater) => return Some(Ordering::Greater),
-                    Some(Ordering::Equal) => continue,
-                },
-            }
-        }
```
alancai98, line 201: Having both `Ord` and `PartialOrd` implemented (with a non-default implementation) results in a clippy error: https://rust-lang.github.io/rust-clippy/master/index.html#/incorrect_partial_ord_impl_on_ord_type

am357 (approved):
Nice dive deep! Thank you for the fix!

PR #437, opened 2023-10-26
partiql-logical: add method to visualize logical plan

rajivr:
*Issue #, if available:*

*Description of changes:*

Adds a simple method to create a dot graph of the logical plan. It is modeled after a similar method in `partiql_eval::eval::EvalPlan` (physical plan).

### DIFF
--- a/partiql-logical/Cargo.toml
+++ b/partiql-logical/Cargo.toml
@@ -22,6 +22,7 @@ bench = false
 
 [dependencies]
 partiql-value = { path = "../partiql-value", version = "0.5.*" }
+petgraph = "0.6.*"
 ordered-float = "3.*"
 itertools = "0.10.*"
 unicase = "2.6"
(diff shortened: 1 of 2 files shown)


am357 (commented):
@rajivr for context, we have made an intentional decision to not to be dependent to `petgraph` to keep the `partiql-logical` lean. 

There is already a feature in `partiql-rust-cli` for this; would that work for your use-case?
https://github.com/partiql/partiql-rust-cli/pull/21

alancai98:
Another option if you want the petgraph visualization is to define the visualization method in your own code. It appears that the methods you are using are already public methods/functions.

Regarding your other PR #418 adding the partiql-rust-cli, the PartiQL Rust team is generally ok w/ moving over the visualization code to partiql-lang-rust. If you're okay w/ using what you had included in #438, mind if I close this PR?

rajivr:
Thanks for the feedback @alancai98 . I'll work on the changes requested in #438 later this week or early next week.

PR #438, opened 2023-10-31
partiql-visualize-dot: add

rajivr:
Extract code from `partiql-rust-cli/src/visualize`

*Issue #, if available:*

*Description of changes:*

As mentioned [here](https://github.com/partiql/partiql-lang-rust/pull/437#issuecomment-1785233611), I have tried to extract code from `partiql-rust-cli/src/visualize`, so we can easily use it in our tests and other places.

One way of using it might be as follows.

In `Cargo.toml`

```
[dev-dependencies]
# ...
partiql-visualize-dot = { path = "../partiql-visualize-dot", version = "0.5.*" }
```

And then within a test we can do

```rust
use partiql_visualize_dot::{PlanToDot, ToDotGraph};

#[test]
fn test {
    // create `logical` which is a value of `LogicalPlan<BindingsOp>`
    let plan_to_dot = PlanToDot::default();
    println!("{}", plan_to_dot.to_graph(&logical));
}
```


rajivr:
To add further context, here is how I was able to visualize [`tests::custom_ion_scan_text`](https://github.com/partiql/partiql-lang-rust/blob/57ce9b80172923173a09d4c3e10565f7ebf796fd/extension/partiql-extension-ion-functions/src/lib.rs#L240-L243) which uses a custom catalog.

graph-ast

graph-logical

graph-physical

Cargo.toml
```diff
@@ -25,7 +25,7 @@ members = [
   "partiql-rewriter",
   "partiql-types",
   "partiql-value",
-
+  "partiql-visualize-dot",
```
alancai98, line 28: Could you put this new crate under extensions? Similar to what was done for `partiql-extension-ion` and `partiql-extension-ion-functions`?

partiql-visualize-dot/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+version.workspace = true
+edition.workspace = true
+
+[lib]
+bench = false
+
+[dependencies]
+partiql-ast = { path = "../partiql-ast", version = "0.5.*" }
+partiql-logical = { path = "../partiql-logical", version = "0.5.*" }
+
```
alancai98, line 26: Could you put a `features` section for dot visualization and call it something like `visualize-dot`?

partiql-visualize-dot/src/ast_to_dot.rs
alancai98, line 1: Question for these `src` files: are these essentially the same as the ones in `partiql-rust-cli`?

alancai98 (commented):
Thanks for the addition! The PartiQL Rust team is ok w/ moving the partiql-rust-cli dot visualization over. Since this visualization code was part of the experimental partiql-rust-cli, we would prefer if it was under the `extension` directory and also be behind a `features` flag.

Could you also add
- An entry to the CHANGELOG for this addition
- A brief README for this new crate, containing some of the verbiage from the partiql-rust-cli about this crate being experimental, subject to change, etc.

Once those are added and the other comments are addressed, we should be good to approve your PR.

rajivr:
> Could you also add
> 
>     * An entry to the CHANGELOG for this addition
> 
>     * A brief README for this new crate, containing some of the verbiage from the partiql-rust-cli about this crate being experimental, subject to change, etc.
> 

`CHANGELOG` and `README` has been added.

Cargo.toml
rajivr: I've moved the crate to `extension` and renamed it to `partiql-extension-visualize`

partiql-visualize-dot/Cargo.toml
rajivr: I am not entirely clear how adding `visualize-dot` to `features` would be helpful. Please allow me to explain and let me know your thoughts.

From what I can make out, the code in `partiql-rust-cli/src/visualize` creates an output in `dot` format, using `dot-writer` crate.

It then uses [other](https://github.com/partiql/partiql-rust-cli/blob/74f10e7b854bc3b0afab0cc8ab30feffc89a3582/Cargo.toml#L76-L82) crates to support rendering in different formats, such as png, svg, etc.,

In the current PR, I've only extracted the code to just create `dot` output. I am guessing when we add support for png, svg output, those can be behind feature flags in order to minimize the dependency graph?

alancai98: Sorry perhaps I should have added more detail in my initial comment. Since the dot visualization of the logical plan is still experimental / a PoC, we'd like to put the new functions and structures behind a feature similar to what was done for `partiql-ast`'s `serde` feature:
- `partiql-ast` feature for `serde`: https://github.com/partiql/partiql-lang-rust/blob/b9a8885e1f8f8a8a0876b820fd9f15ed87a2e448/partiql-ast/Cargo.toml#L31-L36
- example of using a [`cfg` expression](https://doc.rust-lang.org/reference/conditional-compilation.html) to conditionally include code when `serde` feature is enabled: https://github.com/partiql/partiql-lang-rust/blob/b9a8885e1f8f8a8a0876b820fd9f15ed87a2e448/partiql-ast/src/ast.rs#L24

With how we had defined the `serde` feature, the above code would only be compiled and built when the `serde` feature is specified (or `--all-features`). We would like something similar for `partiql-extension-visualize`.

alancai98 (commented):
Thanks for applying the comments. Changes since first review look good (and thanks for merging/rebasing off of `main`!). I left a followup to the `features` comment from the prior review. Once that's resolved, think this is ready to merge.

partiql-visualize-dot/Cargo.toml
rajivr: I have added `visualize-dot` feature. Please review.

alancai98 (approved):
Looks good. Thanks for the PR!

alancai98:
Ignoring coverage CI failure since this is experimental/PoC.

PR #446, opened 2024-03-06
Add session context to evaluation

### LINKED ISSUE #388: Feature Request: EvaluationSession.context equivalent
Please provide some way of allowing custom functions to access an application-specified context, such as a current transaction object.   However implemented, this functionality would be equivalent to [`EvaluationSession`](https://github.com/partiql/partiql-lang-kotlin/blob/e0eea801892f9cb96342f0918d8c9273ca3a11c4/partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluationSession.kt#LL33C14-L33C14), which is passed to [`ExprFunction.call`](https://github.com/partiql/partiql-lang-kotlin/blob/e0eea801892f9cb96342f0918d8c9273ca3a11c4/partiql-lang/src/main/kotlin/org/partiql/lang/eval/ExprFunction.kt#L50).  With `EvaluationSession`, [`context`](https://github.com/partiql/partiql-lang-kotlin/blob/e0eea801892f9cb96342f0918d8c9273ca3a11c4/partiql-lang/src/main/kotlin/org/partiql/lang/eval/EvaluationSession.kt#LL33C14-L33C14) variables may be assigned before query evaluation and accessed by the `ExprFunction` or `bexpr` implementations.  [Example](https://github.com/partiql/partiql-lang-kotlin/blob/e0eea801892f9cb96342f0918d8c9273ca3a11c4/partiql-lang/src/test/kotlin/org/partiql/lang/compiler/memorydb/operators/GetByKeyProjectRelationalOperatorFactory.kt#LL75C38-L75C38).  We need basically the same thing for the Rust implementation.

Without this, we have to use thread-local storage for this, which is sub-optimal for obvious reasons.

jpschorr:
Fixes #388 

Adds both a system-level and a user-level session context to the evaluation context. Also plumbs the context through to `BaseTableExpr` used in `partiql_catalog::Extension`s.

There is a lot of fiddly lifetime management and testing changes, as well as some slight trait refactors.

It's probably easiest to start with 
- the new integration test in `partiql/tests/user_context.rs` to see what kinds of things are possible,
- `partiql-catalog/src/context.rs` for the new context traits, and
- `partiql-eval/src/eval/mod.rs` for how it's used in evaluation.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - Adds quotes to the attributes of PartiQL tuple's debug output so it can be read and transformed using Kotlin `partiql-cli`
-- Changes the interface to `EvalPlan` to accept an `EvalContext`
+- [breaking] Changes the interface to `EvalPlan` to accept an `EvalContext`
 
 ### Added
 - Add `partiql-extension-visualize` for visualizing AST and logical plan
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - Adds quotes to the attributes of PartiQL tuple's debug output so it can be read and transformed using Kotlin `partiql-cli`
+- Changes the interface to `EvalPlan` to accept an `EvalContext`
 
 ### Added
 - Add `partiql-extension-visualize` for visualizing AST and logical plan
+- Add a `SessionContext` containing both a system-level and a user-level context object usable by expression evaluation 
 
 ### Fixed
 - Fixed `ORDER BY`'s ability to see into projection aliases
--- a/partiql-eval/src/eval/mod.rs
+++ b/partiql-eval/src/eval/mod.rs
@@ -1,6 +1,9 @@
 use itertools::Itertools;
+use std::any::Any;
 use std::cell::RefCell;
+use std::collections::HashMap;
 
+use delegate::delegate;
 use std::fmt::Debug;
 
 use petgraph::algo::toposort;
@@ -8,7 +11,7 @@ use petgraph::dot::Dot;
 use petgraph::prelude::StableGraph;
 use petgraph::{Directed, Outgoing};
 
-use partiql_value::Value;
+use partiql_value::{DateTime, Value};
 
 use crate::env::basic::MapBindings;
 use crate::env::Bindings;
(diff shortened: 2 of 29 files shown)


partiql-eval/src/eval/mod.rs
```diff
@@ -140,33 +144,52 @@ pub struct Evaluated {
...
+impl<'a> SessionContext<'a> for BasicContext<'a> {
     fn bindings(&self) -> &dyn Bindings<Value> {
         &self.bindings
     }
 
+    fn system_context(&self) -> &SystemContext {
+        &self.sys
+    }
+
+    fn user_context(&self, name: &str) -> Option<&(dyn Any)> {
```
sadderchris, line 184: Is there _any_ way this API can output a generic instead of a `Option<&dyn Any>`?  I'd really like to implement these traits myself and just return a type that I choose, instead of going through a `dyn Any`.

partiql-eval/src/eval/mod.rs
```diff
@@ -140,33 +144,52 @@ pub struct Evaluated {
...
-#[derive(Default, Debug)]
-pub struct BasicContext {
-    bindings: MapBindings<Value>,
-    errors: RefCell<Vec<EvaluationError>>,
+#[derive(Debug)]
+pub struct BasicContext<'a> {
+    pub bindings: MapBindings<Value>,
+
+    pub sys: SystemContext,
+    pub user: HashMap<String, &'a (dyn Any)>, // TODO: Unicase the keys?
```
sadderchris, line 160: We'll probably define our own context type, but for the most flexibility, this should be `&'a (dyn Any + 'a)` to support types that contain lifetimes themselves (e.g. `&'a Foo<'a>` - else you can only store something that is `&'a Foo<'static>`).

...This doesn't seem to just work however.  I think the `user_context` method on the SessionContext trait needs to be updated as well, there might be some other hoops that need to be jumped through as well.

partiql-eval/src/eval/mod.rs
```diff
@@ -62,8 +65,9 @@ impl EvalPlan {
 
     /// Executes the plan while mutating its state by changing the inputs and outputs of plan
     /// operators.
-    pub fn execute_mut(&mut self, bindings: MapBindings<Value>) -> Result<Evaluated, EvalErr> {
-        let ctx: Box<dyn EvalContext> = Box::new(BasicContext::new(bindings));
+    pub fn execute_mut<'c>(&mut self, ctx: &'c dyn EvalContext<'c>) -> Result<Evaluated, EvalErr> {
+        //let ctx: Box<dyn EvalContext> = Box::new(BasicContext::new(bindings));
```
am357, line 69:
```suggestion
```

partiql-eval/src/eval/mod.rs
am357, line 160: > // TODO: Unicase the keys?
I think so

jpschorr: `Any` is defined as 
```rust
pub trait Any: 'static {...}
```
(See https://doc.rust-lang.org/std/any/trait.Any.html)

So you can't add a `+ 'a`

partiql-eval/src/eval/mod.rs
jpschorr: If there is, I don't see how.

Making it a generic parameter of the `SessionContext` would mean that any and all extensions that are loaded would need to be able to deal with that single type.

CHANGELOG.md
```diff
@@ -8,9 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - Adds quotes to the attributes of PartiQL tuple's debug output so it can be read and transformed using Kotlin `partiql-cli`
+- Changes the interface to `EvalPlan` to accept an `EvalContext`
```
am357, line 11:
```suggestion
- [breaking] Changes the interface to `EvalPlan` to accept an `EvalContext`
```

am357 (approved):
Approved with one suggestion

partiql-eval/src/eval/mod.rs
sadderchris: Ok, this won't work for us then.  Do you have some time to talk about this in person tomorrow?

PR #448, opened 2024-03-07
Add argument lifetime to `BaseTableExpr`'s `evaluate`

jpschorr:
---

### DIFF
--- a/extension/partiql-extension-ion-functions/src/lib.rs
+++ b/extension/partiql-extension-ion-functions/src/lib.rs
@@ -99,11 +99,14 @@ impl BaseTableFunctionInfo for ReadIonFunction {
 pub(crate) struct EvalFnReadIon {}
 
 impl BaseTableExpr for EvalFnReadIon {
-    fn evaluate<'c>(
+    fn evaluate<'a, 'c>(
         &self,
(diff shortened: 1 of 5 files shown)


am357 (approved):

extension/partiql-extension-ion-functions/src/lib.rs
```diff
@@ -99,11 +99,14 @@ impl BaseTableFunctionInfo for ReadIonFunction {
 pub(crate) struct EvalFnReadIon {}
 
 impl BaseTableExpr for EvalFnReadIon {
-    fn evaluate<'c>(
+    fn evaluate<'a, 'c>(
         &self,
-        args: &[Cow<Value>],
+        args: &'a [Cow<Value>],
```
sadderchris, line 104: There's also an implicit lifetime on the `Cow` as well (the type is actually `Cow<'_, Value>`.  Can you check this doesn't also need to be referenced?

As a nit: you may want to throw `#![warn(rust_2018_idioms)]` or even `#![deny(rust_2018_idioms)]` in you crates' lib.rs to enforce that lifetimes always get written out (unless you feel really strongly about the added verbosity - IMO the benefits of this lint outweigh the downsides).

PR #454, opened 2024-03-15
 Fix subquery global binding lookup and error propagation

jpschorr:
1) Add subquery in `FROM` clause tests reference bindings
2) Fix subquery global binding lookup
3) Fix subquery error propagation

---

### DIFF
--- a/partiql/src/subquery_tests.rs
+++ b/partiql/src/subquery_tests.rs
@@ -4,15 +4,15 @@
 
 #[cfg(test)]
 mod tests {
-    use std::any::Any;
-    use partiql_catalog::{Catalog, PartiqlCatalog};
     use partiql_catalog::context::SystemContext;
+    use partiql_catalog::{Catalog, PartiqlCatalog};
     use partiql_eval::env::basic::MapBindings;
     use partiql_eval::error::EvalErr;
     use partiql_eval::eval::BasicContext;
     use partiql_eval::plan::EvaluationMode;
     use partiql_logical::{LogicalPlan, ProjectValue, VarRefType};
-    use partiql_value::{Bag, bag, BindingsName, DateTime, tuple, Value};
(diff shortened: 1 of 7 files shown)


partiql/src/subquery_tests.rs
```diff
@@ -0,0 +1,152 @@
...
+        let bindings = MapBindings::default();
+        let res = evaluate(&catalog, plan, bindings, &[]).expect("should eval correctly");
+        dbg!(&res);
+        assert!(res != Value::Missing);
+        assert_eq!(res, Value::from(bag![tuple![("a", "b")]]))
+    }
+
+    #[test]
+    fn globals_in_subqueries() {
+        let mut sub_query = partiql_logical::LogicalPlan::new();
```
am357, line 96:
```suggestion
        //  `foo` is defined in global environment as `<<{'a': 'b'}>>`
        //  `SELECT VALUE _1 FROM (SELECT VALUE foo FROM foo AS foo) AS _1;`
        let mut sub_query = partiql_logical::LogicalPlan::new();
```

partiql/src/subquery_tests.rs
```diff
@@ -0,0 +1,152 @@
...
+        let mut ctx = BasicContext::new(bindings, sys);
+        for (k, v) in ctx_vals {
+            ctx.user.insert(k.as_str().into(), *v);
+        }
+        plan.execute_mut(&ctx).map(|out| out.result)
+    }
+
+    #[test]
+    fn locals_in_subqueries() {
+        let mut sub_query = partiql_logical::LogicalPlan::new();
```
am357, line 42:
```suggestion
        //  `SELECT VALUE _1 from (SELECT VALUE foo from <<{'a': 'b'}>> AS foo) AS _1;`
        let mut sub_query = partiql_logical::LogicalPlan::new();
```

am357 (approved):

PR #477, opened 2024-07-11
Replace `BTreeSet` with `IndexSet` in type to order more intuitively

jpschorr:
This ensures that output order for things like basic DDL match specification order

---


extension/partiql-extension-ddl/tests/ddl-tests.rs
```diff
@@ -1,27 +1,27 @@
...
-    let shape = bag![r#struct![BTreeSet::from([
+    let shape = bag![r#struct![IndexSet::from([
         StructConstraint::Fields(fields),
         StructConstraint::Open(false)
     ])]];
 
     let ddl_compact = PartiqlBasicDdlEncoder::new(DdlFormat::Compact);
     let actual = ddl_compact.ddl(&shape).expect("ddl_output");
-    let expected = r#""address" VARCHAR NOT NULL,"id" INT,"name" VARCHAR,"details" OPTIONAL STRUCT<"age": INT>"#;
+    let expected = r#""id" INT,"name" VARCHAR,"address" VARCHAR NOT NULL,"details" OPTIONAL STRUCT<"age": INT>"#;
```
jpschorr, line 24: Note that it is the same set of fields and their types, but the order matches the order in which it was created.

E.g., 
```
    let fields = [
        StructField::new("id", int!()),
        StructField::new("name", str!()),
        StructField::new("address", PartiqlShape::new_non_nullable(Static::String)),
        StructField::new_optional("details", details.clone()),
    ]
```

am357 (approved):

PR #480, opened 2024-07-25
Add Parsing of `EXCLUDE`

jpschorr:
For the `EXCLUDE` clause, implements:
- AST nodes 
- parsing,
- pretty-printing
 

Cf. https://github.com/partiql/partiql-docs/pull/51

---


partiql-ast/src/ast.rs
```diff
@@ -690,6 +697,27 @@ pub struct PathExpr {
...
 }
 
+#[derive(Visit, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct ExcludePath {
+    pub root: AstNode<VarRef>,
+    pub steps: Vec<ExcludePathStep>,
+}
+
+/// A "step" within a path expression; that is the components of the expression following the root.
```
alancai98, line 707: Should clarify that this is a "step" within an `EXCLUDE` path

```suggestion
/// A "step" within an exclude path; that is the components of the exclude path following the root.
```

partiql-ast/src/ast.rs
```diff
@@ -690,6 +697,27 @@ pub struct PathExpr {
...
+#[derive(Visit, Clone, Debug, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ExcludePathStep {
```
alancai98, line 710: I like that the naming of the enum variants are the same as `PathStep`

alancai98 (approved):

alancai98 (approved):

am357 (approved):

PR #484, opened 2024-08-08
Move NodeId Generation to a separate crate

am357:
*Description of changes:*

Moves the `NodeId` and its Id generation to a new crate, `partiql-common` so that it can get consumed by multiple crates.


partiql-ast/src/ast.rs
```diff
@@ -17,13 +17,10 @@ use std::fmt;
 use serde::{Deserialize, Serialize};
 
 use partiql_ast_macros::Visit;
+use partiql_core::node::NodeId;
 
 pub type AstTypeMap<T> = IndexMap<NodeId, T>;
```
jpschorr, line 22: Should this move and rename as well?

jpschorr: If it does, then `partiql-source-map` could stop importing `ast`.

jpschorr (commented):
I wonder if there is a better name than `partiql-core`?

Perhaps we could move this type and `partiql-source-map` somewhere...

partiql-ast/src/ast.rs
am357: Makes sense; could be `NodeTypeMap` or something similar.

am357:
> I wonder if there is a better name than `partiql-core`?
Not sure TBH—happy to change if you have a suggestion.

> Perhaps we could move this type and `partiql-source-map` somewhere...
I could see `partiql-source-map` going under: `<partiql-core>/src/syntax-source/...`. Thoughts?

partiql-ast/src/ast.rs
am357: > Makes sense; could be `NodeTypeMap` or something similar.

Just realized I have already moved but haven't removed this one:
`pub type NodeMap<T> = IndexMap<NodeId, T>;`

jpschorr:
> > I wonder if there is a better name than `partiql-core`?
> 
> Not sure TBH—happy to change if you have a suggestion.

We talked out-of-band and thing `partiql-common` may be less ambiguous.


> > Perhaps we could move this type and `partiql-source-map` somewhere...
> 
> I could see `partiql-source-map` going under: `<partiql-core>/src/syntax-source/...`. Thoughts?

Sure

jpschorr (approved):

PR #485, opened 2024-08-12
Adds `PartiqlShapeBuilder` with `NodeId` generation for the `StaticType`

am357:
*Description of changes:*

This PR
- adds `NodeId` to `StaticType`; this is to be able to use the `id` as a reference to add additional data to the types out of band.
- makes `AutoNodeIdGenerator` thread-safe
- adds `PartiqlShapeBuilder` and moves some `PartiqlShape` APIs to it; this is to be able to generate unique `NodeId`s for a `PartiqlShape` that includes static types that themselves can include other static types.
- adds a static thread safe `shape_builder` function that provides a convenient way for using `PartiqlShapeBuilder` for creating new shapes.
- prepends existing type macros with `type` such as `type_int!` to make macro names more friendly.
- removes `const` PartiQL types under `partiql-types` in favor of `PartiqlShapeBuilder`.

**_The majority of the diffs are related to the macro renames or replacement with the previous `const` types. The main change is in `partiql-types/src/lib.rs` file._**

### DIFF
--- a/partiql-types/Cargo.toml
+++ b/partiql-types/Cargo.toml
@@ -21,7 +21,7 @@ edition.workspace = true
 bench = false
 
 [dependencies]
-
+partiql-common = { path = "../partiql-common", version = "0.10.*"}
 ordered-float = "3.*"
 itertools = "0.10.*"
 unicase = "2.6"
@@ -32,6 +32,7 @@ thiserror = "1.*"
 indexmap = "2.2"
 
 derivative = "2.2"
+lazy_static = "1.5.0"
 
 [dev-dependencies]
 criterion = "0.4"
--- a/partiql-types/Cargo.toml
+++ b/partiql-types/Cargo.toml
@@ -32,7 +32,6 @@ thiserror = "1.*"
 indexmap = "2.2"
 
 derivative = "2.2"
-lazy_static = "1.5.0"
 
 [dev-dependencies]
 criterion = "0.4"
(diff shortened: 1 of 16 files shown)


jpschorr (commented):
`PartiqlShapeBuilder.any_of`:
```
    pub fn any_of<I>(&self, types: I) -> PartiqlShape
    where
        I: IntoIterator<Item = PartiqlShape>,
    {
        let any_of = AnyOf::from_iter(types);
        match any_of.types.len() {
            0 => type_dynamic!(),
            1 => {
                let AnyOf { types } = any_of;
                types.into_iter().next().unwrap()
            }
            // TODO figure out what does it mean for a Union to be nullable or not
            _ => PartiqlShape::AnyOf(any_of),
        }
    }
```

The `AnyOf::from_iter(types)` uses an `IndexSet` internally to deduplicate types, thus the match on `any_of.types.len()` could "flatten" `AnyOf`s that had duplicates. With the addition of IDs, this deduplication no longer happens...

partiql-types/Cargo.toml
```diff
@@ -32,6 +32,7 @@ thiserror = "1.*"
 indexmap = "2.2"
 
 derivative = "2.2"
+lazy_static = "1.5.0"
```
jpschorr, line 35: lazy_static usage is currently explicitly disallowed, as `OnceCell` is preferred: https://github.com/partiql/partiql-lang-rust/blob/684fff3092311722c15b5b86506c375a49f78d1a/deny.toml#L126-L128

am357: Yeah this one pulled in by accident, thanks for the fix!

jpschorr, line 35:
```suggestion
```

jpschorr (approved):

PR #519, opened 2024-11-18
Change modeling of boxed ion literals to be lazy until evaluator.

jpschorr:
Changes the logical plan to have a distinct `Lit` type to hold literals instead of embedded `Value`

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #517, opened 2024-11-16
Change modeling of Literals in the AST remove ambiguity

jpschorr:
Change parsing and AST-modeling of literals to not share AST structures with non-scalar expressions.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #507, opened 2024-10-11
Change Lexing/Parsing of embedded docs to not eagerly validate

jpschorr:
This changes the lexer and parser to pass through strings enclosed in backticks un-parsed. (At current, these documents are parsed during lowering).

Since embedded documents may themselves contain backticks, beginning and ending delimiters consist of an arbitrary odd numbers of backticks (e.g., `` ` ``, `` ``` ``, `` ````` `` etc.) that must be paired (e.g., `` `$ion_data_here::[]` ``, `` ```$ion_data_here::[ $string_with_embedded_backtick:"`" ]``` ``, etc.).

As opening and closing delimiters are required to be odd in count of backticks, a contiguous string of backticks that is even is interpreted as an empty document.

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #523, opened 2024-12-05
Refactor Value & Enable Lazy Ion literals.

jpschorr:
Merges the following previously PRed changes to `main`:
- #507
- #509 
-  #510 
-  #517 
-  #519

Also fixes a few minor lifetime warnings from newer versions of rust/clippy.

---


alancai98 (approved):
Looks like there's a slight regression in some conformance tests. Could take a brief look at those

jpschorr:
> Looks like there's a slight regression in some conformance tests. Could take a brief look at those

They are changes to static analysis tests. In making ion data lazy, the ion parsing is no longer happening at a time that triggers the failure for these tests. 

e.g., 

```
fail_semantics("char_length(`()`)");
```

No longer immediately errors, since the `()` isn't parsed prior to lowering to a logical plan anymore. I'll either change the way semantics tests are validated as part of the larger lazy document initiative, or tests like these will be changed, as 'semantic' is a fuzzy line between syntax and evaluation anyway.

PR #533, opened 2025-01-13
Introduce Datum, an interface to introspecting Values

jpschorr:
This PR introduces `Datum` and `DatumCategory`. These serve as interfaces for abstracting certain properties of PartiQL `Value`s without introspecting the `Value` directly.

This forms part of the basis for supporting external `Variant`s. 

Cf. https://github.com/partiql/partiql-lang-rust/pull/529

---

### DIFF
--- a/partiql-value/src/datum.rs
+++ b/partiql-value/src/datum.rs
@@ -1,0 +1,257 @@
+use crate::{Bag, BindingsName, List, Tuple, Value};
+use std::borrow::Cow;
+use std::error::Error;
+
+pub type DatumLowerError = Box<dyn Error>;
+pub type DatumLowerResult<T> = Result<T, DatumLowerError>;
+
+pub trait Datum<D>
+where
+    D: Datum<D>,
+{
+    #[inline]
+    /// Returns true if and only if Value is to be interpreted as `NULL`
+    #[must_use]
+    fn is_null(&self) -> bool {
+        false
+    }
+
+    #[inline]
+    /// Returns true if and only if Value is to be interpreted as `MISSING`
+    #[must_use]
+    fn is_missing(&self) -> bool {
+        false
+    }
+
+    #[inline]
+    /// Returns true if and only if Value is null or missing
+    #[must_use]
+    fn is_absent(&self) -> bool {
+        self.is_null() || self.is_missing()
+    }
+
+    /// Returns true if and only if Value is an integer, real, or decimal
+    #[must_use]
+    fn is_number(&self) -> bool;
+
+    #[inline]
+    /// Returns true if Value is neither null nor missing
+    #[must_use]
+    fn is_present(&self) -> bool {
+        !self.is_absent()
+    }
+
+    #[must_use]
+    fn is_sequence(&self) -> bool;
+
+    #[must_use]
+    fn is_ordered(&self) -> bool;
+}
+
+pub trait DatumValue<D>: Clone + Datum<D>
+where
+    D: Datum<D>,
+{
+    fn into_lower(self) -> DatumLowerResult<D>;
+}
+
+pub trait DatumCategory<'a> {
+    fn category(&'a self) -> DatumCategoryRef<'a>;
+    fn into_category(self) -> DatumCategoryOwned;
+}
+
+#[derive(Debug)]
+pub enum DatumCategoryRef<'a> {
+    Null,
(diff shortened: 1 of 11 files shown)


partiql-value/src/datum.rs
```diff
@@ -0,0 +1,255 @@
...
+pub enum DatumCategoryRef<'a> {
+    Null,
+    Missing,
+    Tuple(DatumTupleRef<'a>),
+    Sequence(DatumSeqRef<'a>),
+    Scalar(DatumValueRef<'a>),
+}
+
+#[derive(Debug)]
+pub enum DatumCategoryOwned {
```
alancai98, line 73: Curious about the structuring of these categories. PLK's Datum stores the Datum's kind/code as a PType https://github.com/partiql/partiql-lang-kotlin/blob/a7b772e4102fb4a0d61e2280202ed38aca269469/partiql-spi/src/main/java/org/partiql/spi/value/Datum.java#L47-L51 whereas this approach has Datum defining its own enum.

partiql-value/src/datum.rs
```diff
@@ -0,0 +1,255 @@
...
+    Null,
+    Missing,
```
alancai98, line 75: Any reason to have `NULL` and `MISSING` as their own enum variants? There are the `is_null` and `is_missing` functions you defined above but perhaps having them as their own variants makes some things easier to implement?

partiql-value/src/datum.rs
```diff
@@ -0,0 +1,255 @@
...
+    #[inline]
+    /// Returns true if and only if Value is null or missing
+    #[must_use]
+    fn is_absent(&self) -> bool {
+        self.is_null() || self.is_missing()
+    }
+
+    /// Returns true if and only if Value is an integer, real, or decimal
+    #[must_use]
+    fn is_number(&self) -> bool;
```
alancai98, line 35: Will some other methods be added for other types? Like `is_bool`, `is_text`, etc?

partiql-value/src/datum.rs
jpschorr: For one, the typing in PLK is further along than here. Perhaps, partiql-types will grow to take over some of this.

Using the `PType` approach would also require large blocks of `if`/`else if`/etc. for the cases in which this is used. `enum`s and `match` are the more idiomatic approach in rust.

Also... `PType` is also a bit odd. Decimals are `PTypeDecimal`, Text types are `PTypeWithMaxLength`, and other 'scalar's are `PTypePrimitive`. It's difficult with `PType` to split values into Scalar vs. Sequence vs. Tuple cleanly.

partiql-value/src/datum.rs
jpschorr: This is just moving some tests out of `Value` directly and into the `Datum` interface. Perhaps those methods will be needed in the future, yes. But, I imagine better matching of types in partiql-types will be preferred.

partiql-value/src/datum.rs
jpschorr: The question this is designed to answer is, roughly, 'what is the category of this value? Scalar? Tuple? Sequence?', as that basically helps in the determining how the value interacts with pathing/collection/control-flow operations. Likewise, `Null` and `Missing` influence this type of dispatching. 

We could require the consumer to check `is_null` and `is_missing` before asking for the category. But then, the category return would need to be an `Option` or `Result`, because we can't determine the category for a `Null` value. 

In this PR's model, the consumer needs to:
1. call `category`
   1. dispatch on scalar/sequence/tuple.

In the API model without `Null`/`Missing` category the consumer would need to:
1. call `is_null` and deal with the true case  
3. call `is_missing` and deal with the true case
4. call `category`
   1. unwrap and deal with the error case
   2. dispatch on scalar/sequence/tuple.

jpschorr: See also:
- https://github.com/partiql/partiql-lang-rust/pull/535/files#diff-b3a236452a3b5f704e7f2d2fa80983f726caed63e4dade21bf90b48c4b6546a6R26-R28

jpschorr: I'm not sure. 

The `is_*` here are simply moved out of `Value` and into this Trait.

Over the longer term, I'll take a look at existing usages and see. Perhaps they go away after a refactor.

PR #535, opened 2025-01-13
Add `category` for `PartiqlShape`

jpschorr:
This PR introduces `StaticCategory`, a parallel to `DatumCategory`. This aides in type-checking against `Datum`/`DatumCategory`.

This forms part of the basis for supporting external `Variant`s. 

Cf. https://github.com/partiql/partiql-lang-rust/pull/529

---

### DIFF
--- a/partiql-eval/src/eval/eval_expr_wrapper.rs
+++ b/partiql-eval/src/eval/eval_expr_wrapper.rs
@@ -4,7 +4,7 @@ use crate::eval::expr::{BindError, EvalExpr};
 use crate::eval::EvalContext;
 use itertools::Itertools;
 
-use partiql_types::{PartiqlShape, Static, TYPE_DYNAMIC};
+use partiql_types::{PartiqlShape, Static, StaticCategory, TYPE_DYNAMIC};
 use partiql_value::Value::{Missing, Null};
 use partiql_value::{Tuple, Value};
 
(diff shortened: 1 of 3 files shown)


partiql-eval/src/eval/eval_expr_wrapper.rs
```diff
@@ -14,44 +14,62 @@ use std::hash::Hash;
...
-            (Static::Struct(_), Value::Tuple(_)) => true,
-            (Static::Bag(b_type), Value::Bag(b_values)) => {
-                let bag_element_type = b_type.element_type();
-                let mut b_values = b_values.iter();
-                b_values.all(|b_value| subsumes(bag_element_type, b_value))
-            }
-            (Static::DateTime, Value::DateTime(_)) => true,
+trait TypeSatisfier {
+    fn satisfies(&self, value: &Value) -> bool;
+}
```
alancai98, line 22: nit: could add some docs for what `satisfies` means

alancai98 (approved):

partiql-types/src/lib.rs
```diff
@@ -742,6 +742,12 @@ pub enum Static {
     // TODO Add BitString, ByteString, Blob, Clob, and Graph types
 }
 
+pub enum StaticCategory<'a> {
+    Tuple(),
```
am357, line 746: `Struct`?

am357 (approved):

partiql-types/src/lib.rs
jpschorr: I think we probably actually want this to represent the 'category' of tuple types. Right now, that's just `Struct`, but may include, e.g., `Row` in the future. 

Similar to how it represents both `Bag` and `List` with `Sequence`.

PR #541, opened 2025-01-31
Fix pretty-print implementation to not use transmute.

jpschorr:
---

### DIFF
--- a/partiql-ast/src/pretty.rs
+++ b/partiql-ast/src/pretty.rs
@@ -697,28 +697,14 @@ impl PrettyDoc for Struct {
         D::Doc: Clone,
         A: Clone,
     {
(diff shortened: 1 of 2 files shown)


partiql-ast/src/pretty.rs
```diff
@@ -697,60 +697,32 @@ impl PrettyDoc for Struct {
         D::Doc: Clone,
         A: Clone,
     {
-        let wrapped = self.fields.iter().map(|p| unsafe {
-            let x: &'b StructExprPair = std::mem::transmute(p);
```
alancai98, line 701: Curious why these unsafe + transmute calls were there to begin with? Perhaps it's faster?

alancai98 (approved):

PR #547, opened 2025-03-15
Implement Graph evaluation for matches without alternatives

jpschorr:
An evaluator implementation of basic graph pattern matching that can be unit-tested.

## Description

The targeted language definition is what was proposed in [RFC-0025](https://github.com/partiql/partiql-lang/blob/main/RFCs/0025-graph-data-model.md) and [RFC-0033](https://github.com/partiql/partiql-lang/blob/main/RFCs/0033-graph-query.md).

Supported GPML patterns are those comprising:
- straight-path patterns
- simple label matches
- directed and undirected edge patterns
- no alternatives
- no path length quaifiers

The evaluation here is meant to be a simple and understandable mirror of the algorithm described in Section 6.4 of the GPML paper [[ACM](https://dl.acm.org/doi/10.1145/3514221.3526057), [arxiv](https://arxiv.org/abs/2112.06217).]

---


partiql-value/src/graph.rs
```diff
@@ -59,36 +58,45 @@ impl Ord for Graph {
...
+
+impl GElem {
+    pub fn new(value: Option<Value>, labels: GLabels) -> Self {
+        GElem { value, labels }
+    }
+}
 
 pub struct SimpleGraph {
-    directed: petgraph::stable_graph::StableGraph<GNodeId, GEdgeId, Directed>,
-    undirected: petgraph::stable_graph::StableGraph<GNodeId, GEdgeId, Undirected>,
```
alancai98, line 79: out of curiosity, what's the rationale to use our own representation rather than use petgraph?

partiql-eval/src/test_value.rs
```diff
@@ -0,0 +1,194 @@
+use partiql_value::{EqualityValue, NullableEq, Value};
```
alancai98, line 1: seems like this file is copied directly from partiql-conformance-tests (other than the eval version including `parse_partiql_value_str`). do we want to consolidate the test value logic into one reusable struct at some point? if so, could add a todo

partiql-eval/src/eval/graph/bind_name.rs
```diff
@@ -0,0 +1,45 @@
+use std::sync::atomic::AtomicU32;
+use std::sync::atomic::Ordering::Relaxed;
+
+/// A unicode non-charachter prefixed onto 'anonymous' bind names
```
alancai98, line 4:
```suggestion
/// A unicode non-character prefixed onto 'anonymous' bind names
```

alancai98:
also since the RFCs were moved from partiql-docs to partiql-lang, the above links should be changed to:
- https://github.com/partiql/partiql-lang/blob/main/RFCs/0025-graph-data-model.md
- https://github.com/partiql/partiql-lang/blob/main/RFCs/0033-graph-query.md

partiql-value/src/graph.rs
jpschorr: The complexity of dealing with two different graphs and trying to implement the GPML algorithm on-top of petgraph made it hard to follow. For something intended to be faster/better/etc. than a 'SimpleGraph', petgraph may be better, but it obfuscated the implementation here.

partiql-eval/src/eval/graph/plan.rs
```diff
@@ -0,0 +1,223 @@
...
+    Always,
+    Named(GT::Label),
+    Never,
+}
+
+/// A plan specification for value filtering.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum ValueFilter {
+    #[default]
+    Always,
```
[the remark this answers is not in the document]
jpschorr: That's the intent. GPML allows `WHERE` filters on nodes and edges. This is here to make sure the structure of filtering 'made sense', but is largely a TODO.

partiql-eval/src/test_value.rs
jpschorr: Yes. I'd like to. Where, exactly, to put it is the question.

partiql-eval/src/lib.rs
```diff
@@ -2434,4 +2435,293 @@ mod tests {
...
+            PathMatch, PathPatternMatch, StepFilter, TripleFilter,
+        };
+        use crate::eval::graph::string_graph::types::StringGraphTypes;
+        use crate::eval::{BasicContext, EvalGraphMatch};
+        use crate::test_value::TestValue;
+        use partiql_catalog::context::SystemContext;
+        use partiql_common::pretty::ToPretty;
+        use partiql_value::{BindingsName, DateTime, Value};
+
+        fn graph() -> Value {
```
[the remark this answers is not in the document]
jpschorr: I suppose the following, but I'm not sure it's clearer than the ion data...

```dot
digraph G {
    ranksep="1";
    subgraph nodes {
        rankdir="LR";
        rank = "min";
        node [shape=Mrecord, fontsize="30pt"];
        n1 [label="n1\n|{labels:['a']|payload:`1`}"];
        n2 [label="n2\n|{labels:['b']|payload:`2`}"];
        n3 [label="n3\n|{labels:['c']|payload:`3`}"];
    }
    
    node [shape=record, fontsize="15pt"];
    e12 [label="e12\n|{labels:['e']|payload:`1.2`}"];
    e23 [label="e23\n|{labels:['d']|payload:`2.3`}"];
    e_u [label="e_u\n|{labels:['self']|payload:`\<\< \>\>`}"];
        
    n1 -> e12 [dir=none];
    e12 -> n2 [dir=left];
    
    n2 -> e23 [dir=none];
    e23 -> n3 [dir=left];
    
    n3 -> e_u [dir=none];
    e_u -> n3 [dir=none];
}
```
simple(1)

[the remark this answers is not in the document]
alancai98: thanks, the `dot` graph makes the graph a bit easier to visualize for me

alancai98 (approved):

partiql-eval/src/eval/graph/plan.rs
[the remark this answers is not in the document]
am357: For this and other that is applicable, perhaps a `#[non_exhaustive]` can help

partiql-eval/src/test_value.rs
alancai98: maybe in `partiql-eval` is a better fit since it would be weird for `partiql-eval` to depend on `partiql-conformance-tests`? or possibly some other util/testing crate we create

partiql-value/src/graph.rs
am357: I add that that it is one less dependency which is also a plus; obviously considering that it seems, it is making the implementation harder, based on @jpschorr's note.

partiql-eval/src/test_value.rs
jpschorr: Yeah. I think a new util crate of some sort. I intend to work on it as part of the graph workstream.

alancai98 (approved):

am357 (approved):

PR #558, opened 2025-04-10
Use `Datum` for tuple coercion/bindings

jpschorr:
Use features from #533 for tuple coercion.

---

### DIFF
--- a/partiql-value/src/value.rs
+++ b/partiql-value/src/value.rs
@@ -15,7 +15,10 @@ mod iter;
 mod logic;
 mod math;
 
-use crate::datum::{Datum, DatumLower, DatumLowerResult, DatumValue};
+use crate::datum::{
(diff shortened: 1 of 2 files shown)


partiql-value/src/value.rs
```diff
@@ -15,7 +15,10 @@ mod iter;
 mod logic;
 mod math;
 
-use crate::datum::{Datum, DatumLower, DatumLowerResult, DatumValue};
+use crate::datum::{
+    Datum, DatumCategory, DatumCategoryOwned, DatumCategoryRef, DatumLower, DatumLowerResult,
+    DatumSeqRef, DatumTupleOwned, DatumTupleRef, DatumValue,
```
alancai98, line 20: some clippy warnings but could be fixed in a later PR to prevent rebase/merge conflicts. we should probably also investigate at some point why the CI isn't reporting these

alancai98 (approved):

PR #575, opened 2025-05-13
Lower & evaluate Graph Match Modes

jpschorr:
Lower & evaluate Graph Match Modes (`WALK`, `TRAIL`, `ACYCLIC`, `SIMPLE`) on paths and subpaths in `GRAPH MATCH` expression.

See new tests that are expected to pass: 

https://github.com/partiql/partiql-tests/pull/139

 ---

### DIFF
--- a/partiql-logical-planner/src/graph.rs
+++ b/partiql-logical-planner/src/graph.rs
@@ -2,6 +2,7 @@ use num::Integer;
 use partiql_ast::ast;
 
 use crate::lower::extract_vexpr_by_id;
+use itertools::Itertools;
 use partiql_ast::ast::GraphPathPrefix;
 use partiql_common::node::{IdAnnotated, NodeId};
 use partiql_logical::graph::bind_name::FreshBinder;
@@ -185,7 +186,8 @@ impl GraphToLogical {
         pattern: &ast::GraphPattern,
     ) -> Result<Vec<MatchElement>, String> {
         if pattern.match_mode.is_some() {
-            // The current graph engine is basically doing RepeatableElements currently
+            // The current graph engine is basically executing as though
+            // `pattern.match_mode` is `Some(GraphMatchMode::RepeatableElements)` currently
             not_yet_implemented_result!("MATCH expression MATCH MODE is not yet supported.");
         }
         if pattern.keep.is_some() {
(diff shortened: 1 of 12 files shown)


partiql-logical-planner/src/graph.rs
```diff
@@ -250,10 +184,12 @@ impl GraphToLogical {
         &mut self,
         pattern: &ast::GraphPattern,
     ) -> Result<Vec<MatchElement>, String> {
-        if pattern.mode.is_some() {
+        if pattern.match_mode.is_some() {
+            // The current graph engine is basically doing RepeatableElements currently
```
alancai98, line 188: Could you clarify this comment? What is `RepeatableElements`?

partiql-logical-planner/src/graph.rs
```diff
@@ -60,7 +61,8 @@ impl<'a> Normalize<'a> {
     /// 'Normalize' a series of node/edge/subpath match elements into a rigid pattern of
     ///    (<node> ( <edge> <node>)*)
     pub fn normalize(mut self, elements: Vec<MatchElement>) -> Result<Vec<PathPattern>, String> {
-        self.do_normalize(elements)
+        let path_mode = PathMode::Walk; // TODO replace with input to function?
+        self.do_normalize(elements, path_mode)
```
alancai98, line 65: Should `Walk` be hard-coded here? Perhaps this is causing the code coverage miss for the other `PathMode`s -- https://app.codecov.io/gh/partiql/partiql-lang-rust/pull/575?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=partiql

partiql-logical-planner/src/graph.rs
jpschorr: The current graph engine is basically executing as though `pattern.match_mode` is `Some(GraphMatchMode::RepeatableElements)` currently

partiql-logical-planner/src/graph.rs
jpschorr: It shouldn't matter, as, if you look inside `do_normalize`, this hard-coded mode only applies to single-node and single-edge matches for which mode doesn't matter.

I've added some additional tests to make sure the appropriate lines are hit.

alancai98: Ah I see now

alancai98 (approved):

PR #584, opened 2025-07-08
Add some documentation for printing error messages nicely

jpschorr:
---

### DIFF
--- a/partiql-parser/src/lib.rs
+++ b/partiql-parser/src/lib.rs
@@ -7,6 +7,10 @@
 //! # Usage
 //!
 //! ```
+//! use std::fmt;
+//! use std::fmt::Formatter;
+//! use itertools::Itertools;
+//! use partiql_common::syntax::location::LineAndColumn;
 //! use partiql_parser::{Parser, ParserError, ParserResult};
 //!
 //! let parser = Parser::default();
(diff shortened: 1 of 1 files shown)


partiql-parser/src/lib.rs
```diff
@@ -15,9 +19,35 @@
...
+//!
+//!
+//!
+//! // Print out messages with custom line:column offsets
+//! #[derive(Debug, Copy, Clone, PartialEq, Eq, Ord, PartialOrd, Hash)]
+//! pub struct VerboseLineAndColumn(LineAndColumn);
+//!
+//! impl fmt::Display for VerboseLineAndColumn {
+//!     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+//!         write!(f, "Line {}, Offset {}", self.0.line, self.0.column)
```
dlurton, line 44: Maybe include the message here too?

dlurton: actually, I haven't grokked this fully yet... one sec

jpschorr: `VerboseLineAndColumn` is just a 'newtype' containing `LineAndColumn` so that we can override display to show `Line 1, Offset 20` instead of `1:20`

compare:
```
assert_eq!(err_msg, "Unexpected token `<a:UNQUOTED_IDENT>` at `(1:20..1:21)`");
```

with 

```
Unexpected token `<a:UNQUOTED_IDENT>` at `(Line 1, Offset 20..Line 1, Offset 21)`
```

dlurton: Ok.  Yeah,  The API is.. not what I would have expected, but the example helps.  Thanks.

dlurton (approved):

alancai98 (approved):

PR #630, opened 2026-06-05
refactor: extract partiql-hybrid query pipeline into execute_query

hashPirate:
Moved the parse/lower/compile/execute phases and timing summary out of main into a standalone execute_query function that returns Result instead of calling std::process::exit, so that it can be reused by the upcoming pqlite REPL. main keeps CLI arg parsing for now.

Also added clap and reedline as partiql-tools dependencies.


johnedquinn (approved):
LGTM ! Eventually we can plumb through some flag to leverage the AST/plan/bytecode printing. Nice work.

PR #633, opened 2026-06-08
feat: add interactive REPL in pqlite binary

hashPirate:
* refactor: duplicate partiql-hybrid into new pqlite binary

  Copy partiql_hybrid.rs into src/bin/pqlite.rs and register both
  targets in Cargo.toml so they build independently during the
  transition from the single-shot benchmarking tool to an interactive
  engine. Updated the Cli docstring and command name to pqlite.

  * feat: implement reedline REPL in pqlite

  Replace the 'REPL mode triggered!' placeholder in the no-subcommand
  arm with a reedline read loop. Initializes a Reedline instance with a
  custom 'pqlite> ' prompt and dispatches each non-empty line to the
  existing execute_query, reusing the shared pipeline.

  Signal::Success runs the query and, unlike the exec subcommand,
  reports execute_query errors via eprintln without exiting so the
  session continues. Signal::CtrlC prints '^C' and loops; Signal::CtrlD
  prints 'Exiting...' and breaks for a clean exit.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -14,6 +14,7 @@ use std::time::Instant;
 use clap::{Parser, Subcommand};
 use reedline::{
     FileBackedHistory, History, Prompt, PromptEditMode, PromptHistorySearch, Reedline, Signal,
+    ValidationResult, Validator,
 };
 
 const BATCH_SIZE: usize = 1;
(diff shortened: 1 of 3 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -0,0 +1,577 @@
...
+/// Resolve the user's home directory across platforms.
+fn home_dir() -> Option<std::path::PathBuf> {
+    // `std::env::home_dir` was un-deprecated in Rust 1.85 and is correct on
+    // all supported platforms, so no extra crate is needed here.
+    #[allow(deprecated)]
+    std::env::home_dir()
+}
+
+/// Print the REPL help menu listing meta-commands and query usage.
+fn print_help() {
```
johnedquinn, line 192: There is something that might interest you: https://docs.rs/reedline-repl-rs/latest/reedline_repl_rs/struct.Repl.html

partiql-tools/src/bin/pqlite.rs
```diff
@@ -0,0 +1,577 @@
...
+            MetaOutcome::Handled
+        }
+    }
+}
+
+/// Run the interactive REPL: read a line, execute it as a query, and loop.
+///
+/// Errors from `execute_query` are reported but never terminate the session.
+fn run_repl(data_source: &str, data_path: Option<&String>) {
+    let mut line_editor = Reedline::create()
```
johnedquinn, line 246: Can you add version information to startup? Something like:
```
pqlite version <version from cargo aka 0.15.0-alpha.1>@<git commit sha>
For usage information, enter ".help".
```

johnedquinn (approved):
Looks good. Some non-blocking comments that can be followed-up in a subsequent PR, or a modification of this one. Nice work.

partiql-tools/src/bin/pqlite.rs
hashPirate: of course! working on that now

johnedquinn: And, a `-V`/`--version` to clap that utilizes the same logic, prints, and then exits 😄

hashPirate: that is useful! I'll have both work

johnedquinn (approved):

PR #636, opened 2026-06-10
feat: parse CREATE TABLE ... AS (query) (CTAS) in pqlite

hashPirate:
Adds grammar, AST, and pretty-printing support for CREATE TABLE
  statements, including the optional CTAS clause backed by a full
  subquery.

  Key additions:
  - Add Create lexer token and register it as a keyword
  - Add Ddl/DdlOp/CreateTable/CtasClause grammar rules; AS (query) is optional, so plain CREATE TABLE <name> parses with as_query = None
  - Add CreateTable AST node (table_name + optional boxed query)
  - Re-type the top-level parse result as an Item enum (Ddl/Dml/Query)
  - Implement PrettyDoc for Item/DdlOp/CreateTable, replacing a panicking todo! in the pretty-printer
  - Route DDL through the logical planner as NotYetImplemented (parses cleanly; lowering/eval is future work)
  - Add parser unit tests: plain CREATE TABLE, simple CTAS, complex CTAS with WHERE/ORDER BY/LIMIT, and graceful failure on malformed input
  - Bump pqlite (partiql-tools) to 0.15.0-alpha.2

### DIFF
--- a/partiql-tools/Cargo.toml
+++ b/partiql-tools/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "partiql-tools"
-version = "0.15.0-alpha.2"
+version = "0.15.0-alpha.1"
 edition = "2021"
 publish = false
 autobins = false
--- a/partiql-tools/Cargo.toml
+++ b/partiql-tools/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "partiql-tools"
-version = "0.15.0-alpha.1"
+version = "0.15.0-alpha.2"
 edition = "2021"
 publish = false
 autobins = false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
+- *BREAKING* partiql-parser: `Parsed`'s `ast` field type changed from `ast::AstNode<ast::TopLevelQuery>` to `ast::AstNode<ast::Item>` to support DDL statements alongside queries. Code that accesses `parsed.ast` must now match on `ast::Item` (e.g. `Item::Query`) to reach the query node.
 
 ### Added
+- partiql-parser: Parsing support for `CREATE TABLE <name>` and `CREATE TABLE <name> AS (<query>)` (CTAS). DDL lowering/evaluation is not yet implemented and surfaces a `NotYetImplemented` error.
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Changed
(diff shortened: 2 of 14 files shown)


johnedquinn:
Nice work getting CREATE TABLE wired up through the grammar and AST — the lalrpop additions are clean and the test coverage is good.

A few suggestions on the structural design that I think will make this a stronger foundation going forward:

### 1. Rename `Item` → `Statement`

You already named the grammar rule `Statement` (which generates `StatementParser`) — let's carry that through to the Rust type. `Item` was a placeholder that was never reachable from the parser until now, so this is the right time to give it a proper name. `Statement` is the standard SQL term and makes the API self-documenting.

### 2. Avoid double `AstNode` wrapping at the root

The current design produces `AstNode<Item>` containing `Item::Query(AstNode<TopLevelQuery>)` — two nested `AstNode`s with the same span. In this codebase, each `AstNode` is a visitor scope boundary (`enter_ast_node`/`exit_ast_node` fires for each), and the lowerer tracks them on an `id_stack`. Two nodes with identical spans at the root creates a redundant scope entry.

The existing convention is that `AstNode` marks points where the **span can differ from the parent** and where passes need a distinct scope. For example, `AstNode<TopLevelQuery>` contains `AstNode<Query>` because WITH can shift the inner span. But wrapping `TopLevelQuery` inside another `AstNode<Item>` at the same byte range adds nothing.

The fix is to let the outer `AstNode<Statement>` be the one root scope, and hold the variants bare:

```rust
pub enum Statement {
    Query(TopLevelQuery),   // no extra AstNode — outer provides the scope
    Ddl(Ddl),               // already bare
    Dml(Dml),               // already bare
}
```

Inner sub-expressions (like the query inside CTAS) still get their own `AstNode` wrapping through the normal grammar rules — that's correct because their spans genuinely differ from the statement span.

### 3. Return a `Vec<AstNode<Statement>>` from `Parsed`

Since we're already breaking the `Parsed.ast` type, this is the cheapest time to model the output as a list of statements:

```rust
pub struct Parsed<'input> {
    pub statements: Vec<AstNode<Statement>>,
    // ...
}
```

Real SQL input is semicolon-delimited (`CREATE TABLE foo; SELECT * FROM foo;`). The `StatementList` grammar rule with optional `;` separators is something you could add in this PR or leave for a follow-up — but the `Parsed` type should be updated now to avoid another breaking change later.

### 4. Minor: the `Ddl` wrapper struct

This is pre-existing, not something you introduced — but worth noting: `Ddl` is `{ pub op: DdlOp }`, a single-field struct. Since `Statement::Ddl(...)` already labels it as DDL, you could inline it as `Statement::Ddl(DdlOp)` and remove a layer. Optional — fine either way.

partiql-tools/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "partiql-tools"
-version = "0.15.0-alpha.1"
+version = "0.15.0-alpha.2"
```
johnedquinn: Let's drop this change for now since we won't be releasing anytime soon, and the version is held in the top-level Cargo.toml. We could set the version here to the workspace version, but also not needed in this PR.

hashPirate: Great catches! Thank you so much, will work through those changes now.

johnedquinn (approved):
Looks great! Thanks for this effort.

PR #639, opened 2026-06-15
feat: lower CREATE TABLE AS into a top-level LogicalStatement

hashPirate:
Add CTAS logical planning: CREATE TABLE <name> and CREATE TABLE <name> AS (<query>) now lower into a new top-level LogicalStatement enum that wraps the relational  plan, rather than living inside the BindingsOp graph. DDL is a statement category containing a query, not a relational operator. Planning only for now.

- partiql-logical: Add LogicalStatement enum (Query / CreateTableAs / CreateTable) and target table identity is carried as a case-preserving BindingsName.

- partiql-logical-planner: Add LogicalPlanner::lower_statement, safely reusing
    the existing two-pass pipeline on the inner query. Retained lower() as a back-compatible version and it now rejects DDL at the front door so inner-query errors don't leak.

- pqlite: Wire REPL tool execution flow to lower_statement. Ordinary queries execute on the register VM as before, while CreateTableAs and CreateTable statements intercept execution printing out the planned logical plan graph.

### DIFF
--- a/partiql-logical-planner/src/lib.rs
+++ b/partiql-logical-planner/src/lib.rs
@@ -6,6 +6,7 @@ use crate::lower::AstToLogical;
 use partiql_ast::ast;
 use partiql_ast_passes::error::{AstTransformError, AstTransformationError};
 use partiql_ast_passes::name_resolver::NameResolver;
+use partiql_common::node::NodeId;
 use partiql_logical as logical;
 use partiql_parser::Parsed;
 
@@ -21,16 +22,30 @@ pub struct LogicalPlanner<'c> {
     catalog: &'c dyn SharedCatalog,
 }
 
+/// The uniform "DDL not yet lowerable" error, shared by every DDL-rejection site.
+#[inline]
+fn ddl_not_yet_implemented() -> AstTransformationError {
+    AstTransformationError {
+        errors: vec![AstTransformError::NotYetImplemented(
+            "DDL statement lowering".to_string(),
+        )],
+    }
+}
+
 impl<'c> LogicalPlanner<'c> {
     pub fn new(catalog: &'c dyn SharedCatalog) -> Self {
         LogicalPlanner { catalog }
     }
 
+    /// Lower a parsed statement into a top-level [`logical::LogicalStatement`].
+    ///
+    /// This is the full-fidelity entry point: it preserves the statement
+    /// category (query vs. DDL). [`Self::lower`] is a back-compat shim over this.
     #[inline]
-    pub fn lower(
+    pub fn lower_statement(
         &self,
         parsed: &Parsed<'_>,
-    ) -> Result<logical::LogicalPlan<logical::BindingsOp>, AstTransformationError> {
+    ) -> Result<logical::LogicalStatement, AstTransformationError> {
         if parsed.statements.len() != 1 {
             return Err(AstTransformationError {
                 errors: vec![AstTransformError::NotYetImplemented(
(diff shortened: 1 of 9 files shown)


partiql-logical-planner/src/lib.rs
```diff
@@ -41,21 +56,79 @@ impl<'c> LogicalPlanner<'c> {
...
+                match &ct.as_query {
+                    None => Ok(logical::LogicalStatement::CreateTable { table_name }),
+                    Some(inner) => {
+                        // Wrap the inner AstNode<Query> in a TopLevelQuery to feed the
+                        // existing pipeline. Lossless: Query has no `with` field, so
+                        // `with: None` drops nothing. Id-safe: we clone the original
+                        // node, so its parse-time NodeId is preserved verbatim;
+                        // both NameResolver and AstToLogical key on that id. Never
+                        // reconstruct the node with a fresh id or search_locals breaks.
+                        let wrapped = ast::TopLevelQuery {
```
johnedquinn: Oh, interesting. I didn't realize that CTAS didn't add support for the WITH clause too in the AST. By specification, it should hold a TopLevelQuery -- that's my bad from the last review. We shouldn't need to manually wrap here.

partiql-tools/src/bin/pqlite.rs
```diff
@@ -312,11 +333,38 @@ fn execute_query(query_str: &str, debug: &DebugFlags) -> Result<(), Box<dyn std:
...
 
+    // DDL statements are planning-only for now: print the lowered plan and stop.
+    // (Compile/execute and storage are a later slice; the consumer would
+    // orchestrate writes around the inner query's execution.)
+    let logical = match statement {
+        LogicalStatement::Query(plan) => plan,
+        LogicalStatement::CreateTableAs { table_name, query } => {
+            // `{}` (Display) on the plan, not `{:?}`: LogicalPlan has a readable
+            // Display impl; Debug would dump the raw nodes/edges struct.
+            println!(
```
johnedquinn, line 358: We already have the planning being printed on line 368. Longer term, let's remove these prints.

johnedquinn (changes requested):

partiql-logical-planner/src/lib.rs
hashPirate: Gotcha! Working on a refactor

hashPirate: The source for the CTAS is now a TopLevelQuery in the grammar/AST so the duplicate wrapping should be fixed.

partiql-logical-planner/src/lib.rs
```diff
@@ -21,16 +22,30 @@ pub struct LogicalPlanner<'c> {
...
 
+    /// Lower a parsed statement into a top-level [`logical::LogicalStatement`].
+    ///
+    /// This is the full-fidelity entry point: it preserves the statement
+    /// category (query vs. DDL). [`Self::lower`] is a back-compat shim over this.
     #[inline]
-    pub fn lower(
+    pub fn lower_statement(
         &self,
         parsed: &Parsed<'_>,
```
johnedquinn: Can this new API just take in a statement? Should make more sense from the consumer's perspective.

hashPirate: Good point! I refactored lower_statement to take &AstNodeStatement so that we still have the NodeID (edited cause git formatted my comment haha)

johnedquinn (approved):
Nice work!

PR #641, opened 2026-06-16
feat: add LMDB storage env, --db flag, and _tables catalog to pqlite

hashPirate:
Add a storage module to partiql-tools backed by heed 0.20 (LMDB). HeedDB
  opens a single-file environment and creates the _tables system catalog,
  exposing the env, catalog, and path. The pqlite REPL gains a global --db
  flag that selects the database file (defaulting to ~/.pqlite/default.pal),
  opens it on startup, and hard-fails with a clear message if it cannot be
  opened. The startup banner reports the active path on stderr so stdout
  stays reserved for query results.

  The database is opened lazily, only for the interactive REPL: a one-shot
  exec query is read-only today and must not create a database file as a
  side effect. heed is pulled with default-features disabled since the serde
  codecs are unused.

  Storage stays on the consumer side: the catalog is created but not written
  to, and the handle is not threaded into query execution. Those are
  follow-up PRs.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -5,7 +5,7 @@ use partiql_eval::plan::EvaluationMode;
 use partiql_eval::value::Shape;
 use partiql_eval::{CompilationContext, ExecutionContext, PlanCompiler};
 use partiql_logical::LogicalStatement;
-use partiql_tools::storage::{default_db_path, HeedDB};
+use partiql_tools::storage::HeedDB;
 use partiql_value::{Tuple, Value};
 use std::borrow::Cow;
 use std::time::Instant;
@@ -35,7 +35,8 @@ struct Cli {
     #[arg(long, global = true, value_delimiter = ',')]
     debug: Vec<String>,
 
-    /// Path to the database file. Defaults to ~/.pqlite/default.pal.
+    /// Path to the database file (required for the interactive REPL). The
+    /// parent directory must already exist; it is not created for you.
     #[arg(long, global = true)]
     db: Option<std::path::PathBuf>,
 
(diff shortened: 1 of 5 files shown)


partiql-tools/src/storage.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// Returns `None` if the home directory cannot be resolved; the binary treats
+/// that as a startup error and asks the user to pass `--db`.
+pub fn default_db_path() -> Option<PathBuf> {
+    // `std::env::home_dir` was un-deprecated in Rust 1.85 and is correct on all
+    // supported platforms, matching how the pqlite binary resolves home.
+    #[allow(deprecated)]
+    let mut home = std::env::home_dir()?;
+    home.push(".pqlite");
+    home.push("default.pal");
```
johnedquinn: Why default.pal? Can you make it default.pqlite or default.db?

partiql-tools/src/storage.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+/// Create the parent directory of `path` if needed.
+///
+/// In single-file (`NO_SUB_DIR`) mode LMDB does not create the containing
+/// directory, so we must. A bare filename like `local.pal` has a parent of
+/// `Some("")`; `create_dir_all("")` errors with `NotFound`, so we skip
+/// creation when the parent is absent or empty.
+fn ensure_parent_dir(path: &Path) -> std::io::Result<()> {
```
johnedquinn: I'm looking now at the need for creating directories on behalf of the user. Typically UNIX commands will error out if a parent directory doesn't exist.

I imagine this was created due to the default being `~/.pqlite/...`. I'm now thinking let's just avoid creating a default db file. I think for now we can just require the --db flag.

In the near term (in a later PR), we can extract the storage layer into a trait -- and for repl's that don't have a storage layer, attempting to create/etc should fail or persist to an in-memory DB. But, let's leave that out of scope for now.

partiql-tools/src/storage.rs
```diff
@@ -0,0 +1,284 @@
...
+                .max_dbs(MAX_DBS)
+                .flags(heed::EnvFlags::NO_SUB_DIR)
+                .open(path)?
+        };
+
+        // 3. Open/create the `_tables` catalog in a write transaction.
+        //    `create_database` borrows the RwTxn mutably and opens the db if it
+        //    already exists, so reopening is not an error.
+        let mut wtxn = env.write_txn()?;
+        let tables: TablesDb = env.create_database(&mut wtxn, Some(TABLES_DB))?;
```
johnedquinn, line 104: Note: we're going to want to insert a row to let the tables table know that the tables table exists 😁. We can discuss this offline. But, we can do this in a follow-up.

johnedquinn (changes requested):

partiql-tools/src/storage.rs
johnedquinn: Ignore this for now, given my other comment.

partiql-tools/src/storage.rs
hashPirate: That makes sense and simplifies things a lot, I'll make --db required

partiql-tools/src/bin/pqlite.rs
```diff
@@ -81,7 +87,32 @@ fn main() {
             }
         }
         None => {
-            run_repl(&debug);
+            // The REPL needs a database. We require an explicit --db rather
```
johnedquinn, line 90: We're going to want to add this to the CLI exec as well. But, non-blocking.

johnedquinn (approved):

PR #642, opened 2026-06-18
feat: implement CREATE TABLE AS write path on LMDB

hashPirate:
Wires up CREATE TABLE AS end to end: run the inner query through the VM and write one dummy byte per row into a new LMDB table. Real Ion serialization comes in a later PR. Storage is designed so a plain query via exec doesn't need --db. CREATE TABLE AS opens the --db path when it actually needs it, and the REPL just reuses its session handle.

- HeedDB::create_table_from_rows does the write in a single txn: reserved-name guard for the _ namespace, catalog duplicate check, table creation, then the row loop. Any error partway through drops the txn and LMDB rolls it back.
- canonical_table_key lowercases bare identifiers and leaves quoted ones alone, which matches how PartiQL treats identifier case for the catalog key.
- Pulled build_compiled and build_exec_context out of execute_query so the query and CTAS paths share the same compile and exec setup.
- All storage errors now go through Display, so the error text on stderr is consistent. stdout stays empty on a write.
- Renamed the two leftover .pal test fixtures to .pqlite to finish off that rename.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -5,7 +5,7 @@ use partiql_eval::plan::EvaluationMode;
 use partiql_eval::value::Shape;
 use partiql_eval::{CompilationContext, ExecutionContext, PlanCompiler};
 use partiql_logical::LogicalStatement;
-use partiql_tools::storage::HeedDB;
+use partiql_tools::storage::{HeedDB, StorageError};
 use partiql_value::{Tuple, Value};
 use std::borrow::Cow;
 use std::time::Instant;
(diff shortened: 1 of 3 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -388,14 +476,82 @@ fn execute_query(query_str: &str, debug: &DebugFlags) -> Result<(), Box<dyn std:
...
+            let n = match vm.execute() {
+                Ok(partiql_eval::ExecutionResult::Query(iter)) => {
+                    // Map each VM row to Result<(), StorageError>: PR 2 discards
+                    // the (not-yet-serialized) row data and keeps only
+                    // success/failure. The VM's error type is stringified here
+                    // so storage stays independent of partiql-eval.
+                    let rows = iter.map(|r| {
+                        r.map(|_row| ())
+                            .map_err(|e| StorageError::Execution(format!("{:?}", e)))
+                    });
```
johnedquinn, line 524: Just a note for the next PR: we're going to want to write on every row rather than gathering them, because the RegisterReader gives us an immediate view of the state of the registers per-row. If we accumulate them before writing, then we'll have to write to a buffer and then to Heed.

I say that without looking into the overhead of doing that, but we at least want to look into the performance implications during the next PR.

partiql-tools/src/storage.rs
```diff
@@ -111,6 +170,60 @@ impl HeedDB {
...
+            return Err(StorageError::InvalidName(name.to_string()));
+        }
+
+        let mut wtxn = self.env.write_txn()?;
+
+        // Duplicate check against the catalog, in the same txn. `RwTxn` derefs
+        // to `RoTxn`, so the existence check and the creation are atomic.
+        if self.tables.get(&wtxn, name)?.is_some() {
+            return Err(StorageError::TableExists(name.to_string()));
+        }
```
johnedquinn, line 209: Very nice.

johnedquinn (approved):

partiql-tools/src/bin/pqlite.rs
hashPirate: Completely agree! I actually ran into this issue today too when drafting the design for PR3. Thanks for the green light!!

PR #644, opened 2026-06-24
feat: implement zero-allocation binary row serialization

hashPirate:
We chose custom binary encoding over Ion-rs for CTAS to get full control over our storage footprint and eliminate per-row heap overhead.

This PR implements a custom tagged-union format that encodes VM register state directly to disk. The new storage driver uses a push-based callback pattern instead of passing around vectors of bytes, the VM now pushes rows into a reused, scratch-pad buffer. This gets us to a true zero-allocation hot path after the initial buffer warmup.

Key changes:

- Wire Format: A lean, ordered binary format (LE values, BE keys) that bakes in schema versioning so the future reader knows exactly what it's looking at, even for empty tables.

- Filesystem Integrity: Added a peek-then-open check so that if the encoder rejects a row early on we can back out without ever leaving orphan database files on disk.

- Safety: Added caps (max fields, max name length) and duplicate column detection in the encoder beforehand, catching data issues before they ever reach the storage layer.

- Reliability: Updated our integration tests (27 tests total) to include bit-perfect roundtrip checks and some truncation testing.

### Performance Benchmarks (CTAS Write Path)

| N rows | Exec time | μs per row | RSS | Peak heap |
| :--- | :--- | :--- | :--- | :--- |
| 10K | 12.3 ms | 1.23 | 12 MB | 4.7 MB |
| 100K | 29.2 ms | 0.29 | 16 MB | 8.0 MB |
| 1M | 187.2 ms | 0.19 | 50 MB | 42 MB |
| 10M | 2.04 s | 0.20 | 378 MB | 388 MB |

Each row is one `pqlite exec` run by `/usr/bin/time -lp` on an M4 Pro Macbook by running `CREATE TABLE bench AS (SELECT t.a FROM mem(1, N) t)` at N = 10K, 100K, 1M, and 10M. mem(1,N) is a built in function that makes N rows with the shape {a: i64}, each row ends up being a total of 12 bytes (1B version + 1B struct tag + 4B field count + 4B name length + 1B name + 1B value + 8B i64). 

The reason we scaled to 10 million is because at that point the LMDB B+tree is 3 to 4 levels deep instead of 2 at 10K, and any extra overhead should show up in  μs/row . The flat μs/row from 100k to 10M is evidence that the path stays O(1).

### Updated Benchmarks (Dynamic Type-check and Builder pattern)

| N rows | Exec time | μs per row | RSS | Peak heap |
| :--- | :--- | :--- | :--- | :--- |
| 10K | 28.2 ms | 2.82 | 12 MB | 4.7 MB |
| 100K | 41.9 ms | 0.419 | 16 MB | 8.2 MB |
| 1M | 228.6 ms | 0.229 | 51 MB | 42.8 MB |
| 10M | 2.26 s | 0.226 | 396 MB | 388.8 MB |

Measured exactly the same way as the previous runs, but these are the benchmarks after adding the dynamic type-checking for non-struct values. The change to using builder pattern did allow scaling from 1M to 10M rows to show a **9.87x** multiplier which is less than the previous **10.89x**.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -570,32 +570,29 @@ fn execute_query(
                         }
                     };
 
-                    let driver_result = db.create_table_from_rows(&key, |push_row| {
-                        // SAFETY-RELEVANT: QueryIterator::next uses unsafe
-                        // lifetime extension on the RegisterReader it yields
-                        // (partiql-eval/src/engine/plan.rs:892-895). Aliasing
-                        // a RegisterReader across iter.next() is UB. We
-                        // consume `row` synchronously inside serialize_row,
-                        // then drop it; push_row sees only the encoded bytes.
-                        if let Some(ref bytes) = first_row_bytes {
-                            push_row(bytes)?;
-                        }
-                        for r in iter.by_ref() {
-                            let row = r.map_err(|e| StorageError::Execution(format!("{:?}", e)))?;
-                            partiql_tools::row_codec::serialize_row(
-                                &row,
-                                &row_shape,
-                                &mut scratch_buf,
-                            )
-                            .map_err(|e| StorageError::Codec(format!("{e}")))?;
-                            push_row(&scratch_buf)?;
-                        }
-                        Ok(())
-                    });
-                    match driver_result {
-                        Ok(n) => n,
-                        Err(e) => return Err(format!("Error: {}", e).into()),
+                    let mut writer = db.create_table(&key).map_err(|e| format!("Error: {}", e))?;
+                    // SAFETY-RELEVANT: QueryIterator::next uses unsafe lifetime
+                    // extension on the RegisterReader it yields. Aliasing across
+                    // iter.next() is UB. We consume `row` synchronously inside
+                    // serialize_row, drop it, then push the bytes.
+                    if let Some(ref bytes) = first_row_bytes {
+                        writer
+                            .push_row(bytes)
+                            .map_err(|e| format!("Error: {}", e))?;
                     }
+                    for r in iter.by_ref() {
+                        let row = r.map_err(|e| {
+                            format!("Error: {}", StorageError::Execution(format!("{:?}", e)))
+                        })?;
+                        partiql_tools::row_codec::serialize_row(&row, &row_shape, &mut scratch_buf)
+                            .map_err(|e| {
+                                format!("Error: {}", StorageError::Codec(format!("{e}")))
+                            })?;
+                        writer
+                            .push_row(&scratch_buf)
+                            .map_err(|e| format!("Error: {}", e))?;
+                    }
+                    writer.commit().map_err(|e| format!("Error: {}", e))?
                 }
                 Err(e) => return Err(format!("Execution setup error: {:?}", e).into()),
             };
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -76,11 +76,6 @@ enum Commands {
 /// whole session and passes it as `Open`; `exec` passes `Lazy` so the database
 /// is opened only if a write statement actually requires it — a plain query via
 /// `exec` never opens or creates a file.
-///
-/// This is CLI dispatch plumbing, not a storage abstraction: it encodes the
-/// three valid call states (live handle / lazy path / no path) as two variants
-/// so an illegal "both a handle and a path" state is unrepresentable. It is
-/// unrelated to any future engine-agnostic storage trait.
 #[derive(Clone, Copy)]
 enum DbSource<'a> {
     /// An already-open handle (the REPL session database).
@@ -115,11 +110,9 @@ fn main() {
                 std::process::exit(1);
             });
 
-            // Open the database up front. A database that silently stops
-            // persisting is worse than one that refuses to start, so this
-            // hard-fails (unlike the REPL history file, which falls back to
-            // in-memory by design). The parent directory must already exist;
-            // the filesystem error bubbles up naturally if it does not.
+            // Open the database up front. Hard-fail rather than fall back (a
+            // silently-non-persisting db is worse than one that refuses to
+            // start). The parent directory must already exist.
             let db = match HeedDB::open(&db_path) {
                 Ok(db) => db,
                 Err(e) => {
@@ -320,14 +313,9 @@ fn run_repl(debug: &DebugFlags, db: &HeedDB) {
                     }
                 }
 
-                // Strip the trailing `;` terminator (preserving internal
-                // newlines) before handing the full multi-line text to the
-                // engine — shared with the `exec` subcommand for consistency.
                 let query = normalize_query(trimmed);
 
-                // A lone `;` (e.g. on its own line) leaves nothing to run once
-                // the terminator is stripped — skip it rather than handing an
-                // empty string to the parser.
+                // A lone `;` strips to empty — skip rather than parse "".
                 if query.is_empty() {
                     continue;
                 }
@@ -374,8 +362,8 @@ fn format_table_name(name: &partiql_value::BindingsName<'_>) -> String {
 ///
 /// NOTE: `name` may contain arbitrary bytes from a quoted identifier. An
 /// interior NUL byte would otherwise panic heed's internal
-/// `CString::new(name).unwrap()`; `HeedDB::create_table_from_rows` guards
-/// against that up front and returns `StorageError::InvalidName` instead.
+/// `CString::new(name).unwrap()`; `HeedDB::create_table` guards against that
+/// up front and returns `StorageError::InvalidName` instead.
 fn canonical_table_key(name: &partiql_value::BindingsName<'_>) -> String {
     match name {
         partiql_value::BindingsName::CaseInsensitive(s) => s.to_lowercase(),
@@ -470,22 +458,13 @@ fn execute_query(
         lower_statement(&*catalog, stmt).map_err(|e| format!("Lower error: {:?}", e))?;
     let lower_time = lower_start.elapsed();
 
-    // DDL statements are planning-only for now: print the lowered plan and stop.
-    // (Compile/execute and storage are a later slice; the consumer would
-    // orchestrate writes around the inner query's execution.)
     let logical = match statement {
         LogicalStatement::Query(plan) => plan,
         LogicalStatement::CreateTableAs { table_name, query } => {
             let key = canonical_table_key(&table_name);
 
-            // Validate --db early but DO NOT open the env yet. Deferring the
-            // open until after compile + VM setup means a parse, compile, or
-            // ROW-0 schema-rejection exits without creating any .pqlite file
-            // on disk. A mid-stream rejection (row 50 hits a Bool, etc.)
-            // safely rolls back the wtxn — no catalog entry, no row bytes —
-            // but the env file on disk persists. That is the streaming-engine
-            // trade-off; protecting mid-stream would require buffering the
-            // whole result set, which defeats the per-row write model.
+            // Resolve --db before compile + VM setup so a missing --db on a
+            // CTAS path is a clean CLI error before any expensive work.
             let db_path: Option<&std::path::Path> = match db_source {
                 DbSource::Open(_) => None,
                 DbSource::Lazy(path) => Some(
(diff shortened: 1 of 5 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -478,31 +478,21 @@ fn execute_query(
...
-                }
+            // Validate --db early but DO NOT open the env yet. Deferring the
+            // open until after compile + VM setup means a parse, compile, or
+            // ROW-0 schema-rejection exits without creating any .pqlite file
+            // on disk. A mid-stream rejection (row 50 hits a Bool, etc.)
+            // safely rolls back the wtxn — no catalog entry, no row bytes —
+            // but the env file on disk persists. That is the streaming-engine
+            // trade-off; protecting mid-stream would require buffering the
+            // whole result set, which defeats the per-row write model.
+            let db_path: Option<&std::path::Path> = match db_source {
```
johnedquinn, line 466: I'm kind of thinking we should have two commands:
1. `pqlite open <db_ref>`
2. `pqlite exec <db_ref> <statements>`

For this PR, we don't need to change anything, but somewhere in the near-term let's move to this. Then, we can setup a special db_ref called `:memory:` once we eventually extract some of Heed's logic to a trait. Then, we can create an in-memory implementation of the persistence layer.

So, for this PR, no need to change anything. In follow-ups:
1. Remove `--db` and add both commands above. pqlite top-level command does nothing, just holds subcommands. (you can pull this change in now or later)
2. Create trait for storage layer, and have a Heed impl (we don't want to do this until we have an end-to-end working so that we know what we want out of the trait)
3. Add special support for `:memory:` with an in-memory impl of the trait

partiql-tools/src/bin/pqlite.rs
```diff
@@ -512,22 +502,98 @@ fn execute_query(
...
-                Ok(partiql_eval::ExecutionResult::Query(iter)) => {
-                    // Map each VM row to Result<(), StorageError>: PR 2 discards
-                    // the (not-yet-serialized) row data and keeps only
-                    // success/failure. The VM's error type is stringified here
-                    // so storage stays independent of partiql-eval.
-                    let rows = iter.map(|r| {
-                        r.map(|_row| ())
-                            .map_err(|e| StorageError::Execution(format!("{:?}", e)))
+                Ok(partiql_eval::ExecutionResult::Query(mut iter)) => {
+                    let first_row_bytes: Option<Vec<u8>> = match iter.next() {
```
johnedquinn: Is there a tangible gain for not just opening the file immediately and then writing each row sequentially to heed? I might be missing some context on why the first row materialization is necessary.

If it's just for orphan files, I think I'm fine living with that especially considering my above comment.

partiql-tools/src/storage.rs
```diff
@@ -170,55 +187,70 @@ impl HeedDB {
...
-        // Dummy write loop: sequential u64 key, constant byte value.
         let mut row_id: u64 = 0;
-        for row in rows {
-            row?; // a VM error (Execution) bubbles out; dropping wtxn rolls back.
-            table.put(&mut wtxn, &row_id, &[0x00])?;
+        // Stack-local 8-byte key buffer reused across every push. Combined
+        // with `RowKey = heed::types::Bytes` (Cow::Borrowed), this gives
+        // zero heap allocations per row for the key side of the put.
+        let mut key_buf: [u8; 8] = [0u8; 8];
+        let mut push_row = |row_bytes: &[u8]| -> Result<(), StorageError> {
```
johnedquinn: One structural suggestion: the callback/closure pattern for `create_table_from_rows` works, but I think a builder pattern would give you a cleaner separation of concerns. Right now storage owns the loop structure (via the closure), but it's really pqlite that knows *how* to drive rows — storage should just expose primitives. The `dyn FnMut` signature is also a bit hard to parse at a glance.

Something like:

```rust
// storage.rs exposes:
let mut writer = db.create_table("t")?;  // opens wtxn, dup check, etc.
writer.push_row(&bytes)?;
let n = writer.commit()?;                // drop without commit = rollback

// pqlite.rs drives:
let db = open_db(path)?;
let mut writer = db.create_table(&key)?;
for r in iter {
    serialize_row(&row, &row_shape, &mut scratch_buf)?;
    writer.push_row(&scratch_buf)?;
}
let n = writer.commit()?;
```

partiql-tools/tests/pqlite_cli.rs
```diff
@@ -177,3 +177,464 @@ fn duplicate_ctas_is_rejected_and_names_the_table() {
...
+    // Mirror the encoder's MAX_FIELDS_PER_ROW cap: a corrupt/oversize count
+    // would otherwise OOM the test runner via Vec::with_capacity.
+    assert!(
+        field_count <= MAX_FIELDS_PER_ROW,
+        "sanity: field_count={field_count} exceeds MAX_FIELDS_PER_ROW ({MAX_FIELDS_PER_ROW})"
+    );
+
+    let mut fields = Vec::with_capacity(field_count as usize);
+    for _ in 0..field_count {
+        let name_len = take_u32(&mut i, bytes);
```
johnedquinn: We're going to want to restructure this to be more flexible. PartiQL allows database objects to be of any type, not just collections of rows. So, we could have just a collection of integers, or even just a single integer.

What tangibly changes is that you don't know whether what you're serializing/parsing is a struct or a different type. So we can't just immediately attempt to grab fields here.

Let's update this PR to provide that flexibility, both on serialization and deserialization.

partiql-tools/src/row_codec.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+    }
+}
+
+impl std::error::Error for SerializeError {}
+
+/// Serialize one VM row into `buf` as a tagged-union byte sequence.
+///
+/// `buf` is cleared on entry. On success, `buf` contains:
+///   * `FORMAT_VERSION` (1 byte)
```
johnedquinn: Eventually I believe I want format versions to be either application-level or table-level. We can keep it here for now, but posting this comment for posterity.

partiql-tools/src/row_codec.rs
```diff
@@ -0,0 +1,316 @@
...
+                    .expect("string column value longer than u32::MAX bytes — wire format requires length fits in u32");
+                buf.extend_from_slice(&str_len.to_le_bytes());
+                buf.extend_from_slice(sb);
+            }
+            ValueType::Bool
+            | ValueType::Bytes
+            | ValueType::Missing
+            | ValueType::Tuple
+            | ValueType::List
+            | ValueType::Bag => unreachable!("pre-flight rejects all reserved/container types"),
```
johnedquinn: We can keep these out of scope for this PR, however, still consider my other comment regarding per-row values not necessarily needing to be structs.

johnedquinn (changes requested):
This work is really awesome to see 😄.

Can you provide some more insight into the benchmark query that you ran in the PR description? Also, for some of these generated inline comments, can you turn it down a notch in verbosity unless it's for public documentation or truly clearing up something?

partiql-tools/src/storage.rs
hashPirate: Thats a great call, thank you so much for the code snippet! I can see how the `dyn FnMut` callback ends up being more difficult to read. I will refactor this to use a Writer builder to make it easier to test.

partiql-tools/src/bin/pqlite.rs
hashPirate: I agree, there is no gain other than just being a defensive measure for the orphan `.pqlite` when there is a reject for row-0.  I will remove the peek before opening in order to simplify initialization. The orphan concern shall be handled once we go to `pqlite open/exec` structure.

partiql-tools/src/bin/pqlite.rs
hashPirate: That makes sense, and would definitely make the engine a lot more versatile. I can keep this in mind when I work on the storage driver, and will make sure the TableWriter pattern can be easily changed for that trait.

partiql-tools/tests/pqlite_cli.rs
hashPirate: You're right! My current implementation is a bit too focused on rows. I can refactor the serialization entry to be more flexible to types right at the beginning. I'll make sure the codec supports all PartiQL types.

partiql-tools/src/bin/pqlite.rs
johnedquinn: Also, in that case, we can open the DB immediately upon starting execution. Even before parsing. We can clean that up in a future PR.

partiql-tools/src/row_codec.rs
```diff
@@ -0,0 +1,266 @@
...
+        if name.len() > MAX_NAME_LEN_BYTES as usize {
+            return Err(SerializeError::Unsupported(format!(
+                "column {col}: name length {} bytes exceeds MAX_NAME_LEN_BYTES ({})",
+                name.len(),
+                MAX_NAME_LEN_BYTES
+            )));
+        }
+        // Duplicate-name check: bounded by MAX_FIELDS_PER_ROW so the O(K²)
+        // is acceptable.
+        for prior in &fields[..col] {
```
johnedquinn: PartiQL allows for duplicate field names

partiql-tools/src/row_codec.rs
```diff
@@ -0,0 +1,266 @@
...
+                return Err(SerializeError::Unsupported(format!(
+                    "column '{name}': nested struct values are not supported yet"
+                )));
+            }
+        };
+        let view = row.get_value_view(slot).expect("register slot from shape");
+        validate_value_type(view.get_type(), Some(name))?;
+    }
+
+    // Pre-flight passed: write the row.
```
johnedquinn: I'd honestly lean towards removing the pre-flight checks. Go for immediate serialization, and propagate any serialization error up (no unreachable macros), which will then rollback the transaction. So, no need for the type validation, duplicate name checks, unsupported projections, etc.

johnedquinn (changes requested):

partiql-tools/src/row_codec.rs
hashPirate: Speedy! Will get that done.

hashPirate: I kept the 2 structural limit ones inline during streaming so that we can keep a speedy single-pass but also make sure that the encoder doesn't write a row that will be rejected by the reader later. Very happy about your approach!

partiql-tools/tests/pqlite_cli.rs
```diff
@@ -177,3 +177,510 @@ fn duplicate_ctas_is_rejected_and_names_the_table() {
...
+            let name_len = take_u32(&mut i, bytes);
+            assert!(
+                name_len <= MAX_NAME_LEN_BYTES,
+                "sanity: name_len={name_len} exceeds MAX_NAME_LEN_BYTES ({MAX_NAME_LEN_BYTES})"
+            );
+            let name = std::str::from_utf8(take(&mut i, name_len as usize, bytes))
+                .expect("field name must be valid UTF-8")
+                .to_string();
+            let tag = take(&mut i, 1, bytes)[0];
+            fields.push((name, take_value(&mut i, bytes, tag)));
```
johnedquinn: With this PR, we are able to serialize/parse single structs/tuples, but not nested. If you have the logic, why not add support? Should make some of your code simpler since ParsedValue is no longer a variant of ParsedPayload.

Also, if you're going to address this, can you pull in:
1. the removal of version format
2. rename TAG_STRUCT to TAG_TUPLE to keep the naming aligned

hashPirate: Makes sense! I originally deferred it to PR5, waiting for the schema registry  so that we could handle the tracking on disk. Bringing them together makes sense and it will clean up the code. Will remove versioning and do the rename too. Working on this now.

partiql-tools/src/row_codec.rs
```diff
@@ -0,0 +1,188 @@
...
+            let s = view.get_str().expect("string view");
+            let sb = s.as_bytes();
+            let str_len: u32 = sb.len().try_into().expect("str_len exceeds u32::MAX");
+            buf.extend_from_slice(&str_len.to_le_bytes());
+            buf.extend_from_slice(sb);
+        }
+        ty @ (ValueType::Bool
+        | ValueType::Bytes
+        | ValueType::Missing
+        | ValueType::Tuple
```
johnedquinn, line 175: This is also an avenue for nested structs. Though, when you add support for the other types, we can address this path.

johnedquinn (approved):
Nice work!

PR #645, opened 2026-06-29
feat(pqlite): add read path (CTAS → SELECT loop) 

hashPirate:
# feat: add read path for pqlite

Hey everyone! This PR adds the read path for pqlite under partiql-tools/. 

In our most recent storage PR (#644), we got the write path working (CREATE TABLE AS). This PR finishes the query loop so that if you save data into a table using CTAS, you can read it right back out with a standard SELECT *. 

This satisfies the milestone exit criteria: CTAS followed by SELECT round-trips end-to-end.

7 files changed: +1248 / -666. Verified: 53 tests total (30 library, 3 binary, 20 integration), clippy clean, and fmt clean. Significant test churn replaces old manual parsing fixtures from #644 with a robust end-to-end integration test suite.

------------------------------------------------------------------------
Verification Example:
------------------------------------------------------------------------
pqlite --db test.pqlite exec "CREATE TABLE t AS (SELECT 'hello' AS status FROM mem(1, 1) m)"
pqlite --db test.pqlite exec "SELECT * FROM t"
Output: << { 'status': 'hello' } >>

------------------------------------------------------------------------
Backstory: Where pqlite Came From
------------------------------------------------------------------------
If you haven't been tracking the recent storage sprints, pqlite is a small embeddable PartiQL engine living at partiql-tools/src/bin/pqlite.rs. Think of it as "SQLite for PartiQL." It persists tables in a single-file LMDB database (via the heed crate) and uses partiql-eval as its execution library.

It's been built up across three previous steps:
- PR1 (#641): Added the basic LMDB environment hook, the --db flag for picking a database file, and the internal _tables system catalog to track which tables exist.
- PR2 (#642): Added parsing and logical planning for CREATE TABLE AS. CTAS lives at a new LogicalStatement layer above LogicalPlan, since DDL acts as a statement category rather than a typical relational query operator.
- PR3 (#644): Shipped the write path. Introduced a custom tagged-union wire format for rows (serialize_row in row_codec.rs) that encodes VM register state directly to bytes with zero allocation per row after warmup. All multi-byte values are little-endian, but the row-id keys in LMDB are big-endian so lexicographic byte order matches numeric sort order.
- This PR (Read Milestone): Adds the read path. Brings the catalog, DataSource implementation, and CLI binary plumbing needed to scan tables back out with a standard SELECT.

Going custom on the wire format (instead of throwing Ion or an off-the-shelf serializer at it) was a deliberate design choice in #644: it gives us total control over the on-disk footprint and cuts out per-row heap overhead. The decoder added in this PR is the direct, symmetric companion to #644's encoder. 

------------------------------------------------------------------------
What I Did:
------------------------------------------------------------------------

 ### 1. Added Read Primitives — `partiql-tools/src/storage.rs`

  Three new methods on `HeedDB` so the query engine can pull rows out:
  - `read_txn()` — opens a read-only LMDB transaction.
  - `open_table(&txn, name)` — gets a handle to a persisted table's row store.
  - `list_table_names(&txn)` — lists every table in our `_tables` catalog.

  Also added `StorageError::ScanTooLarge` as a safety check so a single
  scan can't balloon past 4 GiB without surfacing a clean error.

  ### 2. Built the Symmetric Row Decoder — `partiql-tools/src/row_codec.rs`

  Wrote `deserialize_row_into` to unpack raw disk bytes straight into the
  VM's registers with zero middle-man allocations.

  - Marked `unsafe` because it pulls zero-copy string slices directly
    from the source buffer using lifetime extensions. The caller must
    keep that buffer alive for the duration of the scan.

  ### 3. Wired up Catalog & DataSource — `partiql-tools/src/catalog.rs` (new)

  New module connecting our storage layer to `partiql_eval`'s catalog traits.

  - `HeedTableMetadata::resolve` intentionally returns `None` for every
    field name. This forces the compiler down the `WholeValue` projection
    branch, dropping the whole row tuple into register slot 0 so the VM
    can resolve fields at runtime.
  - `HeedExecutionCatalog` eagerly collects scanned rows into one flat
    `Vec<u8>` buffer plus a `Vec<Range<u32>>` index, replacing the legacy
    per-row allocations with a single amortized-growth pattern.

  ### 4. Cleaned up the REPL Binary — `partiql-tools/src/bin/pqlite.rs`

  Refactored `execute_query` to share a single open `HeedDB` instance
  across statements, eliminating the wtxn-commit + fsync overhead of
  reopening the environment on every query.

  - Added `CombinedCatalog`: since `PlanCompiler` hardcodes the single
    catalog name `"default"`, this wrapper structurally bundles our
    persisted tables alongside built-in table functions (`rand`, `mem`,
    `scan_ion`) under one name.
  - Added a fail-fast check during compilation so a typo in a table
    name surfaces as `Error: Table '<name>' not found` instead of
    letting Permissive mode silently yield a fake MISSING row.

  ### 5. Added Solid Test Coverage — `partiql-tools/tests/pqlite_cli.rs`

  Replaced the old manual parsing fixtures with an end-to-end integration
  suite of 20 tests. The suite covers:

  - End-to-end subprocess round-trips: scalars, mixed types, empty tables,
    multi-field tuples, multi-table databases.
  - Case-sensitive vs bare quoted-identifier behavior.
  - Byte-shape anchor test asserting exact disk layout so LE/BE encoding
    never drifts out of sync.
  - Fail-fast tests for SELECT against unknown tables.
  - Malformed byte-injection tests to confirm corrupt rows surface as
    clean stderr errors instead of panics.

------------------------------------------------------------------------
Commit Structure (Fully Bisectable)
------------------------------------------------------------------------
The history is split into 5 layered commits so that each step makes sense and cargo check passes completely on every single revision:

1. 8607f8d -- Storage primitives: read_txn, open_table, list_table_names
2. d4e0781 -- Row codec: The symmetric decoder 
3. 75a5feb -- Catalog: Hooking up HeedCompilationCatalog and HeedTableSource
4. 985cc2f -- Binary wiring: Open-once database caching, CombinedCatalog, and fail-fast checking
5. d9aea11 -- Tests: Full E2E CLI suite and error surfacing blocks

------------------------------------------------------------------------
Known Upstream Limitations (Pre-existing on dev branch)
------------------------------------------------------------------------
These queries fail or return wrong results because of bugs or missing features inside the upstream partiql-eval or partiql-logical-planner crates--this PR didn't touch those directories:
- Aliased FROM clauses: Writing FROM table alias returns MISSING because of a local rewriting bug in search_locals (Logged with John on Slack to fix in a future planner PR). Workaround: unaliased formats like SELECT col FROM table cover all current tests.
- Arithmetic on persisted columns: Performing operations like m.col * 5 fails with "Type mismatch for add!" because it is a downstream symptom of the alias-MISSING bug.
- Constant SELECTs: Writing SELECT 42 fails compilation with InvalidPlan because the upstream planner requires a relational input source to run a projection.
- Cross Joins: Writing FROM a, b errors out because join operators aren't mapped in the VM bytecode loop yet.

------------------------------------------------------------------------
Deferred to Follow-up PRs
------------------------------------------------------------------------
To hit our exact read/write loop targets for this initial milestone, the following items have been pushed out to our future roadmap:

- Universal Type Support (Our Immediate Next Milestone / Next Storage PR):
  Expanding the row codec to read and write all remaining PartiQL types, including deeply nested data like arrays, bags, and structs.

- INSERT Statement Support:
  Adding the parser, AST, and logical planning plumbing for INSERT statements so callers can append rows to existing tables without re-running CTAS.

- System Table Expansion & Value-Based Keying:
  Building out the rest of our system databases (SCHEMAS, CATALOG, FUNCTIONS) and migrating our storage engine to key tables using serialized PartiQL values instead of direct strings.

- Schema Registry & Real Scan Projections:
  Because we do not store table schemas yet, the catalog defaults to a WholeValue layout fallback. True column projection pushdowns are deferred until the schema catalog layer is implemented.

- Streaming Row Iteration:
  Currently, HeedExecutionCatalog materializes all of a table's rows into memory before delivering the first one. A streaming implementation that co-owns the RoTxn and RoIter is a potential follow-up but out of scope for now.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -1,13 +1,16 @@
 use partiql_tools::common;
 
 use common::{create_table_fn_catalog, lower_statement, parse};
+use partiql_common::catalog::CatalogId;
 use partiql_eval::plan::EvaluationMode;
 use partiql_eval::value::Shape;
-use partiql_eval::{CompilationContext, ExecutionContext, PlanCompiler};
+use partiql_eval::{CompilationContext, ExecutionCatalog, ExecutionContext, PlanCompiler};
 use partiql_logical::LogicalStatement;
+use partiql_tools::catalog::{HeedCompilationCatalog, HeedExecutionCatalog};
 use partiql_tools::storage::{HeedDB, StorageError};
 use partiql_value::{Tuple, Value};
 use std::borrow::Cow;
+use std::sync::{Arc, Mutex};
 use std::time::Instant;
 
 use clap::{Parser, Subcommand};
@@ -16,11 +19,9 @@ use reedline::{
     ValidationResult, Validator,
 };
 
-/// Maximum number of lines retained in the REPL history file.
 const HISTORY_CAPACITY: usize = 1000;
 
-/// Version string shared by the `--version` flag and the REPL startup banner:
-/// the crate version joined with the git commit SHA captured in build.rs.
+/// Crate version joined with the git commit SHA captured in build.rs.
 const VERSION: &str = concat!(env!("CARGO_PKG_VERSION"), "@", env!("PQLITE_GIT_SHA"));
 
 /// Pqlite: An interactive PartiQL database engine and REPL.
@@ -35,8 +36,7 @@ struct Cli {
     #[arg(long, global = true, value_delimiter = ',')]
     debug: Vec<String>,
 
-    /// Path to the database file (required for the interactive REPL). The
-    /// parent directory must already exist; it is not created for you.
+    /// Path to the database file. The parent directory must already exist.
     #[arg(long, global = true)]
     db: Option<std::path::PathBuf>,
 
@@ -70,21 +70,6 @@ enum Commands {
     },
 }
 
-/// How `execute_query` obtains a database when a statement needs one.
-///
-/// Only `CREATE TABLE AS` needs storage. The REPL opens one handle for the
-/// whole session and passes it as `Open`; `exec` passes `Lazy` so the database
-/// is opened only if a write statement actually requires it — a plain query via
-/// `exec` never opens or creates a file.
-#[derive(Clone, Copy)]
-enum DbSource<'a> {
-    /// An already-open handle (the REPL session database).
-    Open(&'a HeedDB),
-    /// A `--db` path to open on demand (the `exec` subcommand); `None` if no
-    /// `--db` was given.
-    Lazy(Option<&'a std::path::Path>),
-}
-
 fn main() {
     let cli = Cli::parse();
     let debug = DebugFlags::from_args(&cli.debug);
@@ -96,55 +81,41 @@ fn main() {
                 eprintln!("Error: empty query");
                 std::process::exit(1);
             }
-            if let Err(e) = execute_query(query, &debug, DbSource::Lazy(cli.db.as_deref())) {
+            let db = match cli.db.as_deref() {
+                Some(p) => match HeedDB::open(p) {
+                    Ok(db) => Some(Arc::new(db)),
+                    Err(e) => {
+                        eprintln!("Error: could not open database at {}: {}", p.display(), e);
+                        std::process::exit(1);
+                    }
+                },
+                None => None,
+            };
+            if let Err(e) = execute_query(query, &debug, db) {
                 eprintln!("{}", e);
                 std::process::exit(1);
             }
         }
         None => {
-            // The REPL needs a database. We require an explicit --db rather
-            // than inventing a default path: like standard UNIX tools, we don't
-            // create files or directories on the user's behalf.
+            // Require explicit --db rather than inventing a default path.
             let db_path = cli.db.clone().unwrap_or_else(|| {
                 eprintln!("Error: The `--db <PATH>` option is required to open the database.");
                 std::process::exit(1);
             });
 
-            // Open the database up front. Hard-fail rather than fall back (a
-            // silently-non-persisting db is worse than one that refuses to
-            // start). The parent directory must already exist.
-            let db = match HeedDB::open(&db_path) {
-                Ok(db) => db,
-                Err(e) => {
-                    eprintln!(
-                        "Error: could not open database at {}: {}",
-                        db_path.display(),
-                        e
-                    );
-                    std::process::exit(1);
-                }
-            };
-
-            run_repl(&debug, &db);
+            run_repl(&debug, &db_path);
         }
     }
 }
 
-/// Normalize a query string before handing it to the engine: trim outer
-/// whitespace and strip a single trailing `;` statement terminator, while
-/// preserving any internal newlines. The `;` is accepted as a convenience
-/// terminator (and is what the REPL validator uses to detect a complete
-/// entry) but is not part of the PartiQL grammar, so it must be removed.
-///
-/// TODO: support the trailing `;` statement terminator natively in the
-/// PartiQL grammar/parser so semicolons are a first-class part of the
-/// language, and remove this stripping workaround once that lands.
+/// Trim outer whitespace and strip a single trailing `;` so the REPL can use
+/// `;` as a completion signal without it leaking into the PartiQL grammar.
+/// TODO: drop this once the parser accepts `;` natively.
 fn normalize_query(input: &str) -> &str {
     let trimmed = input.trim();
     trimmed.strip_suffix(';').unwrap_or(trimmed).trim_end()
 }
 
-/// Minimal REPL prompt that renders a fixed `pqlite> ` indicator.
 struct PqlitePrompt;
 
 impl Prompt for PqlitePrompt {
@@ -172,15 +143,8 @@ impl Prompt for PqlitePrompt {
     }
 }
 
-/// Validator that decides when a multi-line REPL entry is complete.
-///
-/// An input is considered complete when, after trimming trailing whitespace,
-/// it either:
-///   * ends with a semicolon (`;`) — a finished PartiQL statement, or
-///   * is a meta-command starting with a dot (`.`).
-///
-/// Anything else is treated as `Incomplete`, so pressing Enter inserts a
-/// newline and lets the user keep typing on the following line.
+/// Marks a REPL entry as complete when it ends with `;` or is a `.`-prefixed
+/// meta-command; anything else keeps reading on the next line.
 struct PqliteValidator;
 
 impl Validator for PqliteValidator {
@@ -194,10 +158,8 @@ impl Validator for PqliteValidator {
     }
 }
 
-/// Build a persistent, file-backed history at `~/.pqlite_history`.
-///
-/// Falls back to in-memory history (never crashing) if the home directory
-/// cannot be resolved or the history file cannot be opened.
+/// File-backed history at `~/.pqlite_history`; falls back to in-memory on
+/// resolution or open failure.
 fn build_history() -> Box<dyn History> {
     let history_path = home_dir().map(|mut p| {
         p.push(".pqlite_history");
@@ -219,22 +181,18 @@ fn build_history() -> Box<dyn History> {
         eprintln!("Warning: could not resolve home directory (using in-memory history)");
     }
 
-    // Fallback: in-memory history. If even this fails, run without history.
     match FileBackedHistory::new(HISTORY_CAPACITY) {
         Ok(history) => Box::new(history),
         Err(_) => Box::new(FileBackedHistory::default()),
     }
 }
 
-/// Resolve the user's home directory across platforms.
 fn home_dir() -> Option<std::path::PathBuf> {
-    // `std::env::home_dir` was un-deprecated in Rust 1.85 and is correct on
-    // all supported platforms, so no extra crate is needed here.
+    // `std::env::home_dir` was un-deprecated in Rust 1.85.
     #[allow(deprecated)]
     std::env::home_dir()
 }
 
-/// Print the REPL help menu listing meta-commands and query usage.
 fn print_help() {
     println!("PartiQL REPL — available commands:");
     println!("  .help     Show this help message");
@@ -250,15 +208,11 @@ fn print_help() {
     println!("Example: SELECT t.a, t.b FROM mem(100, 2) t LIMIT 5;");
 }
 
-/// Outcome of handling a REPL meta-command (a line starting with `.`).
 enum MetaOutcome {
-    /// The command was handled; continue the loop.
     Handled,
-    /// The user requested to quit; break the loop.
     Quit,
 }
 
-/// Handle a dot-prefixed meta-command. `input` must be the trimmed buffer.
 fn handle_meta_command(input: &str) -> MetaOutcome {
     match input {
         ".quit" | ".exit" => MetaOutcome::Quit,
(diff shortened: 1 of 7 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -16,11 +19,9 @@ use reedline::{
     ValidationResult, Validator,
 };
 
-/// Maximum number of lines retained in the REPL history file.
```
xd1313113, line 19: It looks you removed/changed a lot of comments. Do you do intentionally? 
Some comment seems still make sense and do not need remove

partiql-tools/src/catalog.rs
```diff
@@ -0,0 +1,286 @@
...
+fn hash_to_entry_id(name: &str) -> EntryId {
+    let mut hasher = std::collections::hash_map::DefaultHasher::new();
+    std::hash::Hash::hash(name, &mut hasher);
+    EntryId::from(std::hash::Hasher::finish(&hasher))
+}
+
+/// Pack the entire scan into one byte buffer with per-row `u32` offsets.
+/// One `Vec<u8>` (geometric growth) replaces one allocation per row.
+/// Fails with `ScanTooLarge` if the packed size exceeds `u32::MAX`.
+fn read_all_rows(
```
xd1313113, line 168: Do we read whole table for now? Or we have a streaming consumed pattern to avoid large data are loaded into memory directly?

partiql-tools/src/storage.rs
```diff
@@ -194,14 +215,36 @@ impl HeedDB {
...
+            .open_database(txn, Some(name))?
+            .ok_or_else(|| StorageError::TableMissing(name.to_string()))
+    }
+
+    /// Enumerate every table name in `_tables`, in lexicographic key order.
+    pub fn list_table_names(&self, txn: &heed::RoTxn<'_>) -> Result<Vec<String>, StorageError> {
+        let mut out =
+            Vec::with_capacity(self.tables.len(txn).map_err(StorageError::Heed)? as usize);
+        for result in self.tables.iter(txn).map_err(StorageError::Heed)? {
+            let (k, _v) = result.map_err(StorageError::Heed)?;
```
XuechunHHH: nit: why only wrap in `map_err` in this func?

partiql-tools/src/storage.rs
```diff
@@ -194,14 +215,36 @@ impl HeedDB {
...
+    pub fn read_txn(&self) -> Result<heed::RoTxn<'_>, StorageError> {
+        Ok(self.env.read_txn()?)
+    }
+
+    /// Open the named table for read. Returns `TableMissing` if no named
+    /// database called `name` exists in the env (i.e., the table was never
+    /// created or was dropped).
+    pub fn open_table(&self, txn: &heed::RoTxn<'_>, name: &str) -> Result<RowDb, StorageError> {
+        self.env
+            .open_database(txn, Some(name))?
```
XuechunHHH, line 231: double check: Is it possible that a named db existing without the table?

partiql-tools/src/row_codec.rs
```diff
@@ -84,7 +62,12 @@ pub fn serialize_row(
     buf.clear();
     match row_shape {
         RowShape::Struct(fields) => write_tuple(row, fields, buf),
-        RowShape::Register(slot, _) => write_value(row, *slot, None, buf),
+        // A bare scalar tag with no tuple frame round-trips on write but
```
XuechunHHH, line 65: I see @johnedquinn asked about supporting non-struct values:
> PartiQL allows database objects to be of any type, not just collections of rows.

The comment here says this fails because the decoder's `vw.put_scalar `requires a non-empty frame stack. Is that a `ValueWriter` limitation we should fix upstream, or are we accepting this as a permanent constraint?

partiql-tools/src/row_codec.rs
```diff
@@ -186,3 +167,269 @@ fn write_value(
...
+/// `bytes` must outlive every arena reset that touches `target_slot`'s
+/// register. In practice the caller (a `DataSource::next_row`
+/// implementation) must hold the backing row buffer for the duration
+/// of the scan. Violating this precondition causes use-after-free of
+/// the string slices written into the engine's arena via the
+/// lifetime extensions inside `decode_tagged_into`.
+///
+/// String UTF-8 validity is checked inline by `take_str` as each
+/// string is read; no pre-pass is required.
+pub unsafe fn deserialize_row_into(
```
XuechunHHH, line 214: Given that the caller `read_all_rows` already materializes the entire table into a `Vec<u8>`, I'm curious whether the `unsafe` buys measurable gains at current scale.

Do you mind to do a quick benchmark for zero-copy vs a safe memcpy-into-arena?

partiql-tools/src/row_codec.rs
```diff
@@ -186,3 +167,269 @@ fn write_value(
...
+    bytes: &[u8],
+    writer: &mut partiql_eval::source::RegisterWriter<'_, '_>,
+    target_slot: u16,
+) -> Result<(), DeserializeError> {
+    let mut cursor = 0usize;
+    let mut vw = writer
+        .value_writer(target_slot)
+        .map_err(|e| DeserializeError::Unsupported(format!("value_writer({target_slot}): {e}")))?;
+    decode_tagged_into(&mut vw, bytes, &mut cursor)?;
```
XuechunHHH, line 223: nit: not sure if it is worth to add a check for `cursor != bytes.len()`?

partiql-tools/src/catalog.rs
```diff
@@ -0,0 +1,297 @@
...
```
XuechunHHH, line 168: I see this is a tech debt here, do you mind to add a one line comment like:
```
// Eagerly materialized because `Box<dyn DataSource + 'static>` cannot hold a borrowed RoTxn.
```

XuechunHHH (commented):
Nice work! The architecture looks good to me overall. Left a few inline comments, mostly nits and clarification questions. I didn't go through the test suite thoroughly but the coverage looks solid from what I can see.

partiql-tools/src/storage.rs
hashPirate: That is a good catch! I can see the inconsistency and will use just ? for overall cohesiveness.

partiql-tools/src/storage.rs
hashPirate: Great question! For create_table, making the named db and writing the marker into the _tables catalog should happen in the same transaction. I can take a defensive approach here for sure and check  _tables before the return.

partiql-tools/src/row_codec.rs
hashPirate: Yes, good question! Supporting other PartiQL types is the next milestone in the plan and I have been evaluating the options for the past few days. My current plan is to wrap the top level scalars in a single frame on the encode and then unwrap it on the decode. ValueWriter does expose step_in_list and step_in_bag so I would not have to change any engine crates, so this would be a constraint only for this PR.

partiql-tools/src/row_codec.rs
hashPirate: Good idea, I'll do this now.

partiql-tools/src/row_codec.rs
hashPirate: It is worth adding the check, there can be edge cases where the decoder finishes and drops some bytes. Shall do!

partiql-tools/src/row_codec.rs
hashPirate: I forgot about this earlier, but the reason I went with the unsafe block was because of the borrow checker. `ValueWriter::put_str `takes a `&'a str` that is directly tied to the engine arena lifetime. A fresh safe String inside will not be able to comply with that lifetime and the safe path doesn't compile. I wanted to avoid touching other crates like partiql-eval.

I did however run the benchmark on the existing decode and for N=1M the throughput is ~0.8 μs/row if we use a full scan as `{msg: 58-byte string, id: i64}`. The unsafe isn't super expensive and because the safe path required touching another crate, I opted to go with this option. Please let me know if I should try a different method or modify the crate.

partiql-tools/src/storage.rs
```diff
@@ -194,14 +215,36 @@ impl HeedDB {
...
-    /// the module boundary.
+    /// Drop the returned txn to release the read snapshot.
+    pub fn read_txn(&self) -> Result<heed::RoTxn<'_>, StorageError> {
+        Ok(self.env.read_txn()?)
+    }
+
+    /// Open the named table for read. Returns `TableMissing` if no named
+    /// database called `name` exists in the env (i.e., the table was never
+    /// created or was dropped).
+    pub fn open_table(&self, txn: &heed::RoTxn<'_>, name: &str) -> Result<RowDb, StorageError> {
```
xd1313113, line 226: Do you need validate if name has been registered in `_tables`?

partiql-tools/src/bin/pqlite.rs
```diff
@@ -342,113 +299,156 @@ fn run_repl(debug: &DebugFlags, db: &HeedDB) {
...
-/// cost. That only affects the diagnostic timing line under `--debug`; normal
-/// output is unaffected.
+/// Bundles `HeedCompilationCatalog` with the existing `TableFnCompilationCatalog`
+/// under `PlanCompiler`'s single `"default"` catalog name. Also records the
+/// first unresolved bare table name so `build_compiled` can fail fast with
+/// "Table not found" instead of letting Permissive mode yield a MISSING row.
+struct CombinedCatalog {
+    table_fns: common::TableFnCompilationCatalog,
+    heed: Option<HeedCompilationCatalog>,
+    unresolved_table_name: Arc<Mutex<Option<String>>>,
```
xd1313113, line 336: why Mutex is needed here?

partiql-tools/src/bin/pqlite.rs
hashPirate: Yes, this was intentional per the feedback received in the last PR! Some comments were considered stale or verbose so I had trimmed them down. I can do another sweep to restore clarifying ones.

partiql-tools/src/catalog.rs
hashPirate: Good catch! Yes for now we put the whole table into a single byte buffer. This is temporary because of a lifetime constraint and we can't stream out of the db loop yet. `Box<dyn DataSource + 'static>` can't hold a borrowed `RoTxn`. I just pushed a comment for this tech debt per XuechunHHH's suggestion as well.

partiql-tools/src/storage.rs
hashPirate: Great call! Xuechun brought this up too, so I pushed a fix adding a defensive check right at the beginning of the function. This validates it's registered before actually opening the database.

partiql-tools/src/bin/pqlite.rs
hashPirate: Good question, this is a bit forced by the `CompilationCatalog`. It needs `Send+Sync` but `get_table `only takes the `&self`. Mutex ends up being the only way to keep sync mutability inside. This was chosen to avoid touching other crates and stay in scope for this PR.

partiql-tools/src/row_codec.rs
XuechunHHH: Thanks for the context. If the safe path requires modifying `partiql-eval`, it is justified here. Good to keep as-is.

partiql-tools/src/row_codec.rs
XuechunHHH: Sounds good.

XuechunHHH (approved):
LGTM 🚀   Good to merge from my side

xd1313113 (approved):

PR #646, opened 2026-07-02
feat(pqlite): extend row codec to the all current PartiQL values

hashPirate:
Extends the row_codec to all remaining PartiQL types -> Bool, Bytes, Missing, List, Bag, runtime {...} tuples, and top-level scalar/container rows.

`CREATE TABLE t AS (SELECT [{ 'a': 1 }, { 'a': 2 }] AS xs FROM [0])`

`SELECT * FROM t`  →   `<< { 'xs': [{ 'a': 1 }, { 'a': 2 }] } >>`

-------------

- Reshuffled Tags to scalars from 0x00-0x07 and containers from 0x08-0x0A. Decode can tell the difference between a container-rooted row from a scalar one with one `tag >= TAG_TUPLE` test.
- Top-level scalars stay bare without a wrapper.
- Container counts are backpatched reserving u32. Elements stream straight into the buffer. Zero allocation per container, single pass.
- Inline enforced caps on both paths so that the encoder accepts only rows that the decoder can decode later. 
- 50 integration tests, 8 bisectable commits.

### DIFF
--- a/partiql-tools/src/row_codec.rs
+++ b/partiql-tools/src/row_codec.rs
@@ -17,16 +17,21 @@ pub const MAX_FIELDS_PER_ROW: u32 = 1024;
 /// Cap on `name_len` (UTF-8 byte length) to bound allocations on corruption.
 pub const MAX_NAME_LEN_BYTES: u32 = 1024 * 1024;
 
-pub const TAG_INTEGER: u8 = 0x00;
-pub const TAG_DECIMAL: u8 = 0x01;
-pub const TAG_FLOAT: u8 = 0x02;
-pub const TAG_TUPLE: u8 = 0x03;
-// 0x04 = TAG_BAG     — reserved; rejected.
-pub const TAG_STRING: u8 = 0x05;
-pub const TAG_NULL: u8 = 0x06;
-// 0x07 = TAG_MISSING — reserved; rejected.
-// 0x08 = TAG_BOOL    — reserved; rejected.
-// 0x09 = TAG_BYTES   — reserved; rejected.
+// Tag taxonomy: scalars 0x00-0x07, containers 0x08-0x0A. This split is
+// load-bearing: a later commit tells a container-rooted row from a
+// scalar-rooted one with a single `tag >= TAG_TUPLE` test, so no scalar
+// tag may sit at or above TAG_TUPLE.
+pub const TAG_NULL: u8 = 0x00;
+pub const TAG_MISSING: u8 = 0x01;
+pub const TAG_BOOL: u8 = 0x02;
+pub const TAG_INTEGER: u8 = 0x03;
+pub const TAG_FLOAT: u8 = 0x04;
+pub const TAG_DECIMAL: u8 = 0x05;
+pub const TAG_STRING: u8 = 0x06;
+pub const TAG_BYTES: u8 = 0x07;
+pub const TAG_TUPLE: u8 = 0x08;
+pub const TAG_LIST: u8 = 0x09;
+pub const TAG_BAG: u8 = 0x0A;
 
 #[derive(Debug)]
 pub enum SerializeError {
--- a/partiql-tools/src/row_codec.rs
+++ b/partiql-tools/src/row_codec.rs
@@ -17,6 +17,9 @@ pub const MAX_FIELDS_PER_ROW: u32 = 1024;
 /// Cap on `name_len` (UTF-8 byte length) to bound allocations on corruption.
 pub const MAX_NAME_LEN_BYTES: u32 = 1024 * 1024;
 
+/// Cap on `bytes` payload length to bound allocations on corruption.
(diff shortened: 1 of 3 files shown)


partiql-tools/src/row_codec.rs
```diff
@@ -146,25 +221,146 @@ fn write_value(
...
+    tag: u8,
+    kind: &'static str,
+    depth: u32,
+    buf: &mut Vec<u8>,
+) -> Result<(), SerializeError> {
+    check_encode_depth(depth)?;
+    buf.push(tag);
+    // Reserve the element-count slot; backpatched after the elements are
+    // written so we never allocate a scratch buffer to pre-count.
+    let count_pos = buf.len();
```
johnedquinn, line 339: Nice.

partiql-tools/tests/pqlite_cli.rs
```diff
@@ -587,3 +547,964 @@ fn wire_format_byte_shape_is_stable() {
...
+        "expected list row; got: {out}"
+    );
+    assert!(
+        out.contains("'xs': <<7, 8>>"),
+        "expected bag row; got: {out}"
+    );
+}
+
+#[test]
+fn ctas_then_select_null_inside_list_distinct_from_missing() {
```
johnedquinn, line 1488: Many of these tests are pretty much formatted the same way. Non-blocking on this CR, but it'd be nice to get some parameterized tests here at some point in the near future.
```rust
#[rstest]
#[case("CREATE ...", "SELECT ...", "<<...>>")]
#[case("CREATE ...", "SELECT ...", "<<...>>")]
#[case("CREATE ...", "SELECT ...", "<<...>>")]
fn test_roundtrip(#[case] ddl: &str, #[case] query: &str, #[case] expected: &str) {
    // Create table using ddl
    // Execute query and send PartiQL/Ion output somewhere
    // Perform PartiQL/Ion equality
}
```

partiql-tools/tests/pqlite_cli.rs
```diff
@@ -587,3 +547,964 @@ fn wire_format_byte_shape_is_stable() {
...
+    assert!(ok, "decimal CTAS failed: {err}");
+
+    // Anchor: the value must land in field 'd' as an exact 3.14.
+    let row0 = {
+        let db = partiql_tools::storage::HeedDB::open(&dbp).unwrap();
+        partiql_tools::test_support::read_row(&db, "t", 0)
+    };
+    // {'d': 3.14dec}: TAG_TUPLE, fc=1, name_len=1, 'd', TAG_DECIMAL, scale=2,
+    // mantissa=314 (i128 LE: 3a 01 then 14 zero bytes).
+    let expected: &[u8] = &[
```
johnedquinn, line 1223: Non-blocking: personally not too worried about unit testing the bytes right now. More focused on the end-to-end. Nonetheless, nice work!

johnedquinn (approved):
This is great to see. Nice work!

partiql-tools/src/row_codec.rs
hashPirate: Thank you 😄

partiql-tools/tests/pqlite_cli.rs
hashPirate: Shall do!

partiql-tools/src/row_codec.rs
```diff
@@ -146,25 +221,146 @@ fn write_value(
...
+                .advance()
+                .map_err(|e| SerializeError::Unsupported(format!("tuple advance: {e}")))?;
+            if !has_next {
+                break;
+            }
         }
+        view.step_out()
+            .map_err(|e| SerializeError::Unsupported(format!("tuple step_out: {e}")))?;
     }
+    buf[count_pos..count_pos + 4].copy_from_slice(&count.to_le_bytes());
```
johnedquinn, line 319: This is non-blocking but a potential future optimization. For tuples/bags/lists, I feel like the "count" here should instead be a jump to the end. Because here, this solely is for safe reading of the values, but if we change it to instead be bytes to jump, then we can skip-scan over fields. Consider:

```
input = << << 'a', 'b', 'c' >>, 'd' >>
old bag repr = <bag tag> <field count> <bag tag> <field count> <element 'a'> <element 'b'> <element 'c'> <element 'd'>
new bag repr = <bag tag> <byte count> <bag tag> <byte count> <elements 'a'> <element 'b'> <element 'c'> <element 'd'>
```

The byte count will definitely take more bytes, but if we want to read the 2nd element of the top-level bag, then it's trivial. Just jump n bytes to skip over the first top-level element, rather than statefully looping through the elements to reach m fields.

Again, not needed right now, but a glimpse into how we can take advantage of our own encoding.

hashPirate: Thanks for pointing this out, this is a good optimization for getting O(1) scanning. Added it to the task list!

PR #647, opened 2026-07-08
feat(pqlite): stream table scans in bounded chunks

hashPirate:
Changed how we read tables from disk so we don't load the entire table
into memory at once. Now it loads data in small batches of 64 rows at a
time using a reusable buffer, keeping memory usage very low even for large
tables.

- Rewrote HeedTableSource to read in batches and track progress with a cursor.
- Reuses the same memory buffer for each chunk to avoid extra allocations.
- Added tests for batch boundaries, empty tables, and basic streaming.

### Performance and Tradeoffs

To determine the optimal way to implement streaming out of Heed/LMDB under a `'static` trait object constraint, I benchmarked 5 distinct variants using 100k rows, 12-byte payloads, and u64 keys (averaged over a minimum of 7 trials). The metrics track raw storage retrieval and flat-buffer packing (excluding evaluation/decoding time):

| Variant | Speed (ns/row) | Slowdown | Memory | Implementation |
| :--- | :--- | :--- | :--- | :--- |
| A -> eager (originally) | 7.5 ns | Baseline | O(table) | Dumping the whole table into a vector at once. Fast but blows up memory. |
| B -> naive per-row (My first idea) | 324 ns | 43x slower | O(1) | New transaction for every single row. Turns out it's a terrible idea. |
| C -> cached-handle | 250 ns | 33x slower | O(1) | Reused the table handle but still did a txn per row. Proves the txn lifecycle hurts. |
| D -> point-get chunks (K=64) | 116 ns | 15x slower | O(K) | Txn every 64 rows but did independent gets. Hit a wall searching the B-tree from the root. |
| E -> range-scan chunks (K=64) | 13.3 ns | 1.76x slower | O(K) | The fix. Txn every 64 rows but used a range iterator to sequentially walk the leaf pages. |

I selected Variant E as it breaks the O(table) memory scaling into an O(K) heap footprint, keeping overhead lean while saving gigabytes of RAM at runtime.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -482,12 +482,14 @@ fn execute_query(
             // 4 KiB scratch matches LMDB's typical page size; reused per row.
             let mut scratch_buf: Vec<u8> = Vec::with_capacity(4096);
 
-            let n = match vm.execute() {
+            // Drain the source before opening the write txn: a streaming source
+            // holds a read txn open across iteration, and LMDB rejects opening a
+            // table handle in a write txn while one is live (MDB_BAD_DBI).
+            let mut rows: Vec<Vec<u8>> = Vec::new();
+            match vm.execute() {
                 Ok(partiql_eval::ExecutionResult::Query(iter)) => {
-                    let mut writer = db.create_table(&key).map_err(|e| format!("Error: {}", e))?;
-                    // SAFETY: QueryIterator::next uses unsafe lifetime extension on its
-                    // RegisterReader. Aliasing across iter.next() is UB. Consume `row`
-                    // synchronously, drop it, then push the bytes.
+                    // SAFETY: QueryIterator::next lifetime-extends its RegisterReader;
+                    // aliasing across next() is UB. Consume `row` before the next pull.
                     for r in iter {
                         let row = r.map_err(|e| {
                             format!("Error: {}", StorageError::Execution(format!("{:?}", e)))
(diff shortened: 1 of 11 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -496,13 +498,20 @@ fn execute_query(
...
                     }
-                    writer.commit().map_err(|e| format!("Error: {}", e))?
                 }
                 Err(e) => return Err(format!("Execution setup error: {:?}", e).into()),
+            }
+
+            let n = {
+                let mut writer = db.create_table(&key).map_err(|e| format!("Error: {}", e))?;
+                for encoded in &rows {
+                    writer
```
johnedquinn, line 510: I thought that we would also be writing in batches as well. Seemingly, the table source's `next_row()` will only occasionally invoke `load_next_chunk()` which opens/closes read transactions. So, outside of an invocation of `next_row()` it is safe to open write transactions. With this, you can batch the writes at whatever interval you like and maintain constant memory overhead.

johnedquinn (changes requested):

partiql-tools/src/bin/pqlite.rs
hashPirate: Good catch! The reason I chose not to implement it yet is because we would lose atomicity when doing CTAS. A failure in the middle would leave a partially written table. Is this a tradeoff we are willing to accept?

johnedquinn: We already lose atomicity with opening/closing read transactions as it is -- though, yeah it would make the problem worse than it already is. Ok, fair! We really gotta tackle the transaction state soon. Approving 😄

johnedquinn (approved):

PR #651, opened 2026-07-14
feat(pqlite): parse, lower, and execute INSERT INTO <table> SELECT ...

hashPirate:
Wires up `INSERT INTO <table> <query>` (INSERT ... SELECT) end to end, appending a query's result rows to an existing table. This mirrors the recently merged CTAS pipeline at every layer from the grammar down to the LMDB storage driver.

INSERT INTO t SELECT m.a FROM mem(3,1) m      -- append 3 rows to t
INSERT INTO t SELECT src.a FROM src           -- disk-to-disk
INSERT INTO t SELECT t.a FROM t               -- self-insert

`INSERT` and `INTO` are now reserved keywords (per SQL:2003), meaning queries like `SELECT x AS into FROM t` must now quote the identifier (`"into"`).

Key changes:
- **Grammar & AST:** Parses the spec-faithful, unparenthesized form `INSERT INTO <name> <query>`. It avoids the parenthesized syntax of CTAS's `AS ( <query> )` because `INTO` is structurally unique, preventing any alias/projection ambiguity and avoiding future collisions with column-list slots. Reuses the existing `ast::Insert` node, narrowing the target to a bare table-name `VarRef` during lowering.
- **Lowering:** Adds `LogicalStatement::InsertInto { table_name, query }` to mirror `CreateTableAs`. Unsupported shapes degrade safely to `NotYetImplemented` and lower() rejects DML.
- **Storage Driver:** Implements `HeedDB::open_table_for_append`, which opens an existing table (throwing an error if it's missing) and seeds the writer's `row_id` at `high_water_key + 1` to preserve ordering.
- **Execution & Avoid-Locking:** Shares a new `drain_source_rows` helper with the CTAS path. It fully materializes the source rows in memory before starting the LMDB write transaction. This read-before-write ordering prevents LMDB `MDB_BAD_DBI` errors when a query reads from and inserts into stored tables.

26 tests total (9 parser, 5 lowering, 4 storage, 8 CLI/E2E), covering key reservation regressions, self-inserts, and invalid target errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - *BREAKING* partiql-parser: `Parsed`'s `ast` field replaced by `statements: Vec<AstNode<Statement>>`. The `Item` enum is renamed to `Statement` with bare variants (`Statement::Query(TopLevelQuery)`, `Statement::Ddl(DdlOp)`, `Statement::Dml(Dml)`). The single-field `Ddl` wrapper struct is removed; `DdlOp` is held directly. Code that accessed `parsed.ast` must now use `parsed.statements[0]` and match on `Statement` variants.
+- *BREAKING* partiql-parser: `INSERT` and `INTO` are now reserved keywords (per SQL:2003), so they can no longer be used as bare identifiers. A query such as `SELECT x AS into FROM t` must now quote the identifier (`"into"`).
 
 ### Added
 - partiql-parser: Parsing support for `CREATE TABLE <name>` and `CREATE TABLE <name> AS (<query>)` (CTAS). DDL lowering/evaluation is not yet implemented and surfaces a `NotYetImplemented` error.
+- partiql-parser: Parsing support for `INSERT INTO <name> <query>` (INSERT ... SELECT), lowered to `LogicalStatement::InsertInto`. Other DML forms surface a `NotYetImplemented` error.
 
 ### Removed
 
--- a/partiql-logical/src/lib.rs
+++ b/partiql-logical/src/lib.rs
@@ -487,6 +487,11 @@ pub enum LogicalStatement {
     },
     /// `CREATE TABLE <name>` — the target only, no source query.
     CreateTable { table_name: BindingsName<'static> },
+    /// `INSERT INTO <name> <query>` — target table plus the lowered source query.
+    InsertInto {
+        table_name: BindingsName<'static>,
+        query: LogicalPlan<BindingsOp>,
+    },
 }
 
 /// Represents a `PartiQL` value expression. Evaluation of a [`ValueExpr`] leads to a `PartiQL` value as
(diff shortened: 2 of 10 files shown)


partiql-parser/src/parse/partiql.lalrpop
```diff
@@ -60,6 +65,26 @@ CtasClause: Box<ast::AstNode<ast::TopLevelQuery>> = {
...
+}
+
+// `INSERT INTO <name> <query>`. The source is bare: the only spec paren-slot
+// after a table name is the (unsupported) column list. `Query` not
+// `TopLevelQuery`, so a leading `WITH` source is out of scope.
+Insert: ast::Insert = {
+    "INSERT" "INTO" <lo:@L> <target:SymbolPrimitive> <hi:@R> <source:Query> => {
+        let target_ref = state.node(
+            ast::VarRef { name: target, qualifier: ast::ScopeQualifier::Unqualified },
+            lo..hi,
```
johnedquinn, line 79: Interesting. The name should be allowed to be qualified, and VarRef is likely not the right modeling for declarations. However, this modeling is pre-existing, and I'll leave it as out-of-scope for this PR.

I'll create a follow-up issue for this.

partiql-tools/src/bin/pqlite.rs
```diff
@@ -520,15 +552,43 @@ fn execute_query(
...
-            let total_time = parse_time + lower_time + compile_time + exec_time;
+            print_write_footer(parse_time, lower_time, compile_time, exec_time);
+            return Ok(());
+        }
+        LogicalStatement::InsertInto { table_name, query } => {
+            let key = canonical_table_key(&table_name);
+
+            // Resolve --db before compile + VM setup so a missing flag fails
+            // cleanly before any expensive work.
+            let db = db
```
johnedquinn, line 563: Non-blocking: Let's make --db required for everything honestly. Eventually we'll add support for an in-memory backend.

partiql-logical/src/lib.rs
```diff
@@ -487,6 +487,11 @@ pub enum LogicalStatement {
     },
     /// `CREATE TABLE <name>` — the target only, no source query.
     CreateTable { table_name: BindingsName<'static> },
+    /// `INSERT INTO <name> <query>` — target table plus the lowered source query.
+    InsertInto {
+        table_name: BindingsName<'static>,
```
johnedquinn, line 492: Echoing what I mentioned earlier, but that'll be a follow-up later.

CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Changed
 - *BREAKING* partiql-parser: `Parsed`'s `ast` field replaced by `statements: Vec<AstNode<Statement>>`. The `Item` enum is renamed to `Statement` with bare variants (`Statement::Query(TopLevelQuery)`, `Statement::Ddl(DdlOp)`, `Statement::Dml(Dml)`). The single-field `Ddl` wrapper struct is removed; `DdlOp` is held directly. Code that accessed `parsed.ast` must now use `parsed.statements[0]` and match on `Statement` variants.
+- *BREAKING* partiql-parser: `INSERT` and `INTO` are now reserved keywords (per SQL:2003), so they can no longer be used as bare identifiers. A query such as `SELECT x AS into FROM t` must now quote the identifier (`"into"`).
```
johnedquinn, line 12: 🙌

johnedquinn (approved):

partiql-parser/src/parse/partiql.lalrpop
hashPirate: Agreed! Created the follow up issue to track this in our backlog here: #652

partiql-tools/src/bin/pqlite.rs
hashPirate: Gotcha, created a draft in the backlog :)

PR #653, opened 2026-07-20
feat(pqlite): bootstrap system tables

### LINKED ISSUE #648: feat(pqlite): bootstrap system tables on startup
## Summary

Implement startup bootstrapping for pqlite so that the `_tables` system table is automatically created on first run, and the LMDB unnamed default database tracks schema version to gate future migrations.

## Why NOT `INFORMATION_SCHEMA.TABLES` (schema-qualified approach)

We considered naming system tables using SQL-standard schema-qualified paths (e.g., `INFORMATION_SCHEMA.TABLES`). This was rejected because:

1. **Planner gap:** The logical planner (`partiql-logical-planner/src/lower.rs`) resolves only the root `VarRef` in a FROM clause. For `INFORMATION_SCHEMA.TABLES`, it resolves `INFORMATION_SCHEMA` as a standalone name, then `.TABLES` becomes a `PathComponent::Key` navigation step — it never combines them into a single multi-element `DBRef` catalog lookup.
2. **Engine compiler gap:** The bytecode engine's `resolve_reader_factory()` (`partiql-eval/src/engine/compiler.rs` ~line 1675) only handles `ValueExpr::DBRef` or `ValueExpr::VarRef` as scan sources. A `ValueExpr::Path` (which schema-qualified names produce) hits the "unsupported scan expression type" error.
3. **Multi-crate fix required:** Properly supporting this requires coordinated changes across `partiql-logical-planner` (FROM lowering), `partiql-eval` (engine compiler), and potentially the parser — a significant lift for no user-facing benefit in an embedded engine where `SELECT * FROM _tables` works identically.

The `CompilationCatalog` trait already accepts `&[BindingsName]` (a path slice), so the *interface* is ready. But the planner and compiler that feed it are not. We'll revisit this if/when upstream PartiQL adds schema support.

---

### Single Source of Truth: `_tables`

**`_tables` is the single source of truth for table existence.** It already serves this role today — `open_table()` checks `_tables` before opening any named DB. The change is making it **queryable** by storing row-encoded values instead of empty bytes, and allowing it to resolve itself so users can `SELECT * FROM _tables`.

### `_tables` Key Design

The `_tables` LMDB database continues to use **the table name as the key** for O(1) existence lookups. The value changes from empty bytes to row-encoded data.

```rust
// Key: canonical table name.
// Eventually this will be a delimited fully-qualified table name
// (e.g., "schema"."table") once schema support is added.
type TablesDb = heed::Database<heed::types::Str, heed::types::Bytes>;
```

Values are row-encoded structs (same codec as user table rows) so `SELECT * FROM _tables` decodes them uniformly:

| Column | Type | Description |
|--------|------|-------------|
| `name` | `STRING` | Canonical table name |

### Schema Version (unnamed default DB)

The schema version is stored as a raw `u32` (little-endian) in the **LMDB unnamed default database** (`env.create_database(&mut wtxn, None)`). This DB:

- Always exists (it's the root of the LMDB environment)
- Doesn't count against `max_dbs`
- Is readable regardless of row codec version (avoids chicken-and-egg: can't deserialize rows if the codec itself changed between versions)
- Is never queryable via SQL, never visible to users

| Key | Value (bytes) | Purpose |
|-----|---------------|---------|
| `"schema_version"` | `u32` little-endian | Current bootstrap schema version |

### No `_` prefix guard

The current `_` prefix restriction in `create_table()` is **removed**. System tables are a naming convention, not an enforced namespace. Users may create tables with `_` prefixes if they choose.

---

## Bootstrap Script

The bootstrap script lives at `partiql-tools/src/bootstrap/v1.pql`:

```sql
-- pqlite bootstrap v1
-- Executed on first startup when schema_version < 1.
-- The _tables table is the SINGLE SOURCE OF TRUTH for table existence.

CREATE TABLE _tables;
```

On startup, pqlite reads this file and executes it through the **normal query execution pipeline** (parse → plan → compile → execute). The bootstrap path is simply "read `.pql` file, run each statement." No special internal APIs.

---

## How `CREATE TABLE` Works

The `CREATE TABLE` command implementation does two things atomically:

1. Create the Heed named DB for the table
2. Insert an entry into the `_tables` DB: key = table name, value = row-encoded `{ name: "<table_name>" }`

For `CREATE TABLE _tables` (the bootstrap case), this is self-referential but **not circular**:
- Step 1 creates the `_tables` LMDB named DB
- Step 2 opens the DB that was just created in step 1, and inserts the entry keyed by `"_tables"`

No special-casing. The bootstrap just runs a `.pql` script through the same execution path as any user query.

### CTAS (`CREATE TABLE foo AS SELECT ...`)

Same as above, plus writing query result rows:

```
┌─────────────────────────────────────────────────┐
│ Write Transaction                               │
│  1. env.create_database("foo")                  │
│  2. write query result rows → foo DB            │
│  3. _tables.put("foo", encode({ name: "foo" })) │
│  4. commit                                      │
└─────────────────────────────────────────────────┘
```

### SELECT (`SELECT * FROM foo`)

Catalog resolution queries `_tables` (this is already the case today via `open_table()`):

1. `get_table(["foo"])` → read txn on `_tables` DB → look up key `"foo"`
2. If key exists → table is valid, open its LMDB named DB for scanning
3. If key missing → table does not exist, compilation error

`SELECT * FROM _tables` works the same way — it resolves `_tables` by looking itself up, which always succeeds after bootstrap.

---

## Startup Sequence (prescriptive)

```rust
const CURRENT_SCHEMA_VERSION: u32 = 1;

/// Called once during HeedDB::open(). Reads and executes .pql bootstrap
/// scripts through the normal query pipeline.
fn bootstrap(&self) -> Result<(), StorageError> {
    let current_version = self.read_schema_version(); // reads unnamed default DB, returns 0 if key missing

    if current_version >= CURRENT_SCHEMA_VERSION {
        return Ok(());
    }

    if current_version < 1 {
        // Read and execute v1.pql through the normal query pipeline
        let script = include_str!("bootstrap/v1.pql");
        self.execute_script(script)?;

        // Stamp version in the unnamed default DB (raw bytes, no row codec)
        let mut wtxn = self.env.write_txn()?;
        self.default_db.put(&mut wtxn, "schema_version", &1u32.to_le_bytes())?;
        wtxn.commit()?;
    }

    // Future: if current_version < 2 { ... }

    Ok(())
}
```

### Full open() sequence

```rust
pub fn open(path: &Path) -> Result<Self, StorageError> {
    let env = open_env(path)?;

    // Unnamed default DB — always exists, stores schema_version as raw bytes
    let mut wtxn = env.write_txn()?;
    let default_db = env.create_database::<Str, Bytes>(&mut wtxn, None)?;
    wtxn.commit()?;

    let db = HeedDB { env, default_db };
    db.bootstrap()?;

    Ok(db)
}
```

---

## Acceptance Criteria

- [ ] Schema version is stored in the unnamed default DB as raw `u32` LE bytes
- [ ] `_tables` system table is created by executing `v1.pql` through the normal query pipeline
- [ ] `_tables` uses table name as key for O(1) existence lookups
- [ ] `_tables` values are row-encoded `{ name: "..." }` (queryable)
- [ ] `_tables` contains an entry for itself (self-referential)
- [ ] `CREATE TABLE` and `CTAS` both insert row-encoded entries into `_tables` as part of their normal flow
- [ ] `SELECT * FROM foo` resolves `foo` by querying `_tables` (existing behavior, unchanged)
- [ ] `SELECT * FROM _tables` returns all table names including itself
- [ ] Second startup with existing DB skips bootstrap (version check passes)
- [ ] `_` prefix guard is removed from `create_table()`
- [ ] `StorageError::ReservedName` variant is removed
- [ ] Bootstrap script exists at `partiql-tools/src/bootstrap/v1.pql`
- [ ] Existing integration tests continue to pass (update tests that assert on `ReservedName`)
- [ ] New test: fresh DB → verify `_tables` exists and is queryable → create user table → verify it appears in `_tables`

## Follow-up (separate issue)

- **Caching:** Add in-memory `HashSet<String>` cache populated from `_tables` at startup for faster catalog resolution during compilation.

## Files to Modify

| File | Change |
|------|--------|
| `partiql-tools/src/storage.rs` | Remove `_` prefix guard and `ReservedName` variant; add unnamed default DB for version; add `bootstrap()` with `include_str!`; update `create_table()` to write row-encoded values into `_tables` instead of empty bytes |
| `partiql-tools/src/catalog.rs` | Allow `_tables` to resolve as a queryable table (it currently resolves via `list_table_names` but not as a scan target) |
| `partiql-tools/src/bin/pqlite.rs` | Wire up `execute_script` for bootstrap; simplify CTAS since table registration is now in `create_table()` |
| `partiql-tools/src/bootstrap/v1.pql` | New file — canonical bootstrap script |
| `partiql-tools/tests/pqlite_cli.rs` | Add bootstrap + introspection tests; update tests that assert on `ReservedName` |

hashPirate:
# feat(pqlite): bootstrap system tables on startup

Closes #648.

## Summary
When pqlite starts for the first time, it now automatically creates the _tables system catalog. It does this by running bootstrap/v1.pql through the standard query pipeline and saving a schema version in the database to skip this step on future boots. The _tables catalog is now queryable via SELECT * FROM _tables.

## Key Changes

- Stores a version number (u32 LE) in the LMDB default database to detect if first startup.

- _tables automatically includes an entry for itself.

- Removed _ prefix restriction.

- _tables is read-only for users to prevent manual corruption of the catalog.

- Added a utility that handles multi-statement scripts (splits by ; and strips comments). This is limited to bootstrap mode only.

## Deviations from Spec

I made a few adjustments to the original design to improve maintainability and future-proof the implementation:

- I moved bootstrapping out of HeedDB::open() and into the main binary. This keeps the storage module lean by ensuring it only handles low-level primitives instead of needing access to the parser and VM.

- I updated the _tables format to store names as a list ({name: ['foo']}) rather than a single string. This allows us to support qualified table names later without needing a data migration.

- Used a NUL-prefixed key (b"\0schema_version") for the version stamp. This makes it structurally impossible for a user to accidentally overwrite the version key by naming a table schema_version.

- The system now fails fast if the schema version is newer than the binary or if a marked database is missing its self-referential _tables entry.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -1,6 +1,6 @@
 use partiql_tools::common;
 
-use common::{create_table_fn_catalog, lower_statement, parse};
+use common::{create_table_fn_catalog, lower_statement, parse, parse_statements};
 use partiql_common::catalog::CatalogId;
 use partiql_eval::plan::EvaluationMode;
 use partiql_eval::value::Shape;
@@ -126,26 +126,6 @@ fn normalize_query(input: &str) -> &str {
     trimmed.strip_suffix(';').unwrap_or(trimmed).trim_end()
 }
 
-/// Split a trusted bootstrap script on `;`. Strips `--` line comments first so a
-/// `;` in a comment doesn't split. Only `--` comments are handled: no `/* */`,
-/// and no `;` inside string literals.
-fn split_script_statements(script: &str) -> Vec<String> {
-    let stripped: String = script
-        .lines()
-        .map(|line| match line.find("--") {
-            Some(i) => &line[..i],
-            None => line,
-        })
-        .collect::<Vec<_>>()
-        .join("\n");
-    stripped
-        .split(';')
-        .map(str::trim)
-        .filter(|s| !s.is_empty())
-        .map(str::to_string)
-        .collect()
-}
-
 /// Open the DB and bring it to CURRENT_SCHEMA_VERSION. Idempotent + crash-safe.
 fn open_and_bootstrap(path: &std::path::Path) -> Result<Arc<HeedDB>, Box<dyn std::error::Error>> {
     let db = Arc::new(HeedDB::open(path)?);
@@ -183,14 +163,24 @@ fn open_and_bootstrap(path: &std::path::Path) -> Result<Arc<HeedDB>, Box<dyn std
     Ok(db)
 }
 
-/// Run the v1 bootstrap script. Idempotent: a crash-recovery DB whose _tables
-/// already exists yields a typed StorageError::TableExists, reconciled against
-/// the self-entry; any other error propagates.
+/// Run the v1 bootstrap script. The `;`-separated script is parsed as a unit
+/// and each statement executed in order. Idempotent: a crash-recovery DB whose
+/// _tables already exists yields a typed StorageError::TableExists, reconciled
+/// against the self-entry; any other error propagates.
 fn bootstrap_v1(db: &Arc<HeedDB>) -> Result<(), Box<dyn std::error::Error>> {
     let script = include_str!("../bootstrap/v1.pql");
     let debug = DebugFlags::from_args(&[]);
-    for stmt in split_script_statements(script) {
-        match execute_query(&stmt, &debug, Some(Arc::clone(db)), OutputMode::Silent) {
+    let parsed = parse_statements(script).map_err(|e| format!("Parse error: {:?}", e))?;
+    for stmt in &parsed.statements {
+        // parse_time is a bootstrap step, not user-facing; Silent mode prints no
+        // timing footer, so a zero placeholder is never observed.
+        match execute_statement(
+            stmt,
+            &debug,
+            Some(Arc::clone(db)),
+            OutputMode::Silent,
+            std::time::Duration::ZERO,
+        ) {
             Ok(()) => {}
             Err(e) => {
                 let is_self_tables_exists = e
@@ -594,8 +584,6 @@ fn execute_query(
     db: Option<Arc<HeedDB>>,
     mode: OutputMode,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    let catalog = create_table_fn_catalog();
-
     let parse_start = Instant::now();
     let parsed = parse(query_str).map_err(|e| format!("Parse error: {:?}", e))?;
     let parse_time = parse_start.elapsed();
@@ -604,12 +592,28 @@ fn execute_query(
         eprintln!("[AST] {:?}", parsed);
     }
 
-    // Lowering operates on a single statement.
+    // `exec` and the REPL run one statement at a time; a `;`-separated script is
+    // parsed with `parse_statements` and each statement run via `execute_statement`
+    // in a loop (see `bootstrap_v1`).
     let stmt = match parsed.statements.as_slice() {
         [stmt] => stmt,
         // Wording matches `LogicalPlanner::lower` for parity across error sites.
         _ => return Err("Lower error: multi-statement input".into()),
     };
+    execute_statement(stmt, debug, db, mode, parse_time)
+}
+
+/// Execute a single already-parsed statement. Split from [`execute_query`] so a
+/// parsed script can dispatch each statement in a loop without re-parsing.
+fn execute_statement(
+    stmt: &partiql_ast::ast::AstNode<partiql_ast::ast::Statement>,
+    debug: &DebugFlags,
+    db: Option<Arc<HeedDB>>,
+    mode: OutputMode,
+    parse_time: std::time::Duration,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let catalog = create_table_fn_catalog();
+
     let lower_start = Instant::now();
     let statement =
         lower_statement(&*catalog, stmt).map_err(|e| format!("Lower error: {:?}", e))?;
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -21,6 +21,16 @@ use reedline::{
 
 const HISTORY_CAPACITY: usize = 1000;
 
+/// The schema version this binary bootstraps to and requires.
+const CURRENT_SCHEMA_VERSION: u32 = 1;
+
+/// Whether `execute_query` prints. Bootstrap runs muted.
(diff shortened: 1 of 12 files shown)


partiql-tools/src/bin/pqlite.rs
```diff
@@ -116,6 +126,89 @@ fn normalize_query(input: &str) -> &str {
     trimmed.strip_suffix(';').unwrap_or(trimmed).trim_end()
 }
 
+/// Split a trusted bootstrap script on `;`. Strips `--` line comments first so a
+/// `;` in a comment doesn't split. Only `--` comments are handled: no `/* */`,
+/// and no `;` inside string literals.
+fn split_script_statements(script: &str) -> Vec<String> {
```
johnedquinn: Can we leverage the parser here and then pass in each individual statement in a loop to be executed?

partiql-tools/src/row_codec.rs
```diff
@@ -89,6 +89,41 @@ pub fn serialize_row(
     }
 }
 
+/// Encode a `_tables` catalog row: `{ name: [<elements>] }`, where `name` is a
+/// PartiQL list of the (unescaped) path elements. Single-element today
+/// (`["foo"]`); multi-element paths are the forward-compatible shape. `buf` is
+/// cleared on entry. Hand-mirrors the wire format — no VM `RegisterReader` needed.
+pub fn serialize_name_row(elements: &[&str], buf: &mut Vec<u8>) -> Result<(), SerializeError> {
```
johnedquinn, line 96: Can we re-use write_value from elsewhere in codec? You'll just need to create a ValueView.

partiql-tools/src/bin/pqlite.rs
hashPirate: Good call, the parser didn't originally support this and parsed a single statement so it rejected `;`. I added support for it as it's the cleaner solution long-term. We now run each statement through the pipeline in a loop and this sets us up to use the multi-statement bootstrap scripts we want later (like permissions table).

partiql-tools/src/row_codec.rs
hashPirate: Resolved as we discussed and left as-is! 

Documenting why, for future reference: Reusing `write_value` needs a `ValueView`, all constructors we could use are `pub(crate)` in `partiql-eval`, and we would have to expose engine internals.

johnedquinn (approved):
Thank you! Great to see multi-statement parsing.

PR #654, opened 2026-07-20
fix: unblock CI on current toolchain (semicolons + ion pin)

hashPirate:
- **Trailing semicolons in `partiql-types` macros** — `type_bag!` / `type_array!`
    had a trailing `;` inside the macro body. The `semicolon_in_expressions_from_macros`
    lint now treats this as a hard error on nightly. Behavior-neutral.
  - **Pin `ion-rs` to `=1.0.0-rc.11`** — the caret requirement floated to rc.12
    (no committed lockfile), which changed Ion text output and broke snapshots.
    The exact pin holds it at rc.11. Unpinning is tracked in #<ION_ISSUE>.
  - **Re-accept `verify_ion_tests` snapshot** — one snapshot drifted from an
    unrelated std change (`TryFromIntError` Debug format: `(())` → `(PosOverflow)`),
    which no ion pin can fix. Re-accepted and audited to confirm only that known
    drift, no behavior change.
    
    
    Made #655 to track the ion-pin as a future TODO.

### DIFF
--- a/extension/partiql-extension-csv/Cargo.toml
+++ b/extension/partiql-extension-csv/Cargo.toml
@@ -32,6 +32,7 @@ itertools = "0.14"
 unicase = "2.7"
 rust_decimal = { version = "1.36.0", default-features = false, features = ["std"] }
 ion-rs_old = { version = "0.18", package = "ion-rs" }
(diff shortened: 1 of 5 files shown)


am357 (approved):
LGTM—Thank you for fixing this miss!

PR #658, opened 2026-07-27
feat(pqlite): file-driven end-to-end test harness

hashPirate:
This PR introduces a file-driven `.test.ion` end-to-end test harness for `pqlite` along with the library extraction that makes it possible.

Each fixture is a stream of top-level `{sql, expect}` Ion structs. These are discovered by `rstest` and run against a single long-lived `PqliteSession`. This mirrors the CLI dispatch path so our tests exercise the real engine.
```ion
{ sql: "CREATE TABLE t AS (SELECT * FROM << {'a':1} >>)", expect: { created_table: { name: ["t"], rows: 1 } } }
{ sql: "SELECT * FROM t", expect: { rows: $bag::[ { a: 1 } ] } }

// Skip a known-broken case while keeping it in-tree.
skip::"SELECT * FROM not_yet_supported"
```

The CLI `exec` and REPL logic has been extracted from `bin/pqlite.rs` into a shared session module. The `exec` subcommand also gains a `--format ion` option to provide a stable, machine-parseable envelope. Text output remains byte-identical to the `dev` branch.

### Key Changes

* **Library extraction:** `bin/pqlite.rs` is now about 680 lines shorter. The exec and REPL logic moved into `PqliteSession::run(&Commands, out, err)`. The CLI and the test harness now dispatch through the exact same `clap`-typed entry point.
* **Streaming SELECT:** Rows now stream row-by-row through `stream_query_ion` and `stream_query_text` directly into the caller's writer. Peak memory usage is reduced to just one row rather than the entire result set. Additionally, `--format ion` output is silent on `stderr` for successful queries.
* **Loader:** We now use a top-level Ion stream that fails loudly at load time on unknown or duplicate fields, empty expects, empty files, and `expect: { error: "" }`. This ensures a fixture typo cannot silently disable a test case. Table-driven schema tests now cover every `LoadError` variant.
* **Matcher:** Strict `IonEq` via `IonData` is implemented (e.g., NULL is not equal to MISSING, type-strict, decimal precision is preserved).
* **`skip::` annotation:** Fixture authors can keep known-broken test cases in the tree while the compiler catches up. Using `skip::"bare SQL"` or `skip::{sql, expect}` will record the step without executing it.
* **LMDB isolation:** A per-case mutex bounds live environments to one. Otherwise, many-core hosts hit `EINVAL` on environment-open with a 1 GiB `map_size` multiplied by N parallel cases.
* **Canonical encoder reuse**: Value→Ion goes through PartiqlEncodedIonValueEncoder from partiql-extension-ion instead of a hand-rolled encoder.

### Testing Summary

Added 152 tests (60 lib, 68 CLI, 24 e2e) across 17 fixtures. These cover:
* CREATE, INSERT, SELECT, and CTAS round-trips (decimal, list, mixed-scalar, bool)
* Aggregates, nested structs, scalars, and empty containers
* SELECT VALUE
* MISSING vs NULL
* Quoted identifiers and WHERE clauses
* Runtime errors and one `skip::` demo

Removed 6 CLI tests whose only assertion was `stdout.contains("<literal>")` because they are now fully covered by fixtures with stricter `IonEq` semantics.

### DIFF
--- a/partiql-tools/src/bin/pqlite.rs
+++ b/partiql-tools/src/bin/pqlite.rs
@@ -1,35 +1,14 @@
-use partiql_tools::common;
-
-use common::{create_table_fn_catalog, lower_statement, parse, parse_statements};
-use partiql_common::catalog::CatalogId;
-use partiql_eval::plan::EvaluationMode;
-use partiql_eval::value::Shape;
-use partiql_eval::{CompilationContext, ExecutionCatalog, ExecutionContext, PlanCompiler};
-use partiql_logical::LogicalStatement;
-use partiql_tools::catalog::{HeedCompilationCatalog, HeedExecutionCatalog};
-use partiql_tools::storage::{HeedDB, StorageError};
-use partiql_value::{Tuple, Value};
 use std::borrow::Cow;
-use std::sync::{Arc, Mutex};
-use std::time::Instant;
 
-use clap::{Parser, Subcommand};
+use partiql_tools::session::{normalize_query, Commands, DebugFlags, OutputFormat, PqliteSession};
+
+use clap::Parser;
 use reedline::{
     FileBackedHistory, History, Prompt, PromptEditMode, PromptHistorySearch, Reedline, Signal,
     ValidationResult, Validator,
 };
 
 const HISTORY_CAPACITY: usize = 1000;
-
-/// The schema version this binary bootstraps to and requires.
-const CURRENT_SCHEMA_VERSION: u32 = 1;
-
-/// Whether `execute_query` prints. Bootstrap runs muted.
-#[derive(Clone, Copy)]
-enum OutputMode {
-    User,
-    Silent,
-}
 
 /// Crate version joined with the git commit SHA captured in build.rs.
 const VERSION: &str = concat!(env!("CARGO_PKG_VERSION"), "@", env!("PQLITE_GIT_SHA"));
@@ -54,54 +33,32 @@ struct Cli {
     command: Option<Commands>,
 }
 
-struct DebugFlags {
-    ast: bool,
-    plan: bool,
-    program: bool,
-}
-
-impl DebugFlags {
-    fn from_args(args: &[String]) -> Self {
-        let all = args.iter().any(|s| s == "*");
-        DebugFlags {
-            ast: all || args.iter().any(|s| s == "ast"),
-            plan: all || args.iter().any(|s| s == "plan"),
-            program: all || args.iter().any(|s| s == "program"),
-        }
-    }
-}
-
-#[derive(Subcommand)]
-enum Commands {
-    /// Execute a single query immediately
-    Exec {
-        /// The mandatory PartiQL query string to run
-        query: String,
-    },
-}
-
 fn main() {
     let cli = Cli::parse();
     let debug = DebugFlags::from_args(&cli.debug);
 
-    match &cli.command {
-        Some(Commands::Exec { query }) => {
-            let query = normalize_query(query);
-            if query.is_empty() {
+    match cli.command {
+        Some(Commands::Exec { query, format }) => {
+            // Reject empty query before opening the database so a `--db <path>`
+            // invocation with a blank query does not create the file on disk.
+            if normalize_query(&query).is_empty() {
                 eprintln!("Error: empty query");
                 std::process::exit(1);
             }
-            let db = match cli.db.as_deref() {
-                Some(p) => match open_and_bootstrap(p) {
-                    Ok(db) => Some(db),
+            let cmd = Commands::Exec { query, format };
+            let session = match cli.db.as_deref() {
+                Some(p) => match PqliteSession::open(p, debug) {
+                    Ok(s) => s,
                     Err(e) => {
                         eprintln!("Error: could not open database at {}: {}", p.display(), e);
                         std::process::exit(1);
                     }
                 },
-                None => None,
+                None => PqliteSession::open_without_db(debug),
             };
-            if let Err(e) = execute_query(query, &debug, db, OutputMode::User) {
+            let mut stdout = std::io::stdout();
+            let mut stderr = std::io::stderr();
+            if let Err(e) = session.run(&cmd, &mut stdout, &mut stderr) {
                 eprintln!("{}", e);
                 std::process::exit(1);
             }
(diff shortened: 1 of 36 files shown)


partiql-tools/tests/pqlite/cases/errors/skipped_step_demo.test.ion
```diff
@@ -0,0 +1,4 @@
+// A step known to fail today; skip it so the rest of the case still runs.
+skip::"SELECT does_not_exist_yet FROM ghost"
```
johnedquinn, line 2: What is this test trying to convey? I know it's skipped, but I'm trying to understand what it's for

partiql-tools/src/bin/pqlite.rs
```diff
@@ -54,54 +33,32 @@ struct Cli {
...
+            let session = match cli.db.as_deref() {
+                Some(p) => match PqliteSession::open(p, debug) {
+                    Ok(s) => s,
                     Err(e) => {
                         eprintln!("Error: could not open database at {}: {}", p.display(), e);
                         std::process::exit(1);
                     }
                 },
-                None => None,
+                None => PqliteSession::open_without_db(debug),
```
johnedquinn, line 78: Some of this logic regarding exec accepting an optional db, while repl requiring a db could be useful by making it `pqlite open "mydb.db"` and `pqlite exec --db "mydb.db"`. So, clap handles mandatory/optional args/flags. Non-blocking, but something I might change.

partiql-tools/tests/pqlite_e2e/runner.rs
```diff
@@ -0,0 +1,231 @@
...
+                    return Err(CaseError::Exec {
+                        step_index: i,
+                        sql: step.sql.clone(),
+                        source: format!(
+                            "expected top-level struct envelope, got {}",
+                            actual.ion_type()
+                        ),
+                    });
+                }
+                if IonData::from(actual.clone()) != IonData::from(expect.clone()) {
```
johnedquinn: Nice 😄 . This will only work if we maintain ordering even for bags, but the implementation today internally does that anyways. We'll cross that if/when it actually becomes an issue.

partiql-tools/src/session.rs
```diff
@@ -0,0 +1,373 @@
...
+                }
+                Ok(())
+            }
+            Ok(RunResult::NonQuery(outcome)) => {
+                flush_debug(outcome.debug(), err)?;
+                if let OutputFormat::Ion = format {
+                    write_outcome_ion(&outcome, out)?;
+                } else {
+                    render_non_query_text(&outcome, err)?;
+                }
```
johnedquinn: I feel like the testing facilities have bled into the implementation of session itself. Session should likely return the RunResult/StatementOutcomes actual objects. Then, the CLI can render it to the end-user through whatever facility, whereas the testing framework would convert to Ion.

That way, you don't need to be actually printing to stdout/stderr in the test suite. We're just comparing the RunResult.

Maybe we have some shared module for i/o, but those would just be leveraged by pqlite and the test harness directly.

partiql-tools/src/session/value.rs
```diff
@@ -0,0 +1,201 @@
...
+
+fn scalar_err<E: std::fmt::Debug>(e: E) -> RowConvertError {
+    RowConvertError(format!("scalar read: {e:?}"))
+}
+
+fn walk_err<E: std::fmt::Debug>(op: &'static str) -> impl FnOnce(E) -> RowConvertError {
+    move |e| RowConvertError(format!("{op}: {e:?}"))
+}
+
+/// `ValueView::step_in` returns `IllegalState("cannot step into empty X")` for
```
johnedquinn, line 119: Wow, I'm so surprised by this! This should be fixed, though non-blocking for this PR.

partiql-tools/src/session/planner.rs
```diff
@@ -0,0 +1,172 @@
...
+    let catalog_id = context.add_catalog("default", catalog);
+
+    let mut compiler = PlanCompiler::new(&context, EvaluationMode::Permissive);
+    let compiled = compiler
+        .compile(logical)
+        .map_err(|e| format!("Compile error: {:?}", e))?;
+
+    // Recover from a poisoned lock so a panic in one probe doesn't downgrade a
+    // hard "Table not found" error into a Permissive-mode MISSING row.
+    let first_unresolved = unresolved_table_name
```
johnedquinn, line 87: Interesting. This will need to be fixed. Even in permissive mode, unresolved table names should fail planning/compilation.

Non-blocking.

partiql-tools/src/session/planner.rs
```diff
@@ -0,0 +1,172 @@
...
+/// `ScanId → table-name` mappings are populated before the VM calls `create()`.
+pub(super) fn build_exec_context(
+    db: Option<Arc<HeedDB>>,
+    catalog_id: CatalogId,
+    compiled: &partiql_eval::CompiledPlan,
+) -> ExecutionContext {
+    let mut exec_context = ExecutionContext::new();
+    exec_context.register_table_function("rand", Arc::new(common::RandTableFunction));
+    exec_context.register_table_function("mem", Arc::new(common::MemTableFunction));
+    exec_context.register_table_function("scan_ion", Arc::new(common::ScanIonTableFunction));
```
johnedquinn, line 112: I'd love to eventually use your functions and then trash these and the passed in column names for the `TableFnCompilationCatalog`.

partiql-tools/src/session/outcome.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub exec: Duration,
+}
+
+/// AST/plan/program captured under `--debug`, flushed to stderr.
+#[derive(Debug, Default)]
+pub struct DebugCapture {
+    pub ast: Option<String>,
+    pub plan: Option<String>,
+    pub program: Option<String>,
+}
```
johnedquinn, line 42: Love this! Thank you!

partiql-tools/src/session/exec.rs
```diff
@@ -0,0 +1,372 @@
...
+/// What `exec::run` streamed to `out`. Query rows are already on the wire; the
+/// caller only needs count + timing for the footer. Non-query outcomes carry
+/// their own metadata for the caller to render.
+pub(super) enum RunResult {
+    Query {
+        row_count: u64,
+        timing: StatementTiming,
+    },
+    NonQuery(StatementOutcome),
+}
```
johnedquinn: Why didn't Query make it into StatementOutcome? Or the other way around?

partiql-tools/tests/pqlite/cases/errors/skipped_step_demo.test.ion
hashPirate: Ah, this was just an example in a real case so that someone opening cases/ could see what a skip:: step looks like. I can remove it if it is unnecessary.

partiql-tools/src/bin/pqlite.rs
hashPirate: I actually like this better, I'll add it to this PR.

partiql-tools/src/session/value.rs
hashPirate: Yeah, surprised me too haha. This can go away once `partiql-eval` changes.

partiql-tools/src/session/outcome.rs
hashPirate: Thank you :)))

partiql-tools/src/session/planner.rs
hashPirate: Yeah! Super excited for the ingestion functions 😄 , I can remove these in the same PR.

partiql-tools/src/session.rs
hashPirate: Great call, restructured! `session.run` returns a RunOutcome. Moved renderers to a `session::render` module that pqlite and the harness can both call. The harness compares the StatementOutcome for writes and query cases go to Ion. Merged RunResult into RunOutcome as well as suggested by the other comment 💪

partiql-tools/src/session/exec.rs
hashPirate: Addressed at the other comment :)

partiql-tools/tests/pqlite_e2e/runner.rs
```diff
@@ -136,46 +131,229 @@ pub fn run_case(path: &Path) -> Result<(), CaseError> {
...
+                source: format!("{e}"),
+            })?;
+            continue;
+        };
+
+        match outcome {
+            RunOutcome::Query(handle) => {
+                let mut buf = Vec::<u8>::new();
+                handle
+                    .drain(|rows, shape| render_query_ion(rows, shape, &mut buf))
```
johnedquinn, line 175: Nice!

johnedquinn (approved):
LGTM !

PR #659, opened 2026-08-06
fix(eval): stop arena reallocation from invalidating live pointers

hashPirate:
## Summary / Problem
The per-row arena in the VM was using a single `Vec<u8>` to store both the ingested row and temporary per-field values. When temporary allocations pushed the `Vec` past its capacity, it would reallocate in memory. This invalidated raw pointers and references that were already handed out into earlier bytes of the buffer.

## Key Changes
- **Switched backing storage to fixed chunks (`Vec<Vec<u8>>`):** Replaced the single `Vec<u8>` with a vector of fixed-size chunks. Because chunks are never resized after creation, any pointer handed out from a chunk stays valid until `reset()` is called.
- **Alignment change:** Alignment is computed from each chunk's real address
  `(chunk.as_ptr().add(from).align_offset(align))`, not a chunk-relative offset, since `Vec<u8>` is only 1-byte aligned
- **End-to-end regression test:** Forces the arena past its default 8 KiB capacity while the source row is actively being processed.

### DIFF
--- a/partiql-eval/src/engine/arena.rs
+++ b/partiql-eval/src/engine/arena.rs
@@ -3,9 +3,8 @@ use std::cell::{Cell, UnsafeCell};
 
 pub type SlotId = u16;
 
-/// Bump allocator for temporary query values
+/// Bump allocator for temporary query values.
 ///
-/// Provides O(1) allocation and O(1) bulk deallocation via `reset()`.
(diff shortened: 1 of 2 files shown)


johnedquinn (approved):
Thank you for this fix!

PR #661, opened 2026-08-07
feat(pqlite): pqlite ingestion functions

hashPirate:
## Summary

  Adds four table functions on top of a shared JSON/Ion decoder. Each one swaps the byte source:

  | Function        | Reads from                                     |
  |-----------------|------------------------------------------------|
  | `read(path)`    | Ion or JSON file                               |
  | `stdin()`       | Ion or JSON piped to stdin                     |
  | `exec(command)` | `sh -c <command>`, decodes its stdout          |
  | `curl(url)`     | HTTP GET, decodes the response body            |
  
  Extends the decoder to handle nested objects/lists and top-level `[...]` wrappers, so real API payloads come through as rows.
  
  Removes `rand()` and `scan_ion()` - `read()` covers the Ion path.
  
  ## Example
  
  ```sh
  pqlite exec --db /tmp/issues.pqlite "
    CREATE TABLE issues AS (
      SELECT number, state, title, user
      FROM curl('https://api.github.com/repos/partiql/partiql-lang-rust/issues?state=all&per_page=100')
    )
  " 
  # Created table issues (100 rows)
  
  pqlite exec --db /tmp/issues.pqlite "
    SELECT number, state, title
    FROM issues
    WHERE user.login = 'hashPirate'
  " 
  # 21 rows, nested-field filter across a fresh process
  ```
  Also works piped:
```
  gh issue list --json number,state,author \
    | pqlite exec --db /tmp/issues.pqlite \
        'CREATE TABLE issues AS (SELECT * FROM stdin())'
  
```
  Notes
  
  - curl() has a 30 s global timeout and a 1 GiB body cap
  - exec() streams the child's stdout to prevent large producers from causing OOM if buffered

### DIFF
--- a/about.toml
+++ b/about.toml
@@ -22,5 +22,6 @@ accepted = [
     "Zlib",
     "ICU",
     "Unicode-DFS-2016",
-    "Unicode-3.0"
+    "Unicode-3.0",
+    "CDLA-Permissive-2.0"
 ]
--- a/deny.toml
+++ b/deny.toml
@@ -40,6 +40,7 @@ allow = [
     "Zlib",
     "ICU",
     "Unicode-3.0",
+    "CDLA-Permissive-2.0",
 ]
 
 # Allow 1 or more licenses on a per-crate basis, so that particular licenses
--- a/partiql-tools/Cargo.toml
+++ b/partiql-tools/Cargo.toml
@@ -71,7 +71,7 @@ clap = { version = "4.6.1", features = ["derive"] }
 heed = { version = "0.20", default-features = false }
 reedline = "0.48.0"
 rust_decimal = { version = "1", default-features = false, features = ["std"] }
-ureq = "3"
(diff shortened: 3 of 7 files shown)


about.toml
```diff
@@ -22,5 +22,6 @@ accepted = [
     "Zlib",
     "ICU",
     "Unicode-DFS-2016",
-    "Unicode-3.0"
+    "Unicode-3.0",
+    "CDLA-Permissive-2.0"
```
johnedquinn, line 26: I'm not sure that we can use this license. I'll need to check with the team.

hashPirate: Hopefully! I was looking through and it doesn't restrict much, but it does have some obligations.

hashPirate: Changed it to use the OS trust store for curl(). Should no longer pull in CDLA on our build targets, but I have left the allowlist entry alone for now depending on what is decided.

